PR #4, opened 2022-02-16
Incorporation isotopic discrimination for C3/C4 plants and soil water…

Alielav:
… stress on beta

- Incorporation of soil water stress on ci/ca via beta

- Change of ci/ca model for C4 plants

- Incorporation of discrimination against 13C for C3 and C4 plants

- Incorporation of discrimination against 14C


davidorme:
@Alielav All of those look in good shape - I'm revising some of the existing formulation to simplify some equations but I will then try and fold them in manually. A few things:

* Some unit testing values would be great - this is all beyond the `rpmodel` functionality and I have been using that for sanity testing so far.
* The CalcDiscrimination class overlaps extremely strongly with CalcOptimalChi, even down to the options of C4 and Prentice14, and thematically these calculations seem to make sense within that existing class. Does it make sense to simply extend CalcOptimalChi to include these calculations?
* So the theta, beta_c3 method is an alternative approach to estimating soil moisture effects? We're rapidly accumulating those 😄 . I haven't quite looked at the flow, but at the moment I think it replaces some options - and I think we should maintain options in parallel for a while at least?

davidorme:
I'm closing this without pulling in the changes. All of the functionality has been incorporated, with some refactoring, in branch `feature/isotopic_discrimination`, along with documentation. Merged into `develop` in commit c262aba.

PR #25, opened 2023-02-01
Remove use of `Union[float, NDArray]` in typing

### LINKED ISSUE #22: Revisit scalar and array inputs
Many of the functions are set up to handle users providing data as scalar values _or_ `numpy.ndarrays`. That makes for more complex typing and code than simply demanding `numpy.ndarray` inputs in the first place. And that is not a particularly onerous requirement for a fairly niche use case. So:

* Review code to remove scalar input options
* Update typing to NDArray only
* Remove code sections that are needed for scalar use scenarios.

davidorme:
This PR removes all of the code associated with allowing users to pass arguments in as either scalar values (type `float` basically) or `numpy` `NDArray` types. Those arguments are now all `NDArray` simplifying the code.

Fixes #22


davidorme:
This PR now also adds some QA chain updates to resolve `poetry` and `isort` issues, which dragged a `black` update in and a bunch of cosmetic changes removing blank lines after `def` and `class`. Which I _personally_ prefer, but that's `black` for you.

PR #47, opened 2023-02-23
Bug in `PModel.summarize`

davidorme:
The `PModel.summarize` method was using `getattr` rather than `hasattr` to test whether the `PModel.estimate_productivity` method had been run. This fixes that error and adds a test for the correct functioning of the `PModel.summarize` method.


davidorme:
Ooops - this PR branch is forked off a feature branch, not `develop`.

PR #85, opened 2023-08-18
Profiling

### LINKED ISSUE #81: Profiling PyRealm To Identify Performance Bottlenecks & Regressions
As discussed, we should investigate profiling the code-base for the p-model to identify bottlenecks and identify areas where we can improve the performance. There is scope to convert this into an automated process to run alongside other testing or create a trigger that causes a failure if a regression is introduced (if desired). 

I'll use this issue to track any profiling work and link a pull request against it if we decide to include profiling as a standard action on push to branches in this repository. 

We could use a pytest plugin like pytest-profiling to perform the dynamic profiling as part of the CI/test suites. This runs cProfile under the hood plus a few other tools to generate heat graphs etc. but we could limit the scope of any automated process to only using cProfile and responding to the output data. Other tools like PyInstrument exist for statistical profiling. 

For benchmarking we could use the profiling tools, however this may be heavy handed and something like tesit or pytest-benchmarking or something directly accessible from GitHub Actions may be more suitable. This could be worth tracking as a separate issue. 

**Additional context**
We discussed using a dataset that was >1GB in size as a real-world example. For the time being, we could use the data in the example and scale it up as needed.

a-smith-github:
Changes required for running profiling. Usage is 

`poetry run pytest tests/pmodel/test_profiling.py --profile-svg`

 I've not included the data file that the test uses. Will confer and consider best way of handling large file in the context of a cloud-based CI test

Fixes #81


davidorme:
For the large testing file, one option would be to put it on https://zenodo.org.  That would give us a DOI for the test file and Zenodo deposits are versionable, so we can link to most recent or individual files. We can then provide details to `curl` the test files for developers without needing LFS, but I don't know how that plays with the CI process on GitHub Actions?

mondus:
Another option for data is Git LFS. As we are under 2GB then this should be a solution even on the free tier (although expect Imperial RSE to be on Team subscription).

davidorme:
The run time on this has got to be 20 minutes or so. Is that something we want as part of the standard test suite for every push?

Can we make it run only on 'important' pushes - I was wondering if we can configure a separate `pytest` with profiling as a GithubAction job on pushing to `release` branches. That doesn't seem quite right - I think of `release` branches as minor tinkering and version bumping - not the place to identify and solve major performance issues...

davidorme:
We also need to doc the use of `lfs` in `CONTRIBUTING.md` - I've certainly made an awful mess using `lfs` in the past 😄

a-smith-github:
Thanks David, some really great points. Agree this doesn't need to run on every push. For benchmarking tests, you should run them periodically (e.g. every week) and get them to auto-create issues to investigate when a significant change in runtime occurs (I think GitHub can do that). For profiling or coverage, I'd recommend running those tests when the PR gets created or when review requests get sent out (again, I think GH can do this) so reviewers have that information to hand. You _could_ also run benchmarking at this point. Going forward it might be worth making a new test subfolder to run for benchmarking and profiling data and considering which tests (from all folders) we include in the coverage subset. 

Agree we should update contributing as part of this change before we merge it, should we chose to do so. It _should_ just be a case of making sure git-lfs gets installed on the developer's machine but...

davidorme:
@a-smith-github I should have thought of this earlier, but it might be worth looking at reducing the latitudinal ranges and keeping a longer time series. 

It's not a big deal - but once we've got a feel for what size of data works well it might be worth trialling a different slice. The main reason for this is that the spatial structure of the data doesn't matter for most functions, but there are some functions that iterate over the time dimension, so we can probably tune the dataset to provide better testing.

a-smith-github:
Thanks David, this might be a good plan going forwards. Instinctively, I'd be looking to do make the outer most loop / iterable smaller if I wanted a representative profile so whichever fits the bill is fine. 

My plan is to convert this PR into running periodically rather than on each push (and try to store the test artefacts) and consider splitting the the test into smaller tests as discussed yesterday. I think then we could move this into develop. You should keep the branch alive to pursue further work.

a-smith-github:
I've split the tests into some groupings and created a shared fixture. However there is a lot of overlap. Initially I tried to create a test for isotopes but it was too intertwined with the previous test to be beneficial (for the engineering effort).

Profiling tests could be moved into their own folder and run periodically rather than on every push

MarionBWeinzierl:
@a-smith-github , could you please make sure that the profiling process is documented?

MarionBWeinzierl (approved):

PR #91, opened 2023-08-22
Improved README and new CONTRIBUTING

davidorme:
This PR:

* Moves the bulk of `README.md` to an updated and clarified `CONTRIBUTING.md`.
* Updates `README.md` to provide a shorter simpler overview for both GitHub and PyPi and adds funder information on package development.

Closes #65
Closes #71 
Closes #57

PR #92, opened 2023-08-22
Fix doctests

### LINKED ISSUE #90: Broken `pytest` configuration doesn't run doctest
**Describe the bug**

The current configuration of `pytest` in `setup.cfg` does not run `doctest` on the package modules. Although `--doctest-modules` is set, the `testpaths` setting only includes `tests`, so none of the module code is tested.

**To Reproduce**

If `testpaths = tests` is changed to `testpaths = tests pyrealm` then the doctests are discovered, many of which fail.

davidorme:
This closes #90 by:

* reimplementing the doctests by editing `setup.cfg` to look in `tests` and `pyrealm`.
* fixing all the accumulated broken doctests, which are mostly just changed arg names and array not scalar return types.

### DIFF
--- a/pyrealm/hygro.py
+++ b/pyrealm/hygro.py
@@ -37,15 +37,15 @@ def calc_vp_sat(ta: NDArray, const: HygroConst = HygroConst()) -> NDArray:
         >>> # Saturated vapour pressure at 21°C
         >>> import numpy as np
         >>> temp = np.array([21])
-        >>> round(calc_vp_sat(temp), 6)
-        2.480904
+        >>> calc_vp_sat(temp).round(6)
+        array([2.480904])
         >>> from pyrealm.constants import HygroConst
         >>> allen = HygroConst(magnus_option='Allen1998')
-        >>> round(calc_vp_sat(temp, const=allen), 6)
-        2.487005
+        >>> calc_vp_sat(temp, const=allen).round(6)
+        array([2.487005])
         >>> alduchov = HygroConst(magnus_option='Alduchov1996')
-        >>> round(calc_vp_sat(temp, const=alduchov), 6)
-        2.481888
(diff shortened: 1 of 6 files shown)


davidorme:
Ack. I forked this branch off of an existing issue fix - not sure how I remove all those commits that are on #91

MarionBWeinzierl:
> Ack. I forked this branch off of an existing issue fix - not sure how I remove all those commits that are on #91

I'd recommend a rebase (as described [here](https://stackoverflow.com/questions/36168839/how-to-remove-commits-from-a-pull-request)), but you can also do a revert (also described in that post).

MarionBWeinzierl (approved):
LGTM

pyrealm/hygro.py
```diff
@@ -37,15 +37,15 @@ def calc_vp_sat(ta: NDArray, const: HygroConst = HygroConst()) -> NDArray:
...
         >>> temp = np.array([21])
-        >>> round(calc_vp_sat(temp), 6)
-        2.480904
+        >>> calc_vp_sat(temp).round(6)
+        array([2.480904])
         >>> from pyrealm.constants import HygroConst
         >>> allen = HygroConst(magnus_option='Allen1998')
-        >>> round(calc_vp_sat(temp, const=allen), 6)
-        2.487005
+        >>> calc_vp_sat(temp, const=allen).round(6)
```
a-smith-github, line 44: Example of where the help text misses the import. Although I suspect it's likely that the user could recover (user needs to understand the context from which they are accessing the method i.e. maybe they should use pyrealm.hygro.calc_vp_sat(args).

a-smith-github (commented):
Looks really good David, just one suggestion about considering the import text in the examples (or at least one of them anyway).

PR #97, opened 2023-08-29
Adds minimum example to README

### LINKED ISSUE #95: Create minimum QuickStart example
It is useful to have a minimum QuickStart example in the ReadMe, which lets you test your installation and gives you a first idea of how the software works and what the inputs and outputs are.

This minimum example could potentially be deduced from the SPLASH model  (see #69). 

As a first, not-quite-as-useful-but-ok step, single functions could be called.

davidorme:
This adds a simple example to the README to allow users to verify their installation and get started with `pyrealm`. It fits a relatively complicated thing (the P Model) but for a single site, so is compact and runs fast.

Closes #95

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,30 @@ The `pyrealm` package requires Python 3.9 or greater and can be installed from
 pip install pyrealm
 ```
 
-At present, the package is designed for use with Python scripts or Jupyter notebooks.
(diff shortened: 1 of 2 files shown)


MarionBWeinzierl (approved):
Thanks for the change, looks good to me

PR #100, opened 2023-09-05
Update of water density and viscosity functions

### LINKED ISSUE #96: Give the option to choose between different functions in calc_viscosoty_h2o
In the profiling ( #81 ) it was found that one bottleneck seems to be the function https://github.com/ImperialCollegeLondon/pyrealm/blob/fcbba0eead6857fd412522ff7299af03c4d31b95/pyrealm/pmodel/functions.py#L671 .

@davidorme explains that there are various, sometimes very simple alternatives for this calculation. 

It would be useful to implement the option to switch between different variants. These could be profiled separately, and the user can then be given the option to trade speed against precision.

davidorme:
Following from the discussion on issue #96 , the current implementation of `calc_density_h2o` and `calc_viscosity_h2o` are slow (and probably memory hungry, although I haven't checked that) as a result of using 'fancy' numpy calls to do outer matrix multiplication and using `np.power` to calculate arrays of powers. Turns out neither of these things is as efficient as I'd though.

This PR:

* Replaces the matrix methods with simpler loops and the power calculations with `tc*tc*tc...` and `tc**n`.
* Adds a new method for water density (Chen et al. 2008), used in some previous applications and added here for benchmarking.
* Converts `calc_density_h2o` into a wrapper around the available individual water density functions.

This closes #96 but possibly only for now - we are still missing well documented really simple methods for these two, although it isn't clear that there _is_ actually a reasonably precise simple function.

### DIFF
--- a/pyrealm/constants/pmodel_const.py
+++ b/pyrealm/constants/pmodel_const.py
@@ -130,6 +130,9 @@ class PModelConst(ConstantsClass):
     k_CtoK: float = 273.15
     """Conversion from °C to K   (:math:`CtoK` , 273.15, -)"""
 
+    water_density_method: str = "fisher"
+    """Set the method used for calculating water density ('fisher' or 'chen'."""
+
     # Fisher Dial
     fisher_dial_lambda: NDArray[np.float32] = np.array(
         [1788.316, 21.55053, -0.4695911, 0.003096363, -7.341182e-06]
--- a/pyrealm/constants/pmodel_const.py
+++ b/pyrealm/constants/pmodel_const.py
@@ -17,11 +17,17 @@ class PModelConst(ConstantsClass):
     value and units shown in brackets and the sources for default parameterisations are
     given below:
 
-    * **Density of water**. Values for the Tumlirz equation taken from Table 5 of
-      :cite:t:`Fisher:1975tm`:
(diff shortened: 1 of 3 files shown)


pyrealm/constants/pmodel_const.py
```diff
@@ -124,6 +130,9 @@ class PModelConst(ConstantsClass):
     k_CtoK: float = 273.15
     """Conversion from °C to K   (:math:`CtoK` , 273.15, -)"""
 
+    water_density_method: str = "fisher"
```
MarionBWeinzierl, line 133: Rather than setting the method here in the class definition (and changing the definition when you want to change just this parameter), it would be better to give it as an argument to the constructor.

davidorme: I don't follow. The current setup is that you'd switch using:

`calc_density_h20(tc, patm)`

to

`calc_density_h20(tc, patm, const=PModelConst(water_density_method="fisher")`

The reason for packaging those constant settings is that it is then easier to pass around a single set of shared settings, but I agree that the signatures of some functions are then a bit odd.

MarionBWeinzierl: Apologies, my fault, I misread what you are doing.

PR #108, opened 2023-09-15
Building out better structured data package of golden outputs and example data.

### LINKED ISSUE #24: Use `pytest_datadir` in tests
The test data files are a bit scattered around the place. Swap to using `pytest-datadir` to better structure files and make them easier to load:

https://pypi.org/project/pytest-datadir/

### LINKED ISSUE #60: Remove R code from repo or integrate its execution (with documentation).
There are several "golden output" datasets currently embedded in the `tests` directory. These most logically belong in the `pyrealm_build_data` `sdist` package and indeed a number of bits of test data are already stored in there.

The code, inputs and outputs should probably moved into sensible directory structures in `pyrealm_build_data` along with better documentation and clear instructions on how to regenerate the outputs.

One issue here - I do not really want to make `R` a build dependency, so I'd be tempted to just simply document that developers will need to install R and the `rpmodel` package for that dataset.

davidorme:
This PR:

* Tidies up the `tests` folder by moving data files and code that generate golden outputs from original implementations into the `pyrealm_build_data` package.
* Structures `pyrealm_build_data` into submodules to organise those files.
* Uses `importlib.resources.files("pyrealm_build_data.submodule", "data_file.xyz`)` to reference those files in testing and in the `sphinx` docs.
* Updates the developer links in the docs to include a description of the `pyrealm_build_data` package contents.
* Updates the developer contributing page in `docs` to mirror `CONTRIBUTING.md`.

There are some things that need fixing - the first should be fixed here, but I think we leave the second for now.

1. I'm `include`ing the `CONTRIBUTING.md` contents in `docs/source/development/developers.md` verbatim to avoid duplication of what would basically be identical content, but that breaks some links which make sense in the location of `CONTRIBUTING.md` but not in `docs`. I can't see a sensible solution apart from duplication or making `CONTRIBUTING.md` a simple pointer to the ReadTheDocs, which I don't like as much.

1. The `pyrealm_build_data` contents description includes requirements and shell commands for running generating _some_ of the benchmark datasets but not all the examples are complete:
    * The `subdaily` submodule requires the original R code and some settings, but these are in a separate repo.
    * The `uk_data` submodule was generated from input datasets on fixed paths and those (large) datasets and paths are not replicated here. 
    * The requirements to run these commands are not yet part of the package requirements. These are largely `R` and a couple of reference packages. We can either just document the requirements as here or possible extend `pyproject.toml` to include an optional group for running these scripts. I lean towards the first solution as these _really_ aren't expected to change frequently, if ever.

This closes #24 and #60.


docs/source/development/pyrealm_build_data.md
```diff
@@ -0,0 +1,129 @@
...
+requires an installation of `R` and the `rpmodel` package and can be run as:
+
+```sh
+Rscript test_outputs_rpmodel.R
+```
+
+### Global array test
+
+The remaining files in the submodule are intended to provide a global test dataset for
+benchmarking the use of `rpmodel` on a global time-series - so using XYT data. It is
```
MarionBWeinzierl, line 84: What is XYT data? Does that mean 2D data plus time?

tests/pmodel/test_pmodel_global_array.py
```diff
@@ -1,6 +1,9 @@
 """Runs tests of the P Model using global arrays as inputs.
 
 This is used to benchmark the pyrealm predictions against the rpmodel implementation.
+
+TODO - this test is currently broken because there is an issue with calculating the
```
MarionBWeinzierl, line 5: Should this then be removed, rather than leaving something outdated and broken on here?

MarionBWeinzierl (approved):
Looks ok, as far as I can judge

docs/source/development/pyrealm_build_data.md
davidorme: It does, but I'll clarify.

PR #111, opened 2023-09-20
56 check requirements and ensure that these are consistent between pyprojecttoml and poetry

### LINKED ISSUE #56: Check requirements and ensure that these are consistent between pyproject.toml and poetry.
Intention is to move over to entirely do this within poetry. There are a few dependencies in the environment that are not required as a result of tinkering with libraries.

MarionBWeinzierl:
Closes #56 .

Removes requirements.txt and removes Bottleneck from the requirements


davidorme (approved):
LGTM - as noted in #56, we may want to go from an explicit `NetCDF4` to `xarray[io]` if we need to expand the range of handled file types.

PR #136, opened 2023-10-18
Split pmodel file

### LINKED ISSUE #58: Consider restructuring source files to remove multiple classes from a single source file.
Notably, `pyrealm/pmodel/pmodel.py` defines `PModelEnvironment`, `PModel`, `CalcOptimalChi` and `JMaxLimitation`.

Splitting these up into separate files makes a lot of sense. The only question here is the timing of doing this. I think the revision of the subdaily model to allow C4 plants will involve a refactor of `CalcOptimalChi`, possibly with the definition of an abstract base class for the various calculation options. That should involve splitting this file up and is a development priority, so this can likely wait.

MarionBWeinzierl:
Fixes #58 . The pmodel module is split so that each class has it's own file. The docstrings and documentation are adapted accordingly.

@davidorme , could you double-check that I didn't miss or break anything in the docs?


davidorme (changes requested):
@MarionBWeinzierl That looks great, thanks.

The only question I have (and this is very fussy) is that PEP8 recommends [module names be lower case](https://peps.python.org/pep-0008/#package-and-module-names)? I do think its a bit clearer if CamelCase is reserved for classes.  I'd probably make the modules: `calc_optimal_chi.py`, `jmax_limitation.py`, `pmodel.py` and `pmodel_environment.py`.

Thoughts?

MarionBWeinzierl:
> @MarionBWeinzierl That looks great, thanks.
> 
> The only question I have (and this is very fussy) is that PEP8 recommends [module names be lower case](https://peps.python.org/pep-0008/#package-and-module-names)? I do think its a bit clearer if CamelCase is reserved for classes. I'd probably make the modules: `calc_optimal_chi.py`, `jmax_limitation.py`, `pmodel.py` and `pmodel_environment.py`.
> 
> Thoughts?

Hm, good point. I had used camel case as each file contains now one class, so I took the files as classes. But it would probably be similarly valid to say they are modules containing one class.

davidorme:
That makes a lot of sense, but at least one of them (`CalcOptimalChi.py`) is likely to expand to be provide an abstract base class and then some subclasses of it. I wouldn't rule out putting other small things in these files if it seems like an appropriate home.

MarionBWeinzierl:
> That makes a lot of sense, but at least one of them (`CalcOptimalChi.py`) is likely to expand to be provide an abstract base class and then some subclasses of it. I wouldn't rule out putting other small things in these files if it seems like an appropriate home.

OK, I made the changes

davidorme (approved):
Looks good to me. There are some places where we could leave original import paths, because the components are all imported into `pmodel.__init__.py`, but I don't think there's any advantage to using the flattened namespace over the original locations?

So for example `from pyrealm.pmodel.pmodel_environment import PModelEnvironment` could also stay as `from pyrealm.pmodel import PModelEnvironment`, but I don't think it matters.

MarionBWeinzierl:
> Looks good to me. There are some places where we could leave original import paths, because the components are all imported into `pmodel.__init__.py`, but I don't think there's any advantage to using the flattened namespace over the original locations?
> 
> So for example `from pyrealm.pmodel.pmodel_environment import PModelEnvironment` could also stay as `from pyrealm.pmodel import PModelEnvironment`, but I don't think it matters.

Hi, ok, if you prefer that we can do that (I had gotten so many "not found" errors that I changed those to be on sure). Could you have a look at the "Files changed" tab and check whether I found  all the chances, or whether I missed something/made a mistake?

MarionBWeinzierl:
I asked colleagues, and basically got the answer "flattened namespaces are good, probably no performance issue" as well as "explicit is better than implicit". I made the decision to go back to the flattened namespace, but am now getting lots of error, presumable due to an issue like this: https://stackoverflow.com/questions/59762996/how-to-fix-attributeerror-partially-initialized-module , as pmodel is both the module name and a submodule/class name.

MarionBWeinzierl:
Ha. I resolved the problem by reordering the imports. Then I committed, and the precommits ran, and isort failed and changed the order back!

davidorme:
Good old `isort` - still I don't think having the import process being sensitive to the ordering of the imports within a module sounds robust. I think that being explicit about the imports within the modules should fix that though? Then the module `__init__` provides a flattened namespace but the submodules still point to the correct home?

davidorme:
Before, all of the classes were imported from the same module, so that modules didn't need to contain imports for anything included in the flattened namespace.

MarionBWeinzierl:
Yeah, that will fix it. So I'll revert what I've done today.

PR #139, opened 2023-11-07
Create core module

### LINKED ISSUE #75: Restructure shared functions into a common core

The draft `splash` module (#69) has a bunch of functions that overlap substantially with existing functions in the `pmodel` module. The two sets of functions need to be reconciled but also relocated - they are general utility functions that are shared across modules, so logically are better placed outside those specific science modules. As a precursor to the integration of `splash` it makes sense to have a more logical home for that common code.

Create a `pyrealm.core` module that incorporates these functions in sensible groupings. That probably also brings in the `hygro` and `utilities` modules, to give `pyrealm.core`, `pyrealm.const` and then particular science/model based modules like the `pmodel`, `tmodel`, `splash` etc.

There are enough of these functions that just having small modules for each one would make a more cluttered namespace. Gathering them as submodules under a single module seems cleaner.

Edited to add subtasks (still in progress):

- [x] Create `core` module.
- [x] Move `utilities` into `core` and fix references and documentation.
- [x] Create a new `core` documentation section to describe the functionality in the `core` module
- [x] Move `hygro` module into `core` and fix references and documentation.
- [x] Constants restructure - base constants class definition should be `core` and `hygroConsts` should move with `hygro` module (maybe move _into_ `hygro`), but then other classes move into the appropriate science modules? We then need to unpack `PModelConsts`. Not sure about this one yet! (This is #76 .)
- [x] Inspect `pmodel.functions` to separate out `core` from `pmodel` functions.
- [ ]

### LINKED ISSUE #76: Restructure constant classes to split into more module specific groups.

The original constants class for the PModel (`pyrealm.constants.pmodel_const.PModelConst`) is bloated by a whole bunch of constants and parameters for functions that aren't purely PModel specific. For example the parameterisation of the density of water calculation is in there, and that functionality isn't PModel specific - particularly now that the `splash` draft (#69) also requires that calculation.

The existing constants classes need to be revised to break them into more logical groupings. That will also involve updating some functions and classes to require extra constants classes in order to be able to pass in modified values.

There isn't an obvious alternative - everything got shoved in `PModelConst` in the early implementations and that is not a good long term organisation.

MarionBWeinzierl:
This PR closes #75 . It creates the described core module and moves the utilities and hygro modules as submodules into the core. It also adapts the documentation to reflect these changes.

Updated by @davidorme:

I think we can now move this out of draft status and review it:

From #75:
* The `core` module is now created and includes `utilities`, `hygro`, `water` and `pressure`.
* Those last two submodules are the general functions that were in `pmodel.functions`.
* There is now a `CoreConst` constants class to cover `core` module functions. This also closes #76.
* The `pmodel` module method and function signatures are updated to add `core_const` where required, and then their code is updated to use the correct constant instances.
* There is a (minimal) `api` entry in the docs to expose the docstrings on the core functionality.


davidorme:
Looks good so far!

davidorme:
I left some tests in `pmodel` that should properly now be in `tests\unit\core`. I had forgotten that #151 has already been merged into develop, so was leaving them until they could be moved into the new structure, but they can so should be here.

docs/source/users/pmodel/subdaily_details/subdaily_calculations.md
```diff
@@ -20,7 +20,7 @@ steps used in the estimation process in order to show intermediates results but
 practice, as shown in the [worked example](worked_example), most of these calculations
 are handled internally by the model fitting in `pyrealm`.
 
-```{code-cell}
+```{code-cell} python python
```
MarionBWeinzierl, line 23:
```suggestion
```{code-cell} python
```

MarionBWeinzierl (commented):
Looks good to me, apart from one small  typo

tztsai (approved):
Looks good to me!

Issue #143, opened 2023-11-17
Split tests into unit tests and regression tests


MarionBWeinzierl:
Currently, various tests and even profiling (see #141) are mixed in the testing directory. It would make sense to separate them. This would provide the possibility to run them separately (to be discussed when and if this makes sense), but it also would give a clearer structure to the tests folder.

Issue #142, opened 2023-11-17
Add profiling to CI - Performance regression testing


MarionBWeinzierl:
In #81 it was considered to add profiling to CI. However, it should not be run for every push to develop, but after larger changes, or for new releases. This could be for a push to main, for example.

We need to think about how we make sure that we catch performance deterioration, as it is not a build-in feature on Github. We could look at the output manually, or we could keep track of performance in a separate file, or we could hard-code a runtime limit (and adapt it manually). Are there any better ways to do this?

MarionBWeinzierl:
Another option would be to run against the develop/main branch performance

Issue #141, opened 2023-11-17
Modularise and improve profiling


MarionBWeinzierl:
As described in #81 , the profiling should be pulled out of the general testing, and in particular the profiling directory, and given its own module. It should then be extended to not only cover the pmodel. We should consider how we want to potentially automate things like creating the call graph etc.

Tasks:
- [x] Make profiling its own module. 
- [x] Extend profiling beyond pmodel.
- [x] Automate call graph creation.
- [x] Consider additional profiling outputs and how to collect them (automatically).
- [x] Adapt and extend the documentation about profiling in the CONTRIBUTING.md file.

tztsai:
I think there could be two solution:

1. The first one would be to create a new folder `profiling` in the root directory and move tests/pmodel/test_profiling.py to profiling/test_pmodel_profiling.py. And we also create profiling scripts for the T Model and the Hygro model, as in the following picture:

<img width="1089" alt="Screenshot 2023-11-28 at 2 09 45 PM" src="https://github.com/ImperialCollegeLondon/pyrealm/assets/34415979/15041b0f-3d97-4d40-8ad1-66357c6d256b">

Consequently, we need to update .github/workflows/pyrealm_ci.yaml to run the profiling separately (although this should be addressed later in issue #142), somewhat similar to this:

```yaml

  test:
    ...

    steps:
    - ...

    - name: Profiling
      if: github.ref == 'refs/heads/main'
      run: poetry run pytest profiling/ --profile
```

2. The second one would be to add a `conftest.py` in the tests/ folder with the following content:
<img width="1087" alt="image" src="https://github.com/ImperialCollegeLondon/pyrealm/assets/34415979/7e1d876a-6ff3-47d2-b338-ddc412ccfb69">
This will make all classes or functions decorated by `pytest.mark.profiling` skipped during testing unless the argument `--profiling` is provided. An example of marking the tests in tests/pmodel/test_profiling would be
<img width="1180" alt="image" src="https://github.com/ImperialCollegeLondon/pyrealm/assets/34415979/589ea236-175a-4ffb-8ab2-b04785606f56">

Consequently, we also need to update .github/workflows/pyrealm_ci.yaml later, somewhat similar to this:

```yaml

  test:
    ...

    steps:
    - ...

    - name: Run tests with profiling
      if: github.ref == 'refs/heads/main'
      run: poetry run pytest --cov-report xml --profile

   - name: Run tests without profiling
[...]

PR #149, opened 2023-11-28
Add small dataset for profiling

### LINKED ISSUE #140: Use smaller data set for profiling
The current profiling functionality relies on [a rather larger data set](https://github.com/ImperialCollegeLondon/pyrealm/blob/develop/pyrealm_build_data/inputs_data_24.25.nc) (1.8GB) and then uses just a manageable slice. See also https://github.com/ImperialCollegeLondon/pyrealm/blob/ad32a8eff8ea42ded2a6cefc0d33880e8560332b/tests/pmodel/test_profiling.py#L27 .

This needs some looking into. 

Options to consider:
- Keep the large data set and define a minimum slice that still gives reasonable results but reduces runtime.
- Find/build a smaller data set.
- Create a smaller synthetic data set (which might not even need be populated with realistic values).

### LINKED ISSUE #152: Call graph is not generated in CI given the cmd option `--profile-svg` even though `graphviz` is installed.
**Describe the bug**
image

tztsai:
# Description

- Updated `CONTRIBUTING.md` with information for profiling.
- Added a 50MB dataset for profiling.

## Type of change

## Key checklist

## Further checks


davidorme:
Oh hang on - wait. You're intercepting `pytest --profile` and setting the test collection from that.

tztsai:
CLI Interface:
- `poetry run pytest --profile/--profile-svg` will run tests marked by `profiling` or `profiling_only`.
- `poetry run pytest` will run tests marked by `profiling` or not marked by either markers.

davidorme:
I'm just not sure we need to tie the test collection process to the setting of the profiling option. With one mark we can do `pytest -m "profiling" --profile` as the most common use case to profile only the specific profiling tests, but also run `pytest --profile` or `pytest -m "not profiling" --profile`.

Most of the time, we're going to want to run `pytest -m "not profiling"` to simply run the unit testing to check a commit hasn't broken stuff, but we might want to profile more than just the specific profiling suite. It also occurs to me that once #143 is closed, we can also mark tests as 'regression' and maybe exclude those except on PR. But that's getting ahead of myself.

tztsai:
> I'm just not sure we need to tie the test collection process to the setting of the profiling option. With one mark we can do `pytest -m "profiling" --profile` as the most common use, case to profile only the specific profiling tests, but also run `pytest --profile` or `pytest -m "not profiling" --profile`.
> 
> Most of the time, we're going to want to run `pytest -m "not profiling"` to simply run the unit testing to check a commit hasn't broken stuff, but we might want to profile more than just the specific profiling suite. It also occurs to me that once #143 is closed, we can also mark tests as 'regression' and maybe exclude those except on PR. But that's getting ahead of myself.

OK, I think it's simpler to just use the markers together with command line arguments, so I can remove the `conftest.py`. Accordingly, I can update `.github/workflows/pyrealm_ci.yaml` to run profiling when there is a push to the `main` branch.

davidorme:
Would it be cleaner for `profiling.py` to return a value to an environment variable rather than monitoring `sterr`?

tztsai:
> Would it be cleaner for `profiling.py` to return a value to an environment variable rather than monitoring `sterr`?

I agree. I have updated the script accordingly.

davidorme:
> > Would it be cleaner for `profiling.py` to return a value to an environment variable rather than monitoring `sterr`?
> 
> I agree. I have updated the script accordingly.

That wasn't quite what I meant, but I was unclear. My idea was that `profile.py` uses `sys.exit(0)` or `sys.exit(1)` to return a success code. Then the `profiling` step can capture the resulting exit code (maybe using `profiling_pass=$?`) and use that to control the next steps.

Practically that works in a very similar way but it is clearer in  the `profiling` code that the `profiling_pass` variable is the outcome of an exit code. At the moment that `steps.profiling.outcome == 'success'` requires inspection of the `profile.py` code to understand where it comes from. It feels more canonical bash this way but this is picky!

tztsai:
> > > Would it be cleaner for `profiling.py` to return a value to an environment variable rather than monitoring `sterr`?
> > 
> > 
> > I agree. I have updated the script accordingly.
> 
> That wasn't quite what I meant, but I was unclear. My idea was that `profile.py` uses `sys.exit(0)` or `sys.exit(1)` to return a success code. Then the `profiling` step can capture the resulting exit code (maybe using `profiling_pass=$?`) and use that to control the next steps.
> 
> Practically that works in a very similar way but it is clearer in the `profiling` code that the `profiling_pass` variable is the outcome of an exit code. At the moment that `steps.profiling.outcome == 'success'` requires inspection of the `profile.py` code to understand where it comes from. It feels more canonical bash this way but this is picky!

But if the program exits with code 1, `steps.profiling.outcome` will not be 'success', and the workflow as is will not save the profiling reports. However I am exploring the `continue-on-error` setting for this.

MarionBWeinzierl:
As discussed in the dev meeting today, this PR should be split up into two PRs: One deadline with #141 (and #152)  and one dealing with #142 .

tztsai:
> As discussed in the dev meeting today, this PR should be split up into two PRs: One deadline with #141 (and #152) and one dealing with #142 .

I have moved most of my updates to the new PR #156, which is ready to be merged. I will subsequently work on this PR for modularising the profiling and updating the LFS dataset.

.github/workflows/pyrealm_ci.yaml
```diff
@@ -48,12 +48,12 @@ jobs:
...
-      if: ${{ steps.codecov.outcome == 'success' && github.event_name != 'push' }}
-      run: |
-        git lfs pull
-        git lfs ls-files -n | xargs git lfs track  # track all LFS files
+    # - name: Setup Git LFS
+    #   id: lfs
+    #   if: ${{ steps.codecov.outcome == 'success' && github.event_name != 'push' }}
+    #   run: |
+    #     git lfs pull
+    #     git lfs ls-files -n | xargs git lfs track  # track all LFS files
```
davidorme, line 56: Can you delete rather than comment out? I used to comment retired options too - so that I had a reminder of how to do them - but we have this in the git history if we need it and keeping the files clean is better practice.

.gitignore
```diff
@@ -25,8 +25,12 @@ pyrealm.egg-info
 # Coverage
 .coverage
 
+# mypy
+.mypy_cache/
+
 # Data
-pyrealm_build_data/*.nc
+pyrealm_build_data/inputs_data_24.25.nc
+pyrealm_build_data/eda.py
```
davidorme, line 33: Is `eda.py` the subsetting script? I'm not quite sure what the best thing to do here is. In an ideal world, we'd be able to regenerate the input file completely - but I'd need to add all the download scripting to assemble the file we were using. Ultimately, I think that is where an artificial input would be cleaner.

davidorme (changes requested):
A couple of minor issues/questions but otherwise this looks good.

.gitignore
tztsai: Yes, that's for excluding my local script to analyze and subset the dataset. I can write a bash script to regenerate the input file from the LFS file if it's better.

.github/workflows/pyrealm_ci.yaml
tztsai: I have deleted it now.

davidorme (approved):
LGTM

PR #151, opened 2023-11-30
Split the tests directory into unit and regression

### LINKED ISSUE #143: Split tests into unit tests and regression tests

surbhigoel77:
# Description

Restructured the tests directory and split it into unit and regression modules. The segregation would facilitate a structured execution of the tests and would also provide easier navigation. 
Also added a README.md to empty folders to make them trackable by git.

Fixes #143

## Type of change

## Key checklist

## Further checks


MarionBWeinzierl (commented):
This looks generally good for me. I am just wondering whether we want to use these as packages, in which case we need a __init__.py ( see https://docs.python.org/3/tutorial/modules.html , section 6.4). If we don't need it at the moment or in the foreseeable future it's fine as it is.

Issue #23, opened 2022-12-12
Add unit tests


davidorme:
The bulk of the unit testing at the moment is validation of the `pyrealm` predictions against the predictions of other implementations. The test suites need more simple tests checking that the basic classes and functions work as intended - clearer checking of failure modes.

MarionBWeinzierl:
Start to add tests when errors come up, add unit tests

MarionBWeinzierl:
Split tests into unit tests vs regression/integration tests?

davidorme:
Closing as done.

PR #153, opened 2023-12-06
Add unit tests

surbhigoel77:
# Description

Fixes Issue #23
Fixes Issue #191 

In this PR, we are adding new unit tests for the sub-modules in pmodel that do not have dedicated unit tests. These are the submodules that we are writing tests:

- [x] optimal_chi
- [x] pmodel_environment
- [x] jmax_limitation

Current work in progress is  for `optimal_chi` has following testcases:
- [x] 1. Test `set_beta` setup
- [x] 2. Test `chi` estimation
- [x] 3. Test `subclass` functioning for all methods
- [x] 4. Test `NaN` handling 

Current work in progress is  for `jmax_limitation` has following testcases:
- [x] 1. Test `jmax_limitation` setup
- [x] 2. Test `value error` for invalid method

Current work in progress is  for `pmodel_environment` has following testcases:
- [x] 1. Test `pmodel_environment` setup
- [x] 2. Test `pmodel_environment` generating any abrupt values for 'kmm' 
- [x] 3. Test `pmodel_environment` generating any abrupt values for 'ns_star' 
- [x] 4. Test `pmodel_environment` generating any abrupt values for 'ca' 
- [x] 5. Test `pmodel_environment` generating any abrupt values for 'gammastar' 

## Type of change

## Key checklist

## Further checks


surbhigoel77:
Current setup for submodules with their unit tests in the `pmodel` module:

1. competition -> `test_c3c4competition`
2. functions -> `test_functions`
3. isotopes -> `test_calc_carbon_isotopes`
4. subdaily -> `test_memory_effect`, `test_fast_slow_pmodel`, FastSlowPModel_JAMES
5. fast_slow_scaler -> `test_fast_slow_scaler`
6. cal_optimal_chi -> UT not available
7. jmax_limitation -> UT not available
8. pmodel_environment -> UT not available

davidorme:
@surbhigoel77 I'm not quite sure what to review here? I only see the one line.

surbhigoel77:
> @surbhigoel77 I'm not quite sure what to review here? I only see the one line.

I have mentioned the existing unit tests for present submodules in pmodel. For some of these we already have unit tests. For some, we don't and I have commented 'UT not available' for those. I want a heads-up from you to start writing tests for these submodules.

tests/unit/pmodel/test_optimal_chi.py
```diff
@@ -0,0 +1,182 @@
...
+    argnames=["subclass", "pmodelenv_args", "expected"],
+    argvalues=[
+        (
+            OptimalChiPrentice14,
+            dict(tc=20, patm=101325, co2=400, vpd=1000),
+            dict(chi=0.69435, mc=0.33408, mj=0.7123, mjoc=2.13211),
+        ),
+        (
+            OptimalChiPrentice14RootzoneStress,
+            dict(tc=20, patm=101325, co2=400, vpd=1000, rootzonestress=0.5),
```
davidorme, line 89: Some rough bounds on these input variables given in the bounds checking here: https://github.com/ImperialCollegeLondon/pyrealm/blob/develop/pyrealm/pmodel/pmodel_environment.py#L91-L99 (and below for theta and rootzonestress).

As we discussed, the issue is that remotely sensed inputs have all sorts of odd values in them, so `pyrealm` is permissive about allowing extreme values but does issue warnings in an attempt to alert users. There are some more details here: https://pyrealm.readthedocs.io/en/latest/users/pmodel/pmodel_details/extreme_values.html#realistic-input-values

More concretely:
* `co2` and `patm` _cannot_ be negative. 
* `theta` and `rootzonestress` must be in `[0,1]`.

In the outputs:
* `beta` and `xi` must be positive.
* `mc` and `mj` must be in `[0,1]` and hence `mjoc = mj/mc` must be positive.
* `ci` must be positive and `chi` must be in `[0,1]`.

I'm vaguely wondering whether `pyrealm` should add a second layer of value checking to eliminate _impossible_ inputs. So `bounds_checker` might implement hard and soft bounds, where exceeding the hard bound is physically impossible but exceeding the soft bounds is just a bit unlikely.

davidorme: Let me know if you need more details - sorry for the delay in getting this to you.

davidorme (commented):
LGTM - very clean.  There does seem to be duplication between the two files - is one going to be retired?

surbhigoel77:
> LGTM - very clean. There does seem to be duplication between the two files - is one going to be retired?

@davidorme thanks for pointing out. The one with the name `test_calc_optimal_chi` is the outdated one and shall be removed.

tests/unit/pmodel/test_optimal_chi.py
surbhigoel77: @davidorme As I checked,
 - `bounds_checker` does issue warning for out-of-bound values and 
 - `bounds_mask` replaces the out-of-bound values with `np.nan`
  
But how these nan values are further handled is not clear. When I included the the testcase including nan values, the unit test for optimalchi failed. So maybe we can look for an appropriate method of handling the nan values as they may create issues in calculations in future.

surbhigoel77: I am writing the testcase to check the performance of `bounds_checker` for `Optimal_chi`. That could possibly guide us on whether the second layer of bounds_check is required.

surbhigoel77:
@davidorme Could you share the soft-bound values for the inputs to `OptimalChi`? I will include a unit testcase for the module based on these bounds.

surbhigoel77:
Thanks for sharing this @davidorme. Yes, using the actual test set is a decent idea. 

But, what is the ideal response to crossing the soft bounds?

surbhigoel77:
@davidorme could you review this PR?

There are 3 files to review as mentioned in the PR description.  A couple of special notes:
1. `test_OptimalChi` has a testcase for nan value check but is commented out as it would not let the github checks pass the file. The testcase gives assertion error.
2. `test_pmodelenvironment` has the testcases to check for photosynthesis variables' plausible out of bound values. I have used the available test dataset (tc, patm, co2) to verify the same and none of the variables (kmm, ns_star, gammastar, ca) have any out-of-bound values. (These testcases are linked to #191 )

tests/unit/pmodel/test_optimal_chi.py
```diff
@@ -0,0 +1,182 @@
...
+    OptimalChiC4,
+    OptimalChiC4NoGamma,
+    OptimalChiC4NoGammaRootzoneStress,
+    OptimalChiC4RootzoneStress,
+    OptimalChiLavergne20C3,
+    OptimalChiLavergne20C4,
+    OptimalChiPrentice14,
+    OptimalChiPrentice14RootzoneStress,
+)
+from pyrealm.pmodel.pmodel_environment import PModelEnvironment
```
davidorme, line 17: In general, I've a preference for importing the `pyrealm` components within tests. I've been told this is better practice, because it isolates the import process within the test itself, so the import becomes part of the test and only one test fails if an import fails. @MarionBWeinzierl and @mondus, is that right?

Having said that, this is _wildly_ inconsistent in the code base at the moment. I learned this part way through writing the existing `pyrealm` tests 😄 , so it might be better just to close this PR as is and then have a new issue to regularise the import locations.

tests/unit/pmodel/test_optimal_chi.py
```diff
@@ -0,0 +1,182 @@
...
+        OptimalChiC4RootzoneStress,
+        OptimalChiLavergne20C3,
+        OptimalChiLavergne20C4,
+        OptimalChiC4NoGamma,
+    ],
+)
+def test_set_beta(optimal_chi_class, photo_env):
+    """Test that beta is set correctly."""
+    optimal_chi_instance = optimal_chi_class(env=photo_env)
+    optimal_chi_instance.set_beta()
```
davidorme, line 49: Importing within the test would need something like using strings in the parameterisation and then:

```python
import pyrealm.pmodel.optimal_chi as oc_mod
chi_class = getattr(oc_mod, optimal_chi_class)
```

tests/unit/pmodel/test_pmodel_environment.py
```diff
@@ -0,0 +1,149 @@
...
+    "tc, patm", [(tc, patm) for tc in tc_ar_values for patm in patm_ar_values]
+)
+def test_out_of_bound_output(tc, patm):
+    """Function to calulate kmm."""
+    pmodel_const = PModelConst(
+        bernacchi_kc25=39.97,
+        bernacchi_ko25=27480,
+        bernacchi_dhac=79430,
+        bernacchi_dhao=36380,
+    )
```
davidorme, line 73: You shouldn't need to set these default values: `pmodel_const = PModelConst()` is identical.

tests/unit/pmodel/test_pmodel_environment.py
```diff
@@ -0,0 +1,149 @@
...
+
+# Test values of forcing variables as input to functions
+tc_ar_values = data["tc_ar"]
+patm_ar_values = data["patm_ar"]
+co2_ar_values = data["co2_ar"]
+
+
+""""Test that kmm output is within bounds"""
+kmm_lower_bound = 0
+kmm_upper_bound = 1000
```
davidorme, line 60: These should probably go within the test itself - no particular need to expose them as local variables?

tests/unit/pmodel/test_optimal_chi.py
```diff
@@ -0,0 +1,182 @@
...
+#             dict(chi=0.31305, mc=0.21583, mj=1.0),
+#         ),
+#     ],
+# )
+# def test_nan_handling(pmodelenv_args, subclass, expected):
+#     """Test that subclasses handles NaNs correctly."""
+#     env = PModelEnvironment(**pmodelenv_args)
+#     instance = subclass(env)
+#     for key, value in expected.items():
+#         assert getattr(instance, key) == pytest.approx(value, rel=1e-3)
```
davidorme, line 182: We'd expect an assertion error here - the `np.nan` is propagated through the calculations so that the resulting expected values are also `np.nan`. It would be a serious problem if that didn't happen, so asserting that each attribute is `np.nan` is a good guard.

I'd actually be tempted to extend this and feed in `pmodelenv_args` as complete data and then have a loop that copies the dict and sets each value in turn to `np.nan`. That way we know that missing data in any input results in missing data. Something like:
```suggestion
def test_nan_handling(pmodelenv_args, subclass, expected):
    """Test that subclasses handles NaNs correctly."""
    for var in pmodelenv_args:
        pmodelenv_args_copy = pmodelenv_args.copy()
        pmodelenv_args_copy[var] = np.nan
        env = PModelEnvironment(**pmodelenv_args_copy)
        instance = subclass(env)
        for key, value in expected.items():
            assert getattr(instance, key) == pytest.approx(value, rel=1e-3)
```

davidorme (changes requested):
Look goods to me - some minor structural stuff and a comment on the failing `np.nan` tests. The `import` location stuff is something we should sort out - it would be good to have this fixed here, but it might be more consistent to do a wider tidy up of the inconsistent behaviour, so I don't mind if those changes get parked.

tests/unit/pmodel/test_pmodel_environment.py
```diff
@@ -0,0 +1,149 @@
...
+"""Testing the boundries of variables kmm,gammastar, ns_star, co2_to_ca."""
+
+with open("pyrealm_build_data/rpmodel/test_inputs.json") as f:
+    data = json.load(f)
+
+# Test values of forcing variables as input to functions
+tc_ar_values = data["tc_ar"]
+patm_ar_values = data["patm_ar"]
+co2_ar_values = data["co2_ar"]
+
```
davidorme, line 56: Meant to say - this would probably be cleaner as a fixture? That ensures that each test gets a clean separate copy of the data. I can't see that going wrong here, but it is better practice I think. Also, using `importlib.resources` is a more robust way to get the file.

```suggestion
@pytest.fixture
def function_test_data()
    data_path =  (
        resources.files("pyrealm_build_data.subdaily") / "test_inputs.json"
    )

    with open(data_path) as f:
        data = json.load(f)

    # Test values of forcing variables as input to functions
    return data["tc_ar"],  data["patm_ar"], data["co2_ar"]
```

Tests would then need to do:

```python
def test_out_of_bound_output(function_test_data, tc, patm):
    """Function to calulate kmm."""
    tc_ar_values, patm_ar_values, co2_ar_values = function_test_data
```

tests/unit/pmodel/test_optimal_chi.py
davidorme: That last line should be `assert np.all(np.isnan(getattr(instance, key)))` or similar.

tests/unit/pmodel/test_optimal_chi.py
surbhigoel77: Yeah, makes sense. I will create a new issue to include imports within tests that are using these imports in all test files.

surbhigoel77:
@davidorme Could you accept the changes that you requested? I will only be able to merge once you approve them. A new PR is to be opened by me for putting imports within testcases

davidorme (approved):
Looks good to me.

Can you resolve the outstanding comments? That might just be clicking resolve, but comments on how it was resolved - even if just "implemented" - are also useful.

tests/unit/pmodel/test_optimal_chi.py
surbhigoel77: Implemented.

tests/unit/pmodel/test_pmodel_environment.py
surbhigoel77: Updated.

tests/unit/pmodel/test_pmodel_environment.py
surbhigoel77: Implemented.

tests/unit/pmodel/test_optimal_chi.py
surbhigoel77: Put it in a new PR.

tests/unit/pmodel/test_calc_optimal_chi.py
```diff
@@ -0,0 +1,73 @@
+"""Testing CalcOptimal submodule."""
```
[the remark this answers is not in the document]
surbhigoel77: Implemented.

PR #169, opened 2024-01-15
Handling of incomplete days in FastSlowScaler

### LINKED ISSUE #72: Handling incomplete days at the start and end of time series in FastSlowScaler
**Describe the bug**

The `FastSlowScaler` class expects a whole number of days worth of subdaily observations and throws an error if there are overlaps on either end. That is more or less sensible but is really limiting when dealing with global applications as time series often have to be shifted from UTC to local time. That kind of shift typically leaves incomplete days on either end, which doesn't really matter for applying the model and having to subset the data down to complete days is irritating.

MarionBWeinzierl:
This PR fixes #72 .

It handles potential incomplete first and last days of measurements by padding, both for the datetime array and the value array.

Tasks for this:
- [x] Implement padding for datetimes array
- [x] Implement padding for values array (can we assume consistent incompleteness? does the dimensionality of the values array pose a problem?)
- [x] Consider alternatives for padding with NaNs - especially for datetimes NaNs might not be the best padding, as it throws some other functions off
- [x] Adapt tests to no longer expect error messages for incomplete days
- [ ] Add tests to test new behaviour


davidorme:
I'd been thinking that the `_include` structure handles the padding, but of course the `_set_times` method and other places wants `self.datetimes` to map onto the first axis. It does seem easier to pad the datetimes than to fix all the downstream special cases.

I've done something similar before using `np.unique`, which might be useful?

```python

datetimes = np.arange(
    np.datetime64('2002-01-03 03:15'),
    np.datetime64('2002-01-06 09:45'), 
    np.timedelta64(30, 'm')
)

# Work out observations by date
dates = datetimes.astype("datetime64[D]")
unique_dates, date_counts = np.unique(dates, return_counts=True)
complete_count = date_counts.max()

padding = (complete_count - date_counts[0], complete_count - date_counts[-1])

if padding != (0,0):

    # Do we have at least _one_ complete day
    if len(unique_dates) < 3:
        ValueError("Not enough data")

    # Get a complete set of observation times from day 2 - we have guaranteed that the 
    # second day is complete.
    complete_times = datetimes[dates == unique_dates[1]]
    observation_timedeltas = complete_times - complete_times[0].astype("datetime64[D]")
    
    # Recreate the complete timeseries, using an offset to the last observations to avoid an 
    # off by one error from np.arange.
    datetimes = np.arange(
        unique_dates[0] + observation_timedeltas[0],
        unique_dates[-1] + (observation_timedeltas[-1] + np.timedelta64(1, 's')), 
        np.diff(observation_timedeltas)[0]
    )
```

tests/pmodel/test_fast_slow_pmodel.py
```diff
@@ -135,9 +135,9 @@ def test_FSPModel_corr(be_vie_data_components):
 
     # Test that non-NaN predictions correlate well and are approximately the same
     gpp_in_micromols = fs_pmodel.gpp[valid] / env.const.k_c_molmass
-    assert np.allclose(gpp_in_micromols, expected_gpp[valid], rtol=0.2)
+    assert np.allclose(gpp_in_micromols, expected_gpp[valid], rtol=0.9)  # rtol=0.2)
     r_vals = np.corrcoef(gpp_in_micromols, expected_gpp[valid])
-    assert np.all(r_vals > 0.995)
+    assert np.all(r_vals > 0.98)  # 0.995)
```
MarionBWeinzierl, line 140: @davidorme , is it acceptable here to lower the threshold?

tests/pmodel/test_fast_slow_pmodel.py
```diff
@@ -135,9 +135,9 @@ def test_FSPModel_corr(be_vie_data_components):
 
     # Test that non-NaN predictions correlate well and are approximately the same
     gpp_in_micromols = fs_pmodel.gpp[valid] / env.const.k_c_molmass
-    assert np.allclose(gpp_in_micromols, expected_gpp[valid], rtol=0.2)
+    assert np.allclose(gpp_in_micromols, expected_gpp[valid], rtol=0.9)  # rtol=0.2)
```
MarionBWeinzierl, line 138: I need to look into what is happening here

MarionBWeinzierl:
@davidorme , at the moment (apart from the issues marked with comments in the changes), there is quite a non-pretty issue in the implementation. I had first assumed that all padding could happen within the methods of `fast_slow_scaler`, using the padding as a private function. However, this seems not to be possible due to some of the stuff which is going on in `subdaily`. So, as it stands, I do some padding in `get_window_values`, but then a lot of stuff has to be padded in `subdaily`, too. It might be better to pad the `env` arrays, as discussed in #72 , but then again that would, without further check, lead to double-padding when the `fs_scaler` functions are called.

Do you have any preference how this should be handled? We could
- keep the current configuration, accepting that padding might happen inside as well as outside of `fast_slow_scaler`
- move everything outside of `fast_slow_scaler`
- come up with another clever way I haven't thought of

tests/pmodel/test_fast_slow_pmodel.py
davidorme: I'm surprised it has changed - I think I pushed that required correlation as high as it would go, so something has degraded the fit.

davidorme:
> However, this seems not to be possible due to some of the stuff which is going on in `subdaily`.

I haven't quite got my head around what happens outside of the `FastSlowScaler`. So, with incomplete days:

* `get_daily_means()` takes a fast time series and returns daily values. On the first day, if the fast data starts after the time indices set in `include` then the first daily mean should be NA. On the last day if the fast data stops before the end of those same indices it should also be NA. I think there's a case for a `strict=False` mode that allows a subset of the `include` values to be present, but that's a bonus extra. 
* `fill_daily_to_subdaily` should take daily values and return the data unpadded with a time axis length equal to the original observation datetimes.

I think I'm missing something!

MarionBWeinzierl:
Just to show the difference I get in that test at the moment:

cmp_pyrealm_test

There seems to be a bit of a left shift, but also a difference in magnitude which is stronger towards the middle.

tests/pmodel/test_fast_slow_pmodel.py
MarionBWeinzierl: What confuses me is that in this test case there is/should not be any padding, as the datetimes are full days. `fsscaler.num_missing_values_end` and `fsscaler.num_missing_values_start` are both `0`

pyrealm/pmodel/subdaily.py
```diff
@@ -271,15 +271,15 @@ def __init__(
         r"""Realised daily slow responses in :math:`J_{max25}`"""
 
         # Fill the daily realised values onto the subdaily scale
-        subdaily_tk = self.env.tc + self.env.core_const.k_CtoK
+
+        subdaily_tk = fs_scaler.pad_values(self.env.tc + self.env.core_const.k_CtoK)
```
davidorme, line 275: I'm not sure `pad_values` should ever be used outside of `FastSlowScaler.get_daily_means`?

MarionBWeinzierl: That had been my question in https://github.com/ImperialCollegeLondon/pyrealm/pull/169#issuecomment-1948559160

davidorme: So the scaler extracts a set of daily values along the time axis (given a daily window via a `set` method). The modifications for this PR need to allow the the correct values to be extracted when the datetimes don't wrap neatly onto days.

Going back to the subdaily timescale is I think handled automatically by the `scipy` interpolation - we interpolate from the daily values back to the actual provided datetimes so there is no need to modify the behaviour there.

I wasn't entirely sure I wasn't talking rubbish here, so I've pushed an alternative branch with an implementation: https://github.com/ImperialCollegeLondon/pyrealm/tree/72-alternate-implementation

The main thing I wanted to check here is that padding the `be_vie_data_components` passed the correlation test - the new test `test_FSPModel_corr_padded` on this branch sticks 6 hour chunks at the start, end and both for the datetimes on that test and it works as expected. The GPP predictions shouldn't change at all because the data are padded with `np.nan`, so the real calculations start in the same place, but the function accepts the incomplete days without complaint.

davidorme: I forgot to push the actual commits to that branch - now done.

MarionBWeinzierl: Hm, it still tells me the branches are identical?

davidorme: Gah - pushed `develop` instead. Now done!

MarionBWeinzierl:
OK, what would you prefer at this point, @davidorme ? Do you want to raise a new PR with your implementation, and we close this one, or does it make any sense for me to further work on this one?

pyrealm/pmodel/subdaily.py
MarionBWeinzierl: OK, I removed the padding from the subdaily -- I just still don't know why in this case it makes a difference, as the padding should be zero. :-/

MarionBWeinzierl: The padding was also somehow necessary when I did the profiling test, I thought, so checking this. now that I removed it.

MarionBWeinzierl: Hu, it no longer complains in that case. It must have been resolved through previous changes.

MarionBWeinzierl:
> ## [Codecov](https://app.codecov.io/gh/ImperialCollegeLondon/pyrealm/pull/169?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ImperialCollegeLondon) Report
> 
> Attention: Patch coverage is `88.57143%` with `4 lines` in your changes are missing coverage. Please review.
> 
> > Project coverage is 94.72%. Comparing base [(`b41b416`)](https://app.codecov.io/gh/ImperialCollegeLondon/pyrealm/commit/b41b41695d2df28a0be0080a805cbcf1a5439c57?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ImperialCollegeLondon) to head [(`9efa1af`)](https://app.codecov.io/gh/ImperialCollegeLondon/pyrealm/pull/169?dropdown=coverage&src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ImperialCollegeLondon).
> 
> [Files](https://app.codecov.io/gh/ImperialCollegeLondon/pyrealm/pull/169?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ImperialCollegeLondon) 	Patch % 	Lines
> [pyrealm/pmodel/fast_slow_scaler.py](https://app.codecov.io/gh/ImperialCollegeLondon/pyrealm/pull/169?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ImperialCollegeLondon#diff-cHlyZWFsbS9wbW9kZWwvZmFzdF9zbG93X3NjYWxlci5weQ==) 	88.57% 	[4 Missing ⚠️ ](https://app.codecov.io/gh/ImperialCollegeLondon/pyrealm/pull/169?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ImperialCollegeLondon)
> Additional details and impacted files
> 
> [☔ View full report in Codecov by Sentry](https://app.codecov.io/gh/ImperialCollegeLondon/pyrealm/pull/169?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ImperialCollegeLondon). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=ImperialCollegeLondon).

This is not currently checking the padded version. If we stick with this version of the padding implementation we could cherry-pick your `be_vie_data_components_padded` test from your version, @davidorme .

MarionBWeinzierl:
Close, as replaced by #194

PR #170, opened 2024-01-16
Add unit tests for splash

### LINKED ISSUE #162: Create unit tests for SPLASH module
This issue is related to #153 and #69 and #130.

After the merging of the SPLASH module into Pyrealm, unit tests need to be implemented. These should in particular cover the catching of invalid inputs.

There will be some overlaps with unit tests for core modules which need to be considered appropriately.

tztsai:
# Description

This pull request introduces a unit testing suite for the `SplashModel` class in the `test_splash.py` file. The main changes include:

1. New Test Cases: Added new test cases for the `estimate_daily_water_balance`, `estimate_initial_soil_moisture`, and `calc_soil_moisture` methods.

1. Randomized Testing: Introduced randomized testing by randomly generating `dates` and `wn_init`. This ensures that the tests cover a wide range of possible scenarios and increases the robustness of the test suite.

**Fixes** #162, #175

## Type of change

## Key checklist

## Further checks


MarionBWeinzierl (changes requested):
Tests are failing and need fixing.

tztsai:
> Tests are failing and need fixing.

Some tests are checking whether `SplashModel` raises errors when some input variables are out of bound, as suggested in the last meeting. Since some boundaries checks have not been implemented, these tests fail.

MarionBWeinzierl:
> > Tests are failing and need fixing.
> 
> Some tests are checking whether `SplashModel` raises errors when some input variables are out of bound, as suggested in the last meeting. Since some boundaries checks have not been implemented, these tests are failing.

In this case, the test needs to check that a specific error is raised, and succeed if this error is raised. Have a look at, for example, https://github.com/ImperialCollegeLondon/pyrealm/blob/develop/tests/unit/pmodel/test_fast_slow_scaler.py how this works.

tztsai:
> > > Tests are failing and need fixing.
> > 
> > 
> > Some tests are checking whether `SplashModel` raises errors when some input variables are out of bound, as suggested in the last meeting. Since some boundaries checks have not been implemented, these tests are failing.
> 
> In this case, the test needs to check that a specific error is raised, and succeed if this error is raised. Have a look at, for example, https://github.com/ImperialCollegeLondon/pyrealm/blob/develop/tests/unit/pmodel/test_fast_slow_scaler.py how this works.

Thanks for the link, but I have already implemented it. I meant `SplashModel` does not raise the expected errors when some input variables are out of bound, and that's why some unit tests do not pass.

tztsai:
> > > > Tests are failing and need fixing.
> > > 
> > > 
> > > Some tests are checking whether `SplashModel` raises errors when some input variables are out of bound, as suggested in the last meeting. Since some boundaries checks have not been implemented, these tests are failing.
> > 
> > 
> > In this case, the test needs to check that a specific error is raised, and succeed if this error is raised. Have a look at, for example, https://github.com/ImperialCollegeLondon/pyrealm/blob/develop/tests/unit/pmodel/test_fast_slow_scaler.py how this works.
> 
> Thanks for the link, but I have already implemented it. I meant `SplashModel` has not implemented the boundary checks and does not raise the expected errors when some input variables are out of bound, and that's why some unit tests do not pass.

MarionBWeinzierl:
> > > > Tests are failing and need fixing.
> > > 
> > > 
> > > Some tests are checking whether `SplashModel` raises errors when some input variables are out of bound, as suggested in the last meeting. Since some boundaries checks have not been implemented, these tests are failing.
> > 
> > 
> > In this case, the test needs to check that a specific error is raised, and succeed if this error is raised. Have a look at, for example, https://github.com/ImperialCollegeLondon/pyrealm/blob/develop/tests/unit/pmodel/test_fast_slow_scaler.py how this works.
> 
> Thanks for the link, but I have already implemented them... I meant `SplashModel` has not implemented the boundary checks and does not raise the expected errors when some input variables are out of bound, and that's why some unit tests do not pass.

OK, then you should raise a new issue for this, and ideally implement those checks that you need in SPLASH, so that the checks pass.

davidorme:
I'm also happy for a PR to include test-lead development of the code, so we could add the modifications to SplashModel here. Unless that is bad practice, @MarionBWeinzierl?

MarionBWeinzierl:
> I'm also happy for a PR to include test-lead development of the code, so we could add the modifications to SplashModel here. Unless that is bad practice, @MarionBWeinzierl?

Hm, one could argue that it would be against a  very clear separation of concerns in issues (PR is about adding unit tests). But I am not opposed to what you suggest, if you are happy with it. The question would still be, do you (@davidorme ) want 
1. an additional issue which is closed by this same PR?
2. this work to be included in #162 ?
3. or this work to be done "on the fly" for this PR without an issue?

tztsai:
I've added boundary checks in the Splash module. If necessary I can split it into a separate PR.

MarionBWeinzierl:
> I've added boundary checks in the Splash module. If necessary I can split it into a separate PR.

We do not need a new PR. A new issue explaining the problem and the necessary changes, which could be referenced in this PR, would be good.

davidorme (changes requested):
Most of this looks good to me - let's just clarify what is being handled here and then there are a couple of other points to tackle.

tests/unit/splash/test_splash_model.py
```diff
@@ -0,0 +1,146 @@
...
+
+@pytest.fixture(params=[(0, 365), (0, 366), (123, 567)])  # parametrized fixture
+def calendar(request, grid_benchmarks):
+    """Provide the dates with a random start date."""
+    from pyrealm.core.calendar import Calendar
+
+    dates = grid_benchmarks[0].time.data
+    start, end = request.param
+    yield Calendar(dates[start:end])  # fmt: skip
+
```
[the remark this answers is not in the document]
tztsai: I've added a PR for that.

pyrealm/splash/splash.py
```diff
@@ -62,7 +69,7 @@ def __init__(
         tc: NDArray,
         pn: NDArray,
         dates: Calendar,
-        kWm: NDArray = np.array([150.0]),
+        kWm: float = 150.0,
         core_const: CoreConst = CoreConst(),
```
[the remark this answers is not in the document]
tztsai: I've restored the typing and updated `bounds_checker` because it raised an error when the higher bound is `splash.kWm`.

pyrealm/splash/splash.py
```diff
@@ -144,14 +155,14 @@ def estimate_initial_soil_moisture(
             # Check the shape is the same as the shape of a slice along axis 0
             if wn_init.shape != self.tc[0].shape:
                 raise ValueError("Incorrect shape in wn_init")
-            wn_start = wn_init
+            wn_start = bounds_checker(wn_init, 0, self.kWm)
         else:
```
davidorme, line 159: I think there's a good case here for making this an Exception. 

`bounds_checker` only warns because remotely sensed data inputs often have wild values - that function used to raise Exceptions but it just happened all the time and made it unusable. But here, we need a more explicit test that  values are in `[0, 150]` and it should fail if they aren't.

davidorme (changes requested):
A couple of things to update, please.

pyrealm/splash/splash.py
```diff
@@ -88,6 +95,10 @@ def __init__(
         self.kWm = kWm
         """The maximum soil water capacity for sites."""
```
davidorme, line 96: There's an error in my implementation - the attributes should be typed, as noted here:
https://github.com/ImperialCollegeLondon/pyrealm/pull/172#discussion_r1466147277

tztsai: Fixed in the latest commit

pyrealm/splash/splash.py
tztsai: Fixed in the latest commit.

pyrealm/splash/splash.py
```diff
@@ -88,6 +95,10 @@ def __init__(
...
 
+        # Check variables are within expected ranges
+        for var, (lo, hi) in self.variable_ranges.items():
+            bounds_checker(getattr(self, var), lo, hi)
+
```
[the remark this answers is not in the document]
tztsai: Fixed in the latest commit.

pyrealm/splash/splash.py
```diff
@@ -275,7 +279,7 @@ def calculate_soil_moisture(
         wn_out = np.full_like(self.tc, np.nan)
         ro_out = np.full_like(self.tc, np.nan)
 
-        curr_wn = wn_init
+        curr_wn = bounds_checker(wn_init, 0, self.kWm)
         for day_idx in np.arange(self.pn.shape[0]):
```
davidorme, line 283: This should be a value error like L238 - L240 above. The initial definition of `self.kWm` should be flexible and allow values with a warning if they seem very high - as it now is with the `bounds_checker` - but that sets hard limits and it should be an error to exceed those initial values.

davidorme (changes requested):
Looks good to me except for one final change. Note that adding a hard value check at L278 means that the input typing to `bounds_checker` can be reverted to `float` too.

pyrealm/splash/splash.py
tztsai: Ah sorry for missing this one and thanks for catching it! I also have updated the type annotation of `Bounds_checker`.

PR #171, opened 2024-01-16
hotfix test_evap.py

### LINKED ISSUE #173: Regression test of `splash.evap` does not pass
After the hot fix in #171, the assertion in `tests/regression/test_evap.py` fails.

tztsai:
* Fixes a small error: `assert` is absent for checking `np.allclose(...)`.

@davidorme edits:

* Also fixes an error in the test input: AET needs to be calculated using the soil moisture estimate from the previous day. Including that offset correctly duplicates the expected outputs exactly.
* Fixes #173


MarionBWeinzierl (approved):
Looks good to me.

Issue #119, opened 2023-10-11
Refactor `CalcOptimalChi` to make it extensible and accept slowly responding `xi`


davidorme:
**Is your feature request related to a problem? Please describe.**

The initial implementation of the `subdaily` model only describes C3 plants and the standard C3 model. There are a wider range of P Model implementations - mostly differences in the approaches used in `CalcOptimalChi` - that also need subdaily implementations. The main barrier here is the estimation of slow response in the $\xi$ (`xi`) parameter.

**Describe the solution you'd like**

* The `CalcOptimalChi` class should become an abstract baseclass providing a flexible approach for adopting new models.
* That should provide a registry of class details including C3/C4 status.
* The ABC should stipulate a `calc_chi` method with the signature `calc_chi(estimated_xi: Optional[NDArray])`. If `estimated_xi` is missing, the method should use the instantaneous estimate of `xi` from the provided data, but otherwise, this method can be used to estimate `chi` given a calculated slow response in `xi`.
* This should allow more of the `PModel` class mechanism to be used in calculating `subdaily` models and in switch between different approaches to estimating `chi`.

PR #172, opened 2024-01-16
Draft implementation of new OptimalChi structure

davidorme:
# Description

This is PR implements the first stage of addressing #119, allowing the main `PModel` code to be used within subdaily caculations. These changes provide a complete refactor of the `CalcOptimalChi`class, which has a bunch of different approaches as methods, into the `NewCalcOptimalChi` abstract base class (ABC), which then set of subclasses providing the different approaches.

The main aims here are:

1. Provide a way to recalculate `chi` using modified values of the `xi` variable. The ABC requires subclasses to provide `set_beta` and `estimate_chi`: both of these methods are used as part of `__init__` for a derived class, but `estimate_chi` can also be used separately to inject new `xi` values.
2. Make it easier and cleaner to add new approaches - with the new setup, users can actually create their own subclasses, should they want to. For this reason, a registry object is added that allows specific approaches to be retrieved using a method name.

One thing we might do in future but seems paranoid is make the `method` and `is_c4` class attributes read only, but class attributes in ABCs seem finicky: using `@property` seems like one approach but then causes `mypy` issues with setting the values. That is paranoid though.

The PR now:

* completes the redefinitions of the pre-existing options for `CalcOptimalChi`
* restructures the `rootzonestress` option into `PModelEnvironment` for a more consistent interface and cleaner signatures.
* adds new tests to check rootzonestress handling and confirm that the new approach matches the old one.
* substitutes the new `OPTIMAL_CHI_CLASS_REGISTRY` interface into `PModel` and confirms the test suite passes with the new implementation.

The last completed changes are to:

- [x] Remove the `pmodel.calc_optimal_chi.py` module
- [x] Rename the `pmodel.new_calc_optimal_chi.py` objects and files 
- [x] Fix up any `sphinx` references
- [x] Update the docs notebooks to remove old code and demonstrate the new implementation.

## Type of change

## Key checklist

## Further checks


MarionBWeinzierl:
@davidorme , would  `Final` attributes ([see here](https://mypy.readthedocs.io/en/stable/final_attrs.html)) do what you want, instead of using `@property` ?

MarionBWeinzierl:
Also, and this might be controversial and/or philosophical: I would put the abstract base class and each derivative in its own file, because otherwise the file might become large, and if people derive their own subclasses they might want to keep that with their module, and not all add it into this one file.

If, on the other hand, you believe there won't be any additional implementations of that abstract class it is fine to keep everything together in one file.

davidorme:
> @davidorme , would `Final` attributes ([see here](https://mypy.readthedocs.io/en/stable/final_attrs.html)) do what you want, instead of using `@property` ?

That's interesting. I'm a bit conflicted about `Final` because it seems to step over a line between indicating typing and controlling the usage of objects - I think I've just got a narrow view on what `mypy` does though.

More broadly, I should have deleted those `@property` comments. I've used that approach before but it is really defensive and adds complexity - if people manually change the properties of class attributes, then they should expect to get bitten 😄 I've always tended to be that defensive, but I've learned that this is often at the expense of making the code less transparent.

davidorme:
> if people derive their own subclasses they might want to keep that with their module, and not all add it into this one file.

My intention in the implementation here is that someone can include a new subclass _outside_ of the package code. If they have a python file that does this:

```python
from pyrealm.pmodel.calc_optimal_chi_new import CalcOptimalChiNew

class MyCustomCalcOptChi(CalcOptimalChiNew, method = 'custom_opt_chi', is_c4=False):
    ...
```
then the `__init_subclass__` would add the method to the registry and they could then use `PModel(..., method='custom_opt_chi`). Once, that is, this approach is rolled out so that PModel does something like:

```python
try:
    opt_chi_method = OPTIMAL_CHI_CLASS_REGISTRY[method]
except KeyError:
    ...

self.opt_chi = opt_chi_method(...)
```

davidorme:
> f people derive their own subclasses they might want to keep that with their module

Of course if people wanted to contribute that method to the package, then this could get out of hand, but that seems a way away :smile: I can see the value in splitting the ABC from the subclasses though.

pyrealm/pmodel/calc_optimal_chi_new.py
```diff
@@ -0,0 +1,573 @@
...
+    # def is_c4(cls) -> bool:
+    #     """Does the class represent a C4 pathway."""
+
+    method: str
+    is_c4: bool
+
+    def __init__(
+        self,
+        env: PModelEnvironment,
+        rootzonestress: NDArray = np.array([1.0]),
```
tztsai, line 103: The default value of `rootzonestress` is mutable, so if there is an instance `opt_chi` and a user sets `opt_chi.rootzonestress[0] = 2.0`, the default value of `rootzonestress` of all instances of `NewCalcOptimalChi` will also be changed. This is probably not desirable. See https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments.

pyrealm/pmodel/calc_optimal_chi_new.py
```diff
@@ -0,0 +1,573 @@
...
+    def set_beta(self) -> None:
+        """Set ``beta`` to a constant C3 specific value."""
+        # leaf-internal-to-ambient CO2 partial pressure (ci/ca) ratio
+        self.beta = self.pmodel_const.beta_cost_ratio_prentice14
+
+    def estimate_chi(self, xi_values: Optional[NDArray] = None) -> None:
+        """Estimate ``chi`` for C3 plants."""
+
+        if xi_values is not None:
+            _ = check_input_shapes(self.env.ca, xi_values)
```
tztsai, line 254: Maybe no need to assign the output to `_`.

pyrealm/pmodel/calc_optimal_chi_new.py
```diff
@@ -0,0 +1,573 @@
...
+        (:math:`m_c`)."""
+        self.mj: NDArray
+        r""":math:`\ce{CO2}` limitation factor for light-limited assimilation
+        (:math:`m_j`)."""
+        self.mjoc: NDArray
+        r"""Ratio of :math:`m_j/m_c`."""
+
+        self.set_beta()
+        self.estimate_chi()
+
```
tztsai, line 150: Maybe add a check to ensure that any custom subclass of `NewCalcOptimalChi` sets the values of `beta`, `xi`, etc. in its `set_beta()` and `estimate_chi()` methods? Besides, it seems that some of these variables are derived from the others, so maybe they can be set after calling `self.estimate_chi()`. For example,
```python
self.ci = self.chi * self.env.ca
self.mjoc = self.mj / self.mc
```

pyrealm/pmodel/calc_optimal_chi_new.py
davidorme: Good point. I guess the canonical way of handling this is to do `rootzonestress: None | NDArray = None` and then populate it in `__init__`.

I was thinking of pushing `rootzonestress` back into `PModelEnvironment` - `theta` there is handled in this way and it would be more consistent to pass all the variables via that object. I can then have `rootzonestress` specific subclasses of `CalcOptimalChiNew`.

pyrealm/pmodel/calc_optimal_chi_new.py
davidorme: I tend to use that approach to make it clear that the function does return something, rather than altering its arguments. We could make `check_input_shape` which returns None and simply wraps `get_input_shapes` and drops the return value.

pyrealm/pmodel/calc_optimal_chi_new.py
davidorme: It is unlike me not to have added that checking - I'm usually pretty defensive in coding. I guess my thinking here is that a user might pass in bad inputs by mistake, but someone developing a new class is probably going to take more care. We could simply run check shape again though.

It is true that `ci` and `mjoc` are shared - I'd left them there in case a method came along that needed to define them differently.  That seems very unlikely though.

tztsai: Yes, adding a shape check is a good idea, ensuring both existence of these attributes and the consistency of their shapes.

pyrealm/pmodel/calc_optimal_chi_new.py
davidorme: That doesn't conform to the `NDArray` typing though

davidorme: Yes, but that's not obvious because the `SplashModel` object is missing typing on the attributes - those should all match the typing from the `__init__` signature.

MarionBWeinzierl (approved):
This looks good to me

surbhigoel77 (approved):
The changes in the structure of OptimalChi look good to me. The changes certainly make the submodule more extendable in terms of methods and features.

PR #179, opened 2024-02-02
Use a statistical model fitted to the original dataset to synthesize data

tztsai:
# Description

A statistical model (with a linear component and several seasonal components with different periods) was fitted to the large LFS dataset and stored in `pyrealm_build_data/data_model_params.nc`, by the script `pyrealm_build_data/synth_data.py`. It can be later used to reconstruct the original dataset with a decent accuracy, effectively reducing the data size from 1.8GB to 38MB. 

A testing file was also added in `tests/regression/data/` to ensure the quality of the reconstructed dataset by asserting that its R2 score is at least 0.85 (1.0 means perfect).

### DIFF
--- a/pyrealm_build_data/synth_data.py
+++ b/pyrealm_build_data/synth_data.py
@@ -82,28 +82,3 @@ for k, a in params_ds.items():
 fs = make_time_features(ds.time).to_xarray().to_dataarray()
 params_ds["feature"] = fs
 params_ds.to_netcdf("data_model_params.nc")
-
-
-def reconstruct_data(params_ds: xr.Dataset) -> xr.Dataset:
-    """Reconstruct the data from the model parameters."""
-    rec_ds = xr.Dataset()
-    f = params_ds["feature"]
-    for k, a in params_ds.items():
-        if k != "feature":
-            rec_ds[k] = xr.dot(f, a)
-    return rec_ds
-
-
-rec_ds = reconstruct_data(params_ds)
-
-
-def r2_score(y_true: ArrayLike, y_pred: ArrayLike) -> float:
-    """Compute the R2 score."""
-    return 1 - ((y_true - y_pred) ** 2).sum() / ((y_true - y_true.mean()) ** 2).sum()
-
-
-for k in ds.data_vars:
-    t = ds[k].sel(lat=rec_ds.lat).squeeze()[::321]
-    p = rec_ds[k].squeeze()[::321]
-    r2 = r2_score(t, p)
-    print(f"R2 score for {k}: {r2:.2f}")
--- a/.gitignore
+++ b/.gitignore
@@ -30,7 +30,6 @@ pyrealm.egg-info
 
 # Data
 pyrealm_build_data/inputs_data_24.25.nc
-pyrealm_build_data/eda.py
 
 # Profiling
 prof/
--- a/pyrealm_build_data/synth.py
+++ b/pyrealm_build_data/synth.py
@@ -1,0 +1,109 @@
+"""This script uses a parametrized model to compress the input dataset.
+
+It fits a time series model to the input data and stores the model parameters,
+which can be later used to reconstruct the original dataset.
+"""
+import numpy as np
+import pandas as pd
+import statsmodels.api as sm
+import xarray as xr
+from np.typing import ArrayLike
+from tqdm import tqdm
+
+ds = xr.open_dataset("inputs_data_24.25.nc")
+
+const_vars_over_time = ["patm"]
+const_vars_over_space = ["co2"]
+
+
+def make_time_features(dt: ArrayLike, is_const: bool = False) -> pd.DataFrame:
+    """Make time features for a given time index."""
+    dt = pd.to_datetime(dt).rename("time")
+    df = sm.add_constant(pd.DataFrame(index=dt))
+
+    if is_const:
+        return df
+
+    # Linear trend
+    df["linear"] = (dt - pd.Timestamp("2000-01-01")) / pd.Timedelta("1h")
+
+    # El-Nino and La-Nina seasonality component
+    for k in [1, 2, 3, 4]:
+        df[f"cross_year_{k}_sin"] = np.sin(2 * k * np.pi * dt.year / 12)
+        df[f"cross_year_{k}_cos"] = np.cos(2 * k * np.pi * dt.year / 12)
+
+    # Yearly seasonality component
+    for k in [1, 2, 3, 4]:
+        df[f"yearly_{k}_sin"] = np.sin(2 * k * np.pi * dt.dayofyear / 365)
+        df[f"yearly_{k}_cos"] = np.cos(2 * k * np.pi * dt.dayofyear / 365)
+
+    # Daily seasonality component
+    for k in [1, 2, 3, 4]:
+        df[f"daily_{k}_sin"] = np.sin(2 * k * np.pi * dt.hour / 24)
+        df[f"daily_{k}_cos"] = np.cos(2 * k * np.pi * dt.hour / 24)
+
+    return df
+
+
+def fit_ts_model(da: xr.DataArray, is_const: bool = False) -> pd.DataFrame:
+    """Fit a time series model to the data."""
+    df = da.sel(time=slice(None, None, 2)).to_series().unstack("time").T
+    df = df.dropna(axis=1, how="all")
+    features = make_time_features(df.index, is_const=is_const)
+    params = pd.DataFrame(index=df.columns, columns=features.columns)
+    for k in tqdm(df.columns):
+        y = df[k].dropna()
+        X = features.loc[y.index]
+        model = sm.OLS(y, X).fit()
+        params.loc[k] = model.params
+    return params.astype(float)
+
+
+params_ds = xr.Dataset()
+
+for k in ds.data_vars:
+    print("Fitting", k)
+    if k in const_vars_over_space:
+        da = ds[k].isel(lat=[0], lon=[0])
+    else:
+        da = ds[k]
+    if k in const_vars_over_time:
+        ps = fit_ts_model(da, is_const=True)
+    else:
+        ps = fit_ts_model(da)
+    params_ds[k] = ps.to_xarray().to_dataarray()
+
+for k, a in params_ds.items():
+    if k in const_vars_over_space:
+        params_ds[k] = a.fillna(a.isel(lat=0, lon=0))
+    elif k in const_vars_over_time:
+        params_ds[k] = a.fillna(0)
+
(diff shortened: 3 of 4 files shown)


davidorme (commented):
Can you fix the `qa` issues, please.

I'm not sure about this PR. I think the implementation and the time series model is very neat and the R2 is good, but the resulting coefficients file is 39MB (compressed or not?) and the validation script requires LFS to restore the original 1.6 GB data file. 

The main point of #149 was to remove the large dataset and drop LFS, and the retained inputs are ~50MB. We can tile that small dataset to increase the profiling load. This PR adds more variation in the input data, by retaining more of the original signal of the large dataset, but it isn't clear that this alone improves the profiling.

tztsai:
> Can you fix the `qa` issues, please.
> 
> I'm not sure about this PR. I think the implementation and the time series model is very neat and the R2 is good, but the resulting coefficients file is 39MB (compressed or not?) and the validation script requires LFS to restore the original 1.6 GB data file.
> 
> The main point of #149 was to remove the large dataset and drop LFS, and the retained inputs are ~50MB. We can tile that small dataset to increase the profiling load. This PR adds more variation in the input data, by retaining more of the original signal of the large dataset, but it isn't clear that this alone improves the profiling.

I have fixed the QA now. I think the validation script only needs to run once to ensure the reconstructed dataset is similar enough to the original dataset. After that the profiling can just call the `reconstruct` function without any usage of `LFS`. In addition, the previous coefficients file stores features for each time step of the original dataset. Letting the `reconstruct` function calculate these features dynamically given a custom `datetime` index, the file only needs to occupy ~300KB.

pyrealm_build_data/synth_data.py
```diff
@@ -0,0 +1,81 @@
+"""This script uses a parametrized model to compress the input dataset.
+
+It fits a time series model to the input data and stores the model parameters.
+The dataset can then be reconstructed from the model parameters using the `reconstruct`
+function, provided with a custom time index.
+"""
```
davidorme, line 6: What is this custom time index and the special time features - not clear.

pyrealm_build_data/synth_data.py
```diff
@@ -0,0 +1,81 @@
...
+import xarray as xr
+
+VAR_BOUNDS = dict(
+    temp=(-25, 80),
+    patm=(3e4, 11e4),
+    vpd=(0, 1e4),
+    co2=(0, 1e3),
+    fapar=(0, 1),
+    ppfd=(0, 1e4),
+)
```
davidorme, line 20: Why are these a global variable if they are only used in one function.

pyrealm_build_data/synth_data.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def make_time_features(t: np.ndarray) -> pd.DataFrame:
+    """Make time features for a given time index."""
+    dt = pd.to_datetime(t).rename("time")
+    df = pd.DataFrame(index=dt).assign(const=1.0)
```
davidorme, line 26: Can you explain the features.

pyrealm_build_data/synth_data.py
```diff
@@ -0,0 +1,81 @@
...
+
+    return df
+
+
+def fit_ts_model(df: pd.DataFrame, fs: pd.DataFrame) -> Tuple[pd.DataFrame, float]:
+    """Fit a time series model to the data."""
+    df = df.dropna(axis=1, how="all").fillna(df.mean())
+    Y = df.values  # (times, locs)
+    X = fs.values  # (times, feats)
+    A = np.linalg.pinv(X) @ Y  # (feats, locs)
```
davidorme, line 42: `np.linalg.solve(X, Y)` is usually preferred? See for instance: https://stackoverflow.com/questions/31256252/why-does-numpy-linalg-solve-offer-more-precise-matrix-inversions-than-numpy-li

pyrealm_build_data/synth_data.py
```diff
@@ -0,0 +1,81 @@
...
+        da = ds[k].isel(time=slice(None, None, 4))  # downsample along time
+        df = da.to_series().unstack("time").T  # (datetimes, locations)
+        fs = features.loc[df.index]  # (datetimes, features)
+        fs = fs[special_time_features.get(k, fs.columns)]
+        ps, r = fit_ts_model(df, fs)  # (locations, features)
+        print("Loss:", r)
+        ps[features.keys().difference(ps.columns)] = 0.0
+        model[k] = ps.to_xarray().to_dataarray()
+
+    model.to_netcdf("pyrealm_build_data/data_model.nc")
```
davidorme, line 81: This whole chunk should be its own function with a docstring, that can then be called using `__name__ == "__main__"`.

davidorme (changes requested):
See the comments on this PR about the use of LFS. To add to that and the comments below, can you provide complete docstrings (`Args` etc) and a clear description of the usage of the functions, which is not clear. It would be good to have longer, more explanatory variable names and more comments.

pyrealm_build_data/synth_data.py
tztsai: It's just any array of the `datetime64` data type or a pandas `DatetimeIndex`, possibly generated by something like `pd.date_range("2000-1-1", "2009-12-31", freq="1h")`.

pyrealm_build_data/synth_data.py
tztsai: Good point - I can move it into the function. However, is it worth considering to create a global config of variable bounds over the whole `pyrealm` package, stored somewhere like in `pyrealm/constants`?

pyrealm_build_data/synth_data.py
tztsai: The model approximates the time series by a function $f(t) = a_0 + a_1 t + \sum_f sin(2\pi f t) + cos(2\pi f t)$, where $f = 730.5, 365.25, 12, ...$. Each feature is just a term in this function. I can add this into a comment.

tztsai:
> > I think the validation script only needs to run once to ensure the reconstructed dataset is similar enough to the original dataset.
> 
> Well, yes but the PR adds [`tests/regression/data/test_synth_data.py`](https://github.com/ImperialCollegeLondon/pyrealm/pull/179/files#diff-65a709ae2a608c22174eb25fa7a01e31d7b07a861078a5440e18e59f5d8ae614) to the regression testing suite and that is entirely reliant on the LFS file. It skips if the file is missing, but we're either restoring the need for LFS if we want that file to run or adding a test that will _always_ skip.

Locally I have checked out the LFS object, so the validation script can run locally to check the quality of the reconstructed dataset. In the CI workflow, however, since we removed the step of fetching LFS files, this test file will be skipped. This looks sensible - the developers can check the quality of the dataset synthesis in the local environment, but there's no need to run the quality check in the CI workflow later.

davidorme:
> > > I think the validation script only needs to run once to ensure the reconstructed dataset is similar enough to the original dataset.
> > 
> > 
> > Well, yes but the PR adds [`tests/regression/data/test_synth_data.py`](https://github.com/ImperialCollegeLondon/pyrealm/pull/179/files#diff-65a709ae2a608c22174eb25fa7a01e31d7b07a861078a5440e18e59f5d8ae614) to the regression testing suite and that is entirely reliant on the LFS file. It skips if the file is missing, but we're either restoring the need for LFS if we want that file to run or adding a test that will _always_ skip.
> 
> Locally I have checked out the LFS object, so the validation script can run locally to check the quality of the reconstructed dataset. In the CI workflow, however, since we have removed the step of fetching LFS files, this test file will be skipped. This looks sensible - the developers can check the quality of the dataset reconstruction in the local environment, and the reconstructed dataset will be used in the CI workflow for profiling later without the need of running the quality check again.

Yes but this requires that the whole repo retains the use of `git lfs` so that this flow can be used. And we had deliberately reduced the dataset size to avoid using `git lfs`.

tztsai:
> > > > I think the validation script only needs to run once to ensure the reconstructed dataset is similar enough to the original dataset.
> > > 
> > > 
> > > Well, yes but the PR adds [`tests/regression/data/test_synth_data.py`](https://github.com/ImperialCollegeLondon/pyrealm/pull/179/files#diff-65a709ae2a608c22174eb25fa7a01e31d7b07a861078a5440e18e59f5d8ae614) to the regression testing suite and that is entirely reliant on the LFS file. It skips if the file is missing, but we're either restoring the need for LFS if we want that file to run or adding a test that will _always_ skip.
> > 
> > 
> > Locally I have checked out the LFS object, so the validation script can run locally to check the quality of the reconstructed dataset. In the CI workflow, however, since we have removed the step of fetching LFS files, this test file will be skipped. This looks sensible - the developers can check the quality of the dataset reconstruction in the local environment, and the reconstructed dataset will be used in the CI workflow for profiling later without the need of running the quality check again.
> 
> Yes but this requires that the whole repo retains the use of `git lfs` so that this flow can be used. And we had deliberately reduced the dataset size to avoid using `git lfs`.
> 
> I think the point here is that we need a block of data to use in profiling:
> 
> * We have a block of real data that we can tile at the moment.
> * We could adopt this as a time series model _without_ really caring about the precision of the replication. We probably genuinely don't care although it would be good to have provenance.
> * I had been vaguely thinking about procedurally generated data (something like Perlin noise) instead.

A simple solution at the moment would be to validate the generated dataset using the reduced (1 year) dataset which does not require LFS.

pyrealm_build_data/synth_data.py
tztsai: `np.linalg.solve` only accepts square matrices. The corresponding function for non-square matrices is `np.linalg.lstsq`. I tried it in place of `np.linalg.pinv`, and it only saves ~0.06s for a single variable. However, I've used it in the code instead now.

pyrealm_build_data/synth_data.py
tztsai: I have moved the chunk into `fit_ts_model`.

MarionBWeinzierl:
When #189 is implemented the data set can also be uploaded to Zenodo.

MarionBWeinzierl:
Another option would be to use the reduced dataset for validity checking.

davidorme:
@MarionBWeinzierl

With #256, I think this becomes stale and we can close it. It only really exists because of the processing time and data file size limitations of trying to run the profiling on GithubActions.

If we move back to local profiling as planned, we'd still need test datasets but I think we can scale them back up to give more stringent and stable test and then store them on e.g. Zenodo. We can then close this.

MarionBWeinzierl:
Happy to have this closed.

tztsai:
Storing the dataset on Zenodo is a good alternative.

davidorme:
Closed as we have moved away from profiling on GitHub runners, which removes the need to compress the profiling datasets.

PR #180, opened 2024-02-05
Update to the latest versions of checkout and setup-python actions

### LINKED ISSUE #178: Update github actions.
**Describe the bug**
The CI on GitHub is now reporting deprecation warnings about actions that use Node 16. Most (all?) actions we use now seem to have rolled out Node 20 versions which we can drop in.

**To Reproduce**
Look at reports from github actions.

tztsai:
# Description

This is to fix the deprecation issue of `Node` in the CI workflow. Versions of GitHub actions `checkout@v3` and `setup-python@v4` are updated to `v4` and `v5` respectively.

## Type of change

## Key checklist

### DIFF
--- a/.github/workflows/pyrealm_ci.yaml
+++ b/.github/workflows/pyrealm_ci.yaml
@@ -6,8 +6,8 @@ jobs:
   qa:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: "3.11"
       - uses: pre-commit/action@v3.0.0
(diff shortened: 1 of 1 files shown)


.github/workflows/pyrealm_ci.yaml
```diff
@@ -6,8 +6,8 @@ jobs:
   qa:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
```
davidorme, line 10: The actions are used in more than one job, so we need to update `test` and `docs_build` as well as `qa`.

davidorme (changes requested):
We need to update other jobs too.

.github/workflows/pyrealm_ci.yaml
tztsai: > The actions are used in more than one job, so we need to update `test` and `docs_build` as well as `qa`.

Thanks for the catch! I've updated these ones now.

davidorme (approved):
LGTM

PR #181, opened 2024-02-06
Profiling of the `Splash` module

### LINKED ISSUE #166: Add `splash` to profiling suite.
With the merging of #69, it would be good to add the SPLASH module to the profiling suite so we can monitor it for performance regression and choke points.

Take the example analysis in `docs/sources/users/splash.md` and convert it into a profiling test. That dataset isn't particularly big, so we might want to consider something with a heavier load at some point, but it should be fine for now.

casterkay:
# Description

Profiling tests are now put into the `tests/profiling` folder. A test file `test_profiling_splash.py` is created in this folder to profile the performance of the `Splash` module.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/profiling/test_profiling_pmodel.py
+++ b/tests/profiling/test_profiling_pmodel.py
@@ -120,7 +120,7 @@ class TestClass:
 
         comp.summarize()
 
-    def test_profile_subdaily(self, setup):
+    def test_profiling_subdaily(self, setup):
         """Profiling the subdaily submodule."""
         from pyrealm.pmodel import FastSlowPModel, FastSlowScaler
 
--- a/tests/profiling/test_profiling_splash.py
+++ b/tests/profiling/test_profiling_splash.py
@@ -1,0 +1,53 @@
+"""Profiling test class.
+
+Runs a profiler on the splash model implementation to identify runtime bottlenecks.
+"""
+
+from importlib import resources
+
+import numpy as np
+import pytest
+import xarray
+
+from pyrealm.core.calendar import Calendar
+from pyrealm.splash.splash import SplashModel
+
+
+@pytest.fixture
+def splash_core_constants():
+    """Provide constants using SPLASH original defaults.
+
+    SPLASH v1 uses 15°C / 288.1 5 K in the standard atmosphere definition and uses the
+    Chen method for calculating water density.
+    """
+    from pyrealm.constants import CoreConst
+
+    return CoreConst(k_To=288.15, water_density_method="chen")
+
+
+@pytest.mark.profiling
+class TestClass:
+    """Test class for the profiler running on the pmodel implementation."""
+
+    @pytest.fixture()
+    def setup(self, splash_core_constants):
+        """Set up the the splash model and load the test data set."""
+        dpath = resources.files("pyrealm_build_data.splash")
+        data = xarray.load_dataset(dpath / "splash_test_grid_nw_us.nc")
+
+        self.splash = SplashModel(
+            lat=np.broadcast_to(data.lat.data[None, :, None], data.sf.data.shape),
+            elv=np.broadcast_to(data.elev.data[None, :, :], data.sf.data.shape),
+            dates=Calendar(data.time.data),
+            sf=data.sf.data,
+            tc=data.tmp.data,
+            pn=data.pre.data,
+            core_const=splash_core_constants,
+        )
+
+    def test_profiling_calculate_soil_moisture(self, setup):
+        """Profile the calculate_soil_moisture method of the splash model."""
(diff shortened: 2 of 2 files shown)


profiling/report.py
```diff
@@ -39,37 +49,51 @@
...
+)
+df.dropna(inplace=True)
+if args.selector:
+    df = df[df.filename.str.contains(args.selector)]
+i = next(
+    i
+    for i in range(99)
+    if (lambda ss: any(s[i] != ss[0][i] for s in ss))(df.filename.iloc)
+)
+df["filename"] = [s[i:] for s in df.filename]  # remove common prefix
```
davidorme, line 76: What is this doing? And it looks like a lambda is being used as an anonymous function? This is not easy to follow.

davidorme (changes requested):
Adding the profiling of the SPLASH looks good but there is a lot of additional complexity appearing in this profiling report. This doesn't impact the CI workflow at all but it isn't clear how it is useful. It either needs removing or a whole bunch of documentation needs adding to explain how developers might use these arguments and why.

profiling/report.py
tztsai: `i` is the maximum index such that `s[i]` is the same for every filename `s`. Therefore these lines are used to remove the common prefix of all the filenames in `df`.

profiling/report.py
```diff
@@ -15,7 +15,16 @@
...
+    "--no-save", help="Do not save the profiling results", action="store_true"
+)
+parser.add_argument(
+    "--no-plot", help="Do not plot the profiling results", action="store_true"
+)
+parser.add_argument(
+    "--selector", help="Filter the filenames by regex, e.g. 'splash', 'pmodel'"
+)
+parser.add_argument(
+    "--max-benchmark", help="Max number of benchmarks to plot", type=int, default=5
```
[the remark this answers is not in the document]
tztsai: I agree - I will convert these changes to a new PR with more comments. I also think it would be cleaner to convert the code into a script with these arguments.

profiling/report.py
davidorme: Can you give a concrete example of the data is being changed here?

tztsai: These are several filenames before the change:
```
D:\Personal\Projects\pyrealm\pyrealm\pmodel\pmodel.py
D:\Personal\Projects\pyrealm\tests\profiling\test_profiling_pmodel.py
D:\Personal\Projects\pyrealm\pyrealm\pmodel\pmodel.py
D:\Personal\Projects\pyrealm\pyrealm\pmodel\subdaily.py
D:\Personal\Projects\pyrealm\pyrealm\pmodel\jmax_limitation.py
D:\Personal\Projects\pyrealm\pyrealm\pmodel\jmax_limitation.py
D:\Personal\Projects\pyrealm\pyrealm\splash\splash.py
D:\Personal\Projects\pyrealm\tests\profiling\test_profiling_splash.py
```

And after the change:
```
pyrealm\pmodel\pmodel.py
tests\profiling\test_profiling_pmodel.py
pyrealm\pmodel\pmodel.py
pyrealm\pmodel\subdaily.py
pyrealm\pmodel\jmax_limitation.py
pyrealm\pmodel\jmax_limitation.py
pyrealm\splash\splash.py
tests\profiling\test_profiling_splash.py
 ```

Comment (davidorme) [reply profiling/report.py]: Seems like it would be easier and faster to get the shared path through `__file__` rather than this!

Comment (tztsai) [reply profiling/report.py]: I did this initially but after I installed the package using `pip install -e .`, the files are located somewhere in the `site-packages` folder of the python environment...

Comment (davidorme) [reply profiling/report.py]: But that must still be discoverable from `Path` objects rather than this.

Comment (tztsai) [reply profiling/report.py]: Yeah I agree

File: profiling/report.py [hunk 3]
```diff
@@ -39,37 +49,51 @@
...
+p.print_stats("pyrealm")
 report = sio.getvalue()
+print(report)
 
 # Convert to a DataFrame and save to CSV
-report = report.replace(str(root / "*")[:-1], "")  # remove common path
 _, report = re.split(r"\n(?= +ncalls)", report, 1)  # remove header
 df = (
-    pd.read_csv(StringIO(report), sep=" +", engine="python")
+    pd.read_csv(StringIO(report), sep=" +", engine="python", on_bad_lines="skip")
```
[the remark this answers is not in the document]
tztsai: These lambda functions are also not in the `pyrealm` package, so I think they are probably irrelevant in the profiling report.

profiling/report.py
davidorme: This is resolved as no longer part of this PR.

davidorme (approved):
This slimmed down PR looks good to me.

PR #186, opened 2024-02-23
Fix time handling in Splash

### LINKED ISSUE #165: Fix time handling in SPLASH
At the moment, SPLASH assumes a fixed year length for `SplashModel.estimate_initial_soil_moisture`, but this needs to shift to a calendar aware implementation.

Although the original SPLASH v1 used 1st January as the year start in example, I don't think we care about the first day of the time series, just that we have a year's worth of data to equilibrate.

The code to tackle starts here:

https://github.com/ImperialCollegeLondon/pyrealm/blob/8c8698513982d985ea215d5fbc575958167ccd0c/pyrealm/splash/splash.py#L147-L164

casterkay:
# Description

The update uses `CalendarDay.days_in_year` to find the number of days after a whole year in the `estimate_initial_soil_moisture` method of `SplashModel`. An additional check is also added in `__init__` to ensure that the length of `dates` matches the shape of other variables.

Fixes #165 

## Type of change

## Key checklist

## Further checks


pyrealm/splash/splash.py
```diff
@@ -72,6 +72,9 @@ def __init__(
         self.shape: tuple = check_input_shapes(elv, lat, sf, tc, pn)
         """The array shape of the input variables"""
 
+        if len(dates) != self.shape[0]:
+            raise ValueError("Number of dates must match the first dimension of inputs")
+
```
davidorme, line 77: Good catch - can you add units tests to demonstrate though.

pyrealm/splash/splash.py
```diff
@@ -141,18 +144,20 @@ def estimate_initial_soil_moisture(
 
         if wn_init is not None:
             # Check the shape is the same as the shape of a slice along axis 0
-            if wn_init.shape != self.tc[0].shape:
+            if wn_init.shape != self.shape[1:]:
```
davidorme, line 147: Is there a reason for making these changes to the shape checking (and also `np.arange` to `range`). As far as I can see they behave identically, so I'm not sure why they have been altered.

pyrealm/splash/splash.py
```diff
@@ -162,7 +167,7 @@ def estimate_initial_soil_moisture(
 
             # Loop over the calendar object, updating the soil_moisture array
             wn_day = wn_start
-            for day_idx in np.arange(366):
+            for day_idx in range(day1.days_in_year):
                 # Calculate aet, soil moisture and runoff:
```
davidorme, line 171: This is difficult - what if the first day is '1999-12-30' - that is then going to track forward 365 days in the data to '2000-12-28' because 2000 is a leap year. 

I know we've chatted about minor errors like this - it's also hard because the original implementation is not well set up for this. The original `main.py` assumes it is being fed chunks of data that are a year long and which start on January 1st - but doesn't enforce the length of the data. So - very hard to match the behaviour even if we wanted to.

The basic original behaviour though is to take a a time series (assumed 'XXXX-01-01' to 'XXXX-12-31') and then iterate until those two days have the same value. That's odd - I'd expect it to loop the extra day to make the "same" day match.

Either way, I'm not 100% sure what is better here, but this approach does definitely need documenting in the docstring and commenting to explain what the decision is (see also updating the comments on L155) and then testing to check that shifting the dates in the test benchmarks doesn't alter anything.

davidorme (changes requested):
This isn't elegant but it is hard to know what is 'right' - I don't think the original implementation gives us any clarity here on what should happen, because this simply isn't considered.

Some unexpected changes and also some testing needed.

pyrealm/splash/splash.py
tztsai: I feel it's better to use `self.shape` and make the shape checking independent of a specific variable name. The replacements of `np.arange` by `range` happen to occur in the same lines. I'm not sure why `np.arange` is preferred over `range`.

pyrealm/splash/splash.py
tztsai: I have fixed this with `pandas.Timestamp` and `pandas.DateOffset`. This can address non-January-1st start date and leap year, including starting with Feb 29th in a leap year.

pyrealm/splash/splash.py
tztsai: Yes I will add unit tests for this check.

tztsai: I've added a check in the `splash_model` fixture in the unit test

davidorme (changes requested):
Minor change to testing requested!

tests/unit/splash/test_splash_model.py
```diff
@@ -82,6 +82,7 @@ def splash_model(grid_benchmarks, splash_core_constants, calendar):
     )
 
     assert splash.shape == (len(ds.time), *ds.elev.shape)
+    assert splash.shape[0] == len(calendar)
     return splash
```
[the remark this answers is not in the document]
mondus: Yes I agree (and discussed in dev meeting). There should be an additional test to ensure the correct Exception is raised.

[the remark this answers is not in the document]
tztsai: I have added the test now.

tests/unit/splash/test_splash_model.py
```diff
@@ -67,19 +67,29 @@ def test_splash_model_init(splash_core_constants, grid_benchmarks, var, flag):
...
-        pn=ds.pre.data,
-        core_const=splash_core_constants,
-    )
+    for test_time_index_check in [True, False]:
+        if test_time_index_check:
+            dates = Calendar(calendar.dates[:-1])
+            context = pytest.raises(ValueError)
+        else:
+            dates = calendar
+            context = nullcontext()
```
davidorme, line 81: I don't think this is in the right place. 

Fixtures provide shared data and functionality to different tests and I don't think are intended to implement actual testing. Obviously you _can_ put an `assert` in here, but this fixture is created independently multiple times (it has "function" scope), so this test is repeated every time the fixture is used.

I'd also tend to implement the `test_time_index_check` using `pytest.mark.parameterise` rather than an internal loop, but that's more minor.

davidorme (changes requested):
The functionality still looks good, but the unit test - although it checks the code in a sensible way - is embedded in a fixture, which seems odd.

tests/unit/splash/test_splash_model.py
tztsai: I've put it into `test_splash_model_init`, using `dates` as an additional parameter now.

davidorme: Thanks - that looks better!

tests/unit/splash/test_splash_model.py
```diff
@@ -43,19 +43,28 @@ def test_splash_model_init(splash_core_constants, grid_benchmarks, var, flag):
...
+            ds = ds.sel(time=dates[:-1])
+        context = pytest.raises(ValueError)
+    else:
+        vmin, vmax = bounds[var]
+        arr = ds[var].data
+        if flag < 0:
+            arr.flat[np.random.choice(arr.size)] = vmin - 1e-4
+        else:
+            arr.flat[np.random.choice(arr.size)] = vmax + 1e-4
+        context = pytest.warns(UserWarning)
```
davidorme, line 61: I'd generally go for more test functions (even if repetitious) rather than create branching patterns within tests, but this isn't the most complex internals I've ever seen 😄. If you are going to do this, it does make sense to add brief comments to explain what each branch does.

davidorme (approved):
Looks good to me - would still like more comments but this is clear enough.

davidorme:
I've resolved the `pyproject.toml` and `poetry.lock` conflict - just different package order in the `pyproject.toml` dev dependencies

tests/unit/splash/test_splash_model.py
tztsai: I've added the comments now.

PR #188, opened 2024-02-28
Update `profiling/report.py` and its documentation

tztsai:
# Description

The PR focuses on updating the profiling/report.py file and its documentation within the pyrealm project. It introduces new optional arguments to the script for enhanced control over profiling results, including options to avoid saving the results or generating plots, to filter filenames via regex, and to limit the number of benchmarks in the plots. Additionally, it makes changes to the documentation to reflect these updates and provides information on the usage of these new arguments.

## Type of change

## Key checklist

## Further checks


profiling/report.py
```diff
@@ -39,37 +49,46 @@
...
+)
+df.dropna(inplace=True)
+if args.selector:
+    df = df[df.filename.str.contains(args.selector)]
+i_diff = next(i for i, cs in enumerate(zip(*df.filename)) if len(set(cs)) > 1)
+df["filename"] = [s[i_diff:] for s in df.filename]  # remove common prefix
 dt = datetime.datetime.fromtimestamp(prof_path.stat().st_mtime)
 df.index = pd.Series([dt.strftime("%Y-%m-%d %H:%M:%S")] * len(df), name="timestamp")
 print(df)
 df["label"] = df["filename"].str.extract(r"(\w+).py").squeeze() + "." + df["function"]
```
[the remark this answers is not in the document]
tztsai: I will refactor the code into funcitons and adopt block comments as well.

profiling/report.py
```diff
@@ -13,80 +12,162 @@
...
+        df["filename"].str.extract(r"(\w+).py").squeeze() + "." + df["function"]
+    )
+
+    # Add GitHub event name for information
+    df["event"] = args.event
+
+    return df
+
+
+def plot_profiling(df: pd.DataFrame, save_path: str | bool = False) -> None:
```
davidorme, line 105: We've not moved to Python 3.10 yet, so this needs to be `Union[str, bool]` - I'm also not sure why it can't just be `save_path: str = ""`. I'm not 100% keen on the Python `falsy` values for tests but it seems simpler here.

profiling/report.py
```diff
@@ -13,80 +12,162 @@
...
+    df.plot.barh(
+        y=["tottime_percall", "cumtime_percall"], x="label", figsize=(20, 10)
+    )  # type: ignore
+    plt.ylabel("")
+    plt.gca().invert_yaxis()
+    plt.tight_layout()
+    plt.legend(loc="lower right")
+
+    if args.show:
+        plt.show()
```
davidorme, line 118: This should be a function argument not from the `args` local variable. Is this here to allow the plot to be shown in a plot window? It is probably easier for the function to always write to file and not make this an option.

profiling/report.py
```diff
@@ -13,80 +12,162 @@
...
+        .sort_values(by="cumtime_percall", ascending=False)
+        .query("cumtime > 0.01")  # drop insignificant functions
+        .dropna()
+    )
+
+    return df
+
+
+def process_df_columns(df: pd.DataFrame, path_filter: str = "") -> pd.DataFrame:
+    """Process the DataFrame columns including file name, function name, etc."""
```
davidorme, line 75: Need to docstring the arguments

davidorme (changes requested):
Not sure if this is still a work in progress - no review requested - but some thoughts. 

The `report.py` is still partially written as a script - there is code in there that executes whenever the file is run. I think it would be better to wrap the command line argument into a separate function and use the `if __name__ == "__main__":` approach to make that run only if the file is run as a script. That means that the functions can be imported and used elsewhere without side effects:

Something like:

```python

def profile_report_cli() -> None:
    """Runs the profile report function with a command line interface."""

    parser = ArgumentParser()
    parser.add_argument("--event", help="Github event name", default="none")
    parser.add_argument(
        "--profile_file",
        help="Profiling results directory",
        default="prof",
        type=Path,
    )
    parser.add_argument(
        "--no-save", help="Do not save the profiling results", action="store_true"
    )
    parser.add_argument(
        "--no-plot", help="Do not plot the profiling results", action="store_true"
    )
    parser.add_argument(
        "--selector", help="Filter the filenames by regex, e.g. 'splash', 'pmodel'"
    )
    parser.add_argument(
        "--max-benchmark", help="Max number of benchmarks to plot", type=int, default=5
    )
    args = parser.parse_args()

    read_prof_as_dataframe(prof_path=args.profile_path)
    ...


if __name__ == "__main__":
    profile_report_cli()
```

Obviously there are loads of ways to break this up, but I think the natural breakdown of the functions are:

`convert_prof_to_csv(profile_path, csv_path, ...) -> None:` : appends the recent data to the CSV
`read_profiling(csv_path, n_runs) -> pd.DataFrame:`: extracts a dataframe from the CSV
`run_benchmark(df, c...) -> bool:`: does the dataframe pass benchmark tests
`plot_benchmark(df, png_file, ...) -> None:`: Plots the loaded benchmarks

It's good to keep logical steps separate: `benchmark` and `plot_and_check_benchmark` seem mixed up for example.

profiling/report.py
tztsai: I've passed `args` containing configs like `save_path` instead.

davidorme (approved):
LGTM

PR #190, opened 2024-03-06
Fastslow initval

### LINKED ISSUE #82: Allow FastSlowModel to take initial realised values

When a FastSlowModel is fitted, the first 'realised' values are equal to the optimal value at time $t=0$ and thereafter the memory effect is used to impose a lag. If you break up a time series into chunks, then each chunk will 'reset' to the optimal value at the start of the chunk, so there is a brief period where a memory reset happens.

The FastSlowPModel could accept a dictionary of realised $V_{cmax}$, $J_{max}$ and $\xi$ values, which are used instead of the optimal value at $t=0$. This would allow a time series to maintain a continuous memory effect, even if the models are fitted in chunks. The class would also provide a method to optionally extract that dictionary from a fitted FastSlowPModel.

You would still need to run the time chunks in series (not in parallel), to be able to feed those outputs into the next time chunk.

surbhigoel77:
# Description

The `subdaily` module  uses an extended form of PModel called the `FastSlowPModel`. to estimate two categories of variables - fast variables (that respond to environmental changes instantaneously) and slow variables (that take time to respond to these changes). The module uses a `memory_effect` functions that facilitates the calculation of the slow variables as a combination of  their lag values and optimal values. 

In the current implementation of `subdaily`, at the start of the model run, the slow variables take the optimal value as the initial value and update the succeeding values using the:

$R_{t} = R_{t-1}(1 - \alpha) + O_{t} \alpha$

This PR intends to use an approach in which the last available realised value is used as the initial value for $\xi$, $V_{max}$ and $J_{max}$.

[ETA - @davidorme 2024-10-18]

In addition to adjusting the equation above (so that $R_1$ can be updated from the provided previous values $R_0$ and not just be set as $O_1$), the previous values _also_ need to be supplied to `SubdailyScaler.fill_daily_to_subdaily` to replace the leading `np.nan` values before the first acclimation window.

Fixes #82 

## Type of change

## Key checklist

## Further checks


davidorme:
This is a WIP, @surbhigoel77 ? Can you convert to a draft (link in Reviewers pane).

davidorme:
Also - have you branched this off the `feature/new-unit-tests` branch, rather than `develop`? I think it is bringing the commits to that PR with it?

surbhigoel77:
> Also - have you branched this off the `feature/new-unit-tests` branch, rather than `develop`? I think it is bringing the commits to that PR with it?

Yes, I realised that when I pushed changes to `fastslow-initval`. The commit history shows changes from `feature/new-unit-tests` branch as well.

davidorme:
No bother - it isn't a big deal 😄  I guess we could `git revert` those commits on this branch

pyrealm/pmodel/subdaily.py
```diff
@@ -229,6 +229,7 @@ def __init__(
         allow_holdover: bool = False,
         allow_partial_data: bool = False,
         fill_kind: str = "previous",
+        init_realised: tuple[NDArray, NDArray, NDArray] | None = None,
     ) -> None:
```
davidorme, line 233: Docstring needs updating to include this argument.

pyrealm/pmodel/subdaily.py
```diff
@@ -322,6 +323,14 @@ def __init__(
             1 / calc_ftemp_arrh(tk_acclim, self.env.pmodel_const.subdaily_jmax25_ha)
         )
 
+        """Instantaneous optimal :math:`x_{i}`, :math:`V_{cmax}` and :math:`J_{max}`"""
+        if init_realised is not None:
+            self.init_xi_real, self.init_vcmax_real, self.init_jmax_real = init_realised
```
davidorme, line 328: This needs to validate the provided values - mostly just that they are the correct shape.  I'm not sure they need to be stored as class attributes.

pyrealm/pmodel/subdaily.py
```diff
@@ -322,6 +323,14 @@ def __init__(
...
+        else:
+            self.init_xi_real = self.pmodel_acclim.optchi.xi
+            self.init_vcmax_real = self.vcmax25_opt
+            self.init_jmax_real = self.jmax25_opt
+
         # 5) Calculate the realised daily values from the instantaneous optimal values
         self.xi_real: NDArray = memory_effect(
             self.pmodel_acclim.optchi.xi, alpha=alpha, allow_holdover=allow_holdover
```
davidorme, line 336: The init values need to be passed into `memory_effect` - they should replace the initial assumption that $R_{0} = O_{0}$ at $t=0$. So I think what is needed is an `initial_values: NDArray | None = None` argument to `memory_effect` and then the line `memory_values[0] = values[0]` can conditionally change to `memory_values[0] = initial_values`.

A sensible test for this would be to take the existing test and then split it into two time chunks - feeding the last realised value from the first chunk into the initial values of the second chunk should then duplicate the results of an all in one run.

davidorme (changes requested):
Definitely the right inputs, but then we need to feed those forward into the calculations. This is a little more finicky than I first thought - I'm going to take it off the list for version 1.0.0 (#206)

pyrealm/pmodel/subdaily.py
MarionBWeinzierl: Do you mean something like
```
if init_realised is not None:
    assert(init_realised[0].shape() ==self.xi_real.shape())
    assert(init_realised[1].shape() == self.vcmax25_real.shape())
    assert(init_realised[2].shape() == self.jmax25_real.shape())
    init_xi_real, init_vcmax_real, init_jmax_real = init_realised
```

davidorme: Something like that yes - but with an Exception not `assert` 😄

MarionBWeinzierl: OK. Why would you use an exception here?

davidorme: IIUC `assert` shouldn't be used in production code, because it is sensitive to the way in which `python` is invoked. If the code is run using `python -o` or `python -oo` then those assert statements are discarded and any bad data would sail through.

https://docs.python.org/3/using/cmdline.html#cmdoption-O

pyrealm/pmodel/subdaily.py
MarionBWeinzierl: Attempted to implement in https://github.com/ImperialCollegeLondon/pyrealm/pull/190/commits/9e3d36d326bbcdfdd5aee00218e5b72877f2b8fa

MarionBWeinzierl: There is a mismatch in dimensions when assigning the values in `memory_effect` : `values` , `initial_values`  (as per your suggestion) and `memory_values`  have the same dimension. So assigning `memory_values[0] = initial_values` fails. Should it be `memory_values[0] = initial_values[0]` ? And should all the remaining assignments for `memory_values` stay as they are (following `values`, or go to `initial_values`?

davidorme: The arrays of data in `subdaily` can have multiple dimensions but the first axis must be time. When users are providing initial values, those need to be a single set of observations for the first time point, so should have the shape of a slice along that time axis. I think when that is enforced, the original assignment is correct. 

So the input checking for user values should be:

```python
In [59]: data = np.arange(12).reshape((3,4))
In [60]: data
Out[60]: 
array([[ 0,  1,  2,  3],
       [ 4,  5,  6,  7],
       [ 8,  9, 10, 11]])
In [61]: init = np.arange(4)
In [62]: init
Out[62]: array([0, 1, 2, 3])
In [63]: data[0].shape == init.shape
Out[63]: True
```

pyrealm/pmodel/subdaily.py
MarionBWeinzierl: I changed the shape check to

```
        if init_realised is not None:
            if not (
                (init_realised[0].shape() == self.xi_real[0].shape())
                and (init_realised[1].shape() == self.vcmax25_real[0].shape())
                and (init_realised[2].shape() == self.jmax25_real[0].shape())
            ):
                raise Exception("`init_realised` has wrong shape in Subdaily PModel")
            else:
                init_xi_real, init_vcmax_real, init_jmax_real = init_realised
        else:
            init_xi_real, init_vcmax_real, init_jmax_real = [None, None, None]
```

Does that make sense?

pyrealm/pmodel/subdaily.py
MarionBWeinzierl: After our discussion I corrected that bit back, and adapted the assignment bit as shown in https://github.com/ImperialCollegeLondon/pyrealm/pull/190#discussion_r1786448078, in this commit: https://github.com/ImperialCollegeLondon/pyrealm/pull/190/commits/504ad7410f6bedb7d591289c712f7be650826663 .

MarionBWeinzierl: I implemented a test following your suggestion as follows: 


```
def test_memory_effect_chunked(inputs_whole, alpha):
    from pyrealm.pmodel import memory_effect

    result_whole = memory_effect(inputs_whole, alpha=alpha)

    [inputs_chunk1, inputs_chunk2] = np.split(inputs_whole, [5], axis=0)

    result_chunk1 = memory_effect(inputs_chunk1, alpha=alpha)

    result_chunk2 = memory_effect(
        inputs_chunk2, initial_values=result_chunk1[-1], alpha=alpha
    )

    assert np.allclose(result_whole[-1], result_chunk2[-1])
```

However, `np.allclose` fails -- the values are "kind of close", but different. For example, for the 2D test, it compares `array([ 8.001953,  9.001953, 10.001953, 11.001953], dtype=float32)` and `array([ 7.9414062,  8.941406 ,  9.941406 , 10.941406 ], dtype=float32))` .

MarionBWeinzierl: From @davidorme : 

"If we do this:
```
    memory_values = np.empty_like(values, dtype=np.float32)
    if initial_values is None:
        memory_values[0] = values[0]
    else:
        memory_values[0] = initial_values * (1 - alpha) + values[0] * alpha
```
Then the current test works as expected. I think that’s a better API. People might want to feed in a year at a time for example, and having to get the overlapping observation is finicky. We might then want to rename initial_values to previous_values or something similar to make this clear. "

davidorme:
@MarionBWeinzierl It looks good - I've added a higher level test to run it through the whole SubdailyPModel process. That revealed a couple of minor issues in the value checking. I've also updated the variable name and docstrings in `SubdailyPModel`.

Unfortunately, the new test is failing for reasons I don't quite understand yet.

davidorme:
OK - I figured out the problem and fixed it. I've updated the original PR description to cover the expanded PR coverage.

davidorme (approved):

PR #193, opened 2024-03-13
Handle NaNs when calculating max difference in `Splash.estimate_initial_soil_moisture`

### LINKED ISSUE #192: Verbose mode in SPLASH `estimate_initial_soil_moisture` doesn't handle `np.nan`
**Describe the bug**

Running SPLASH using arrays that contain some `np.nan` values (which is an expected use case - could be water cells in a grid)  outputs the following when running `estimate_initial_soil_moisture`:

```python
In [84]: init_soil_moisture = splash.estimate_initial_soil_moisture(verbose=True)
Iteration: 1; maximum difference: nan
Iteration: 2; maximum difference: nan
```

**Expected behavior**

That `verbose` behaviour should use `np.nanmax` to give more meaningful progress to convergence.

**Edited to add:**

To extend this issue with a couple more linked changes: 

* the error message on failure should also handle `np.nan`
* the function should optionally return the maximum difference that is calculated to estimate convergence as a debugging tool. In this use case, I think it could return the calculated soil moisture *whether or not it has converged* - so it runs for `n_iter` and gives back where it got to. 

  Actually - giving this more thought -  I think this probably should add a `return_convergence: bool = False` argument. When that flag is `True`, the code creates an array with an `n_iter` sized dimension on the first axis that is used to return the successive guesses at `initial_soil_moisture`. That can then be used to explore convergence issues across the input dimensions much more cleanly and keeps the return type as a single `np.array`

casterkay:
# Description

I replaced `np.max` with `np.nanmax` when calculating the maximum difference between the start of year soil moisture and the final day of the year and then update the start point to the end of year in the `estimate_initial_soil_moisture` method of the `Splash` model, to output more meaningful error messages when running this method.

Fixes #192

## Type of change

## Key checklist

## Further checks


davidorme (commented):
@tztsai This looks good - I was thinking about the use case and decided that the `return_convergence` option should _always_ return (otherwise it can only be used when convergence is successful).

I've pushed a commit with those changes - I wanted to think through what I was suggesting - and  added some docs. I've also changed your `noqa` to make it more strict - it will fail again if someone makes the function more complex.

Do those changes make sense? We then need to add unit tests of this functionality, but that should probably be a separate PR rather than piling more on this one.

tztsai:
Shouldn't the function inform whether the iteration has reached equilibrium before returning the array?

davidorme:
Yeah  - that would be useful - we should add a warning to flag that the returned sequence doesn't meet the convergence criteria.

davidorme (approved):
All looks good to me. Fine to merge once #186 has gone in.

PR #208, opened 2024-04-05
Fixing profiling issues

### LINKED ISSUE #207: Profiling issues
There are some issues with the profiling system that it would be good to fix up for 1.0.0. 

I think the list below is going to grow - we might have some revealed bugs to work on.

- [x] The `graphviz` requirement for profiling is currently in the main package requirements (`[tool.poetry.dependencies]`) and this is causing package installation to break. Trying to install the test release (`pip install -i https://test.pypi.org/simple/ pyrealm==1.0.1a3`) breaks with: `ERROR: Could not find a version that satisfies the requirement graphviz (from pyrealm) (from versions: none)` and we really don't want `graphviz` as a _package_ requirement anyway.
- [x] The profiling isn't running - see for instance [this action job](https://github.com/ImperialCollegeLondon/pyrealm/actions/runs/8558378594/job/23453472431). I think this is because the following line refers to a step that no longer exists, so can never equal `success`. https://github.com/ImperialCollegeLondon/pyrealm/blob/971d1a3c492df3fcaa84279ed7a2f74275e6169e/.github/workflows/pyrealm_ci.yaml#L60
- [x] That step runs on _every_ matrix combination in the `test` job. There's value in looking for performance hits in different python versions, but we'd have to extend the profiling database to include matrix details. That seems like overkill for now.
- [x] The profiling output for a run is tagged by date and github event name (typically `push`), but it would be more useful to tag by commit SHA.
- [ ] I'm still not sure about _when_ this runs - it is currently on a ton of events (push, etc), which seems overkill. We may want to break this out into a separate job that only runs on merge into `develop` or `main`? **ETA**: I'm not sure this is avoidable - every push to a PR results in a commit that _could_ be merged to `main` or `develop`, so needs to pass this. The alternative is probably `cron` jobs to do regular checks on the `develop` branch.
- [x] The current code filters the profiling results using the text `pyrealm`, but this runs into issues when running locally, particularly with local `poetry` virtual environments, which include `pyrealm` in the path to site packages. The filtering needs to be more controlled.
- [x] There isn't a clear mechanism to handle benchmarking failures that we want to ignore - simply overriding the test on GitHub doesn't leave much of a trace.
- [x] The process id information used for profiling (file, lineno, function name) are not portable between commits for benchmarking. Method names are not qualified by their class and can method names be duplicated within a source file, and line numbers of function definitions are not stable between commits.
- [x] The workflows for manual benchmarking and CI benchmarking need to be revised and documented to give a longer explanation of the process and clear models of how to use the profiling.

davidorme:
# Description

This PR is to resolve the profiling issues described in #207. It got a bit larger than expected, but there were a few interlinked issues to give clean profiling and benchmarking workflows.

* Updates broken data path in `splash` profiling test.
* Updates profiling CI steps to remove reference to removed `lfs` step and to run only on one OS/python version.
* Renames `report.py` to a more descriptive `run_benchmarking.py` and revises the command line structure.
* Updates the benchmarking functions:
    * Adds a new function `get_function_map` that uses `ast` to label profiled processes by their position in the package tree. This is mostly so that repeated method names within a source file can told apart by their class name and not their line number (which is unstable across versions).
    * Merges `read_prof_as_dataframe` and `process_report_columns` into `convert_and_filter_prof_file`: these functions would always need to be run together.
    * The `plot_profiling` and `plot_benchmark` have now been merged into `create_benchmark_plot`, which is a single plot showing relative performance of individual calls across versions.
    * Adds a `generate_test_database` function, which is mostly just to keep a handy recipe for test inputs when we need to revise this process.
    * Reworked the CLI arguments to provide cleaner manual and automated workflows.
    * Benchmarking is now carried out on all package callables - we might need to alter this, but starting with everything.
* Updates the description of the process in `CONTRIBUTING.md`, including manual and automated profiling workflow.
* We need a separate issue to split `CONTRIBUTING.md` up and avoid content duplication with the `docs/development` content. (ETA: #209 )
* Moved the update of the call graph version in the repo (`prof/combined.svg` --> `profiling/call-graph.svg`) out of `report.py` and into a separate CI step - keeps benchmarking code more focussed.
* Update the profiling test structures and tiles the data to increase the run time.
* Moves the profiling actions into their own workflow that only runs on pulls into `develop` and `main`, so only when PRs get merged. This isn't perfect - a merge might turn out to break profiling - but having the profiling on every commit to a PR is too bulky.

I have a broader concern that this benchmarking is going to be continually throwing up issues that arise from different runner specs, but we'll just have to see.

But - that aside: does this all look sane? Does the new graph make sense? Actually, **something is wrong there** as the plot from a previous failed run hasn't been replaced by the most recent passing run.

@tztsai - it would be great if you could have a look at this, but I realise you're on another project.

## Type of change

## Key checklist

## Further checks


davidorme:
The actions are all now passing:
https://github.com/ImperialCollegeLondon/pyrealm/actions/runs/8633260571

I think the auto-commit is triggering a new push that is stalling somehow. Those profiling auto-commits probably don't need any CI - we could add `[skip ci]` to the message?

.github/workflows/pyrealm_ci.yaml
```diff
@@ -55,22 +55,36 @@ jobs:
...
+      id: benchmarking
+      if: ${{steps.profiling.outcome == 'success'}}
+      run: poetry run python profiling/run_benchmarking.py prof/combined.prof profiling/profiling-database.csv profiling/benchmark-fails.csv ${{ env.COMMIT_SHORT_SHA }} --update-on-pass --plot-path profiling/performance-plot.png
+
+    # If the benchmarking succeeds, then update the call graph. This again only runs on
+    # the ubuntu/3.10 runner used for the profiling.
+    - name: Update call graph
+      id: update_call_graph
+      if: ${{steps.benchmarking.outcome == 'success'}}
+      run: cp prof/combined.svg profiling/call-graph.svg
```
tztsai, line 79: Maybe this command can be appended as a subsequent command in the last step (Benchmark performance)? If the `poetry run ...` command fails, then this command will be automatically skipped.

.github/workflows/pyrealm_ci.yaml
```diff
@@ -55,22 +55,36 @@ jobs:
...
+    
+    # This runs if the benchmark passes or fails (so still only on ubuntu/3.10) and is
+    # skipped on other runners or if the profiling itself fails.
     - uses: stefanzweifel/git-auto-commit-action@v5
-      if: ${{ always() && steps.profiling.outcome != 'skipped' && github.event_name == 'merge_group' }}
+      if: ${{steps.benchmarking.outcome != 'skipped'}}
```
tztsai, line 84: If `always()` is removed, this step will not be executed when the benchmark step fails, so I think `benchmark-fails.csv` will not get updated by the auto-commit.

davidorme: Yup - I'd forgotten that `...outcome == 'failed'` will skip the step.

.github/workflows/pyrealm_ci.yaml
davidorme: The intention was to deliberately only update the call graph on passing benchmark tests - but I'm not sure that's necessarily what we want, thinking about it?

I guess my reasoning is that the call graph shows the profiling of a single version of the code - and we'd like the website to show only the profile of the current release that passes everything. But only passing versions are going to get merged, so we could just copy every time.

davidorme: Yeah - on reflection, I think that is better.

davidorme:
I think we’ve got issues with the benchmarking. I think it is now working as we intended but three runs of essentially the same code (only the profiling CI workflow is changing) is leading to wildly different relative run times. The two profiling graphs being updated in this commit (https://github.com/ImperialCollegeLondon/pyrealm/pull/208/commits/d6dff3497a1b3400e09b7225ac565be012e6c9b3) show the issue.


It could be that the profiling tests have too small a load - they run really fast - to give consistent behaviour or it could be that I’ve done some mad randomisation of the sort order. I don’t think that’s the case though - I manually triggered fails in testing by altering the database and the correct processes failed the benchmarking. My guess is that runner architecture is going to make this process hard to use within CI. My intuition is that we need a single benchmarking machine to run tests?

davidorme:
Also note that the failed `run_benchmarking.py` has clobbered the next line in the `run` section, so the call graph is not copied. That needs its own step.

MarionBWeinzierl:
I would suggest that we should try going back to a bigger problem size, to make sure we exclude random noise being a major factor in the runtimes.

davidorme:
I agree - I think we can simply tile the current inputs to increase the load. A couple of other things:

1. Those plots clearly show increasing variance form the top (longest running) to the bottom (shortest running). I think the code before only looked at calls that took over a threshold value, but I wanted to try and get a more holistic view of execution time. We could put in some kind of tuneable runtime filter? Only include the processes that account for 95% of the total runtime or similar.
2. I think I've also switched from using the maximum previous value for a process as the target to the minimum. I don't like using the maximum but we could use another stat - maybe something to do with the range.

MarionBWeinzierl:
As discussed on Slack, we might also want to reduce when this is run, i.e., only run on merge on `develop` or `main`

.github/workflows/pyrealm_ci.yaml
```diff
@@ -5,8 +5,8 @@ name: Test and build
 on:
   pull_request:
     types: [opened, synchronize, reopened]
-  push:
-    branches: [main, develop]
+  # push:
```
MarionBWeinzierl, line 8: Why is this commented out? Shall it be removed?

tests/profiling/test_profiling_pmodel.py
```diff
@@ -13,129 +13,163 @@
...
-            tc=self.tc, patm=self.patm, vpd=self.vpd, co2=self.co2
+@pytest.fixture(scope="module")
+def setup():
+    """Setting up the pmodel and loading the test data set.
+
+    This uses a module scoped yield fixture to only load the data once and then to
+    delete it all when the tests exit.
+    """
+
+    print("\n***\nSetting up\n***\n")
```
MarionBWeinzierl, line 24:
```suggestion
    print("\n***\nSetting up profiling of pmodel\n***\n")
```

CONTRIBUTING.md
```diff
@@ -247,6 +225,197 @@ error or warning is raised. See the example for `pyrealm.utilities.convert_rh_to
...
+
+Profiling and benchmarking can be run manually - which is useful if you are working on
+code and want to check it doesn't impact performance - but are also run as part of the
+continuous integration process when code is to be merged into the `develop` or `main`
+branches.
+
+#### Generating profiling data
+
+We use the [pytest-profiling](https://pypi.org/project/pytest-profiling/) plugin to run
+a set of profiling tests and generate profiling data. These tests are located
```
MarionBWeinzierl, line 242:
```suggestion
a set of profiling tests and generate profiling data. These tests are located at
```

MarionBWeinzierl (approved):
Approving to see whether it works. Some comments below.

.github/workflows/pyrealm_ci.yaml
davidorme: Yes - this was an experiment. Previous CI runs each test twice (once for PR and once for pull). I'm not quite sure _why_ the `pull` event onto `main` or `develop` gets triggered but it doesn't seem necessary?

PR #216, opened 2024-04-17
Scaling profiling datasets

### LINKED ISSUE #215: Scaling problems with profiling on GitHub Actions and locally
**Describe the bug**

The PR to scale up the size of the profiling test data to make the results less noisy has gone too far. Actions are crashing out and - although the errors are a bit opaque - it seems likely this is due to the runners running out of memory (16GB on `ubuntu-latest`). @MarionBWeinzierl has seen the same problem running the profiling suite on a 16GB Ubuntu laptop.

I've run some simple peak memory usage estimates locally and changing the `pmodel`  scale up factor `lon` gives the following peak memory usage using `/usr/bin/time -l` outputs:

| pmodel_profile_scaleup | peak memory footprint in GB |
| --- | --- |
| 40 (current) | 35.84 |
| 20 | 19.74 |
| 10 | 9.95|

We need to scale down the tests and it would be useful to make the scaling factors configurable.

**To Reproduce**

See here:

https://github.com/ImperialCollegeLondon/pyrealm/actions/runs/8704343221

davidorme:
# Description

This PR restructures the profiling code to do a number of things:

* Move the data payload fixtures into a `profiling/conftest.py`
* Separate the `pmodel` and `subdaily` profile workflows into separate files
* Add custom command line options to `pytest` to allow the scaling of the payload data to be controlled from the command line. The `subdaily` and `pmodel` tests share the same data and scaling factor option.
* Change the default scaling to something that _ought_ not to break the GitHub runners (max of 16GB RAM).

These changes make the profiling more flexible and helps set up for looking at the scaling of functions with input sizes. It allows commands like the following:

```bash
# Look at the peak memory usage,  run all the profiling tests, using an 8 fold scaling of the inputs
/usr/bin/time -l pytest -m "profiling" --pmodel-profile-scaleup 8
# Look at the peak memory usage,  running the splash profile, with an 100 fold scaling of the inputs
/usr/bin/time -l pytest -m "profiling" tests/profiling/test_profiling_splash.py --splash-profile-scaleup 100
```

The peak memory size of running the full test set is basically the combined size of each test and is _very_ roughly:

50Mb * `splash-profile-scaleup` +  1Gb * `pmodel-profile-scaleup` 

The defaults `--splash-profile-scaleup=125`  and `--pmodel-profile-scaleup=6` should give about a 6GB size for each dataset. If we could use a single dataset, we'd be able to use bigger loads....

Fixes #215

## Type of change

## Key checklist

## Further checks


tests/profiling/conftest.py
```diff
@@ -0,0 +1,117 @@
...
+    # values along the longitude axis - scaling along the years axis is more problematic
+    # because the years need to be properly calendar aware for subdaily models and so
+    # would need to handle leap year sequencing. Actually - no it isn't. You'd just
+    # calculate the number of days and then create the dates as a sequence from the
+    # first date - there would be some days that change year but it would be very very
+    # minor and not really relevant to this test.
+
+    def _scale_up(array, scaling_factor: int):
+        # Broadcast along the last axis
+        return np.broadcast_to(array, list(array.shape[:-1]) + [scaling_factor])
```
tztsai, line 47: Is there a reason why `_scale_up` here uses `broadcast_to` while `_scale_up` in `splash_profile_data` uses `tile`?

tests/profiling/conftest.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # Generate and check the PModelEnvironment
+    pm_env = PModelEnvironment(tc=tc, patm=patm, vpd=vpd, co2=co2)
+
+    yield pm_env, fapar, ppfd, local_time
+
+    print("\n***\nTearing down\n***\n")
+
+    # Delete all the local objects
+    del list(locals().keys())[:]
```
tztsai, line 77: Does not `del` only remove the reference to these objects, which will happen after directly returning the results anyhow?

tztsai:
Perhaps `gc.collect()` can be used between profiling `pmodel` and `splash` so that the peak memory size would presumably be the larger size of `pmodel` and `splash` profiling datasets instead of the sum of their sizes.

tests/profiling/conftest.py
```diff
@@ -0,0 +1,117 @@
+"""Some pytest configuration specific to the profiling test suite."""
+
+from importlib import resources
+
+import numpy as np
+import pytest
+import xarray
+
+
+@pytest.fixture(scope="session")
```
tztsai, line 10: Maybe the `pmodel_profile_data` fixture can be moved to `test_profiling_pmodel` and the "session" scope can be changed to "module" so that this fixture will be destroyed after the `pmodel` profiling test, preventing both `pmodel_profile_data` and `splash_profile_data` occupying the RAM together.

tests/profiling/conftest.py
davidorme: I was scaling on longitude _and_ time - can broadcast one but then need tile to do both (although my broadcast-fu is not strong). I've dropped back to the simple case for now - it would be more efficient to broadcast, I think.

tests/profiling/conftest.py
davidorme: I'm honestly not sure. This is a common pattern used for repeatedly used static fixtures. As I understand it, the `yield` rather than `return`  makes the fixture a generator, so the fixture keeps spitting out the data each time it is called and then the fixture is torn down at the end of the scope. It might be that the fixture deletion removes the locals within the function, but I don't think this is causing any issues.

tests/profiling/conftest.py
davidorme: Oh. Right. @tztsai The reason is that the `pmodel` data has shape (X, Y, 1) - that last singleton axis can be broadcast to (X, Y, N). But the `splash` data is (X, 20, 20) and that doesn't broadcast to (X, 20, 160) but does tile.

There may be a cleverer way to do this!

tests/profiling/conftest.py
davidorme: @tztsai That's a good point - it is tricky though because I've separated tests into modules, so data fixtures are  shared via conftest.py and the execution order is just set by the test file name.

I think the cleanest way to do this is to separate the profiling tests into packages, each of which has its own `conftest.py` with package scoped data fixtures. That should create stable groups of tests that use the same data and should also scope the fixture creation and tear down correctly.

Although - when I actually _do_ that I don't seem to see a shift in peak memory usage.

tests/profiling/conftest.py
tztsai: Then perhaps adding `gc.collect()` at the end of tearing down can free the space these objects occupy.

davidorme: I've added that - makes no difference to the reported peak usage. I say we park this for now!

tztsai:
LGTM

tztsai (approved):

PR #220, opened 2024-04-22
Rebuild of QA chain using ruff

### LINKED ISSUE #219: Explore use of `ruff`

The `ruff check` and `ruff format` commands:

*  are very fast 
*  replace `pyupgrade`, `isort`, `flake8` and `black` in our `pre-commit` tool chain and dependencies
* remove the need for separate `setup.cfg` and `pyproject.toml` files (`mypy` and `pytest` support `pyproject.toml`)
* `pydocstyle` has archived with pointers to `ruff`.

Adopt `ruff` in the tool chain

Keeping the existing functional toolchain - speed cost in everyday running of toolchain but no switching cost. 

**Additional context**

### LINKED ISSUE #226: Adding compatibility for Python 3.12
Currently, the pyrealm unit tests will not run and flake8 will not run from the command line when running with Python 3.12.0. This is due to a flake8 and pytest-flake8 version incompatibility with the newer version of python. We should also update our test runners and docs accordingly.

Option 1: Remove pytest-flake8 and bump version of flake8 to 7.0.0
One potential option is to upgrade the version of flake8 specified in pyproject.toml (7.0.0 is compatible with the newer python). However, pytest-flake8 does not have a version compatible with Python 3.12.0. We are currently considering removing this from the pytest config so that it no longer runs as part of the unit tests, since flake8 is already run as part of the pre-commit step. This solution has been confirmed to be backwards compatible with Python 3.10 and 3.11 too.

Option 2: Switch to using ruff and remove the use of flake8 and pytest-flake8
We currently have an open issue to investigate the use of ruff: https://github.com/ImperialCollegeLondon/pyrealm/issues/219 
We may decide to replace flake8 entirely with ruff.

Whichever option is chosen, we should also add test runners for Python 3.12 and potentially considering adding explicit upper bound to the compatible python versions listed in our docs.

Pending investigation into ruff, both solutions are a possibility. 

**Additional context**
N/A

davidorme:
# Description

This PR switches the QA chain away from `flake8`, `black`, `isort` and `pyupgrade` to `ruff`: it is significantly faster and looks to be more robust and less version sensitive. This is needed to close #226 - we should be able to support Python 3.12 seamlessly with the new QA chain.

Although `pytest-ruff` exists, I have also removed QA plugins from the `pytest` configuration, so `pytest-mypy` and `pytest-flake8` are now retired and not replaced. The rationale for this is that they significantly slow down testing - which is about making sure the code works - and the QA steps are already enforced via `pre-commit` on local machines and the GA workflows.

Fixes #219 

## Type of change

## Key checklist

## Further checks


j-emberton (approved):

AmyOctoCat (approved):
LGTM

pyrealm/pmodel/subdaily.py
```diff
@@ -267,7 +267,8 @@ def __init__(
             "co2",
             "patm",
             "vpd",
-        ] + optimal_chi_class.requires
+            *optimal_chi_class.requires,
```
j-emberton, line 270: This seems like a strange change. Is it separate to the edits to support ruff?

davidorme: Yeah - it's this rule: https://docs.astral.sh/ruff/rules/collection-literal-concatenation/

I believe them, but it puzzled me too!

j-emberton: yeah, its not super obvious is it!

PR #225, opened 2024-04-24
Fixing profiling and comparing profiling implementations

### LINKED ISSUE #224: Profiling workflow still broken
The profiling GH action is still broken - the benchmarking tests passed but the problem is that I forgot that the develop and main branch are protected and won’t accept a push of the updates to the benchmarking data and plots. The updates have to come in via a PR.

https://github.com/ImperialCollegeLondon/pyrealm/actions/runs/8813621787/job/24192028740

There isn't a great solution here to the broader question of how and when to add this data. @tztsai has pointed out this thread:

https://stackoverflow.com/questions/62846375/perform-github-action-when-trying-to-merge-branch

That summarizes to:
1. Do the profiling on the target branch (as now) - we then have to PR the changed files.
2. Use the `merge_queue` option as a way to catch run the profiling on a merge_queue events, derailing the PR on failed profiling, which feels kinda hacky.
3. Do the profiling on all PRs, so add the updates to the incoming changes, but reduce the number of times it triggers by targeting something like approval events or review requests (see [here for run on approval](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-when-a-pull-request-is-approved))

davidorme:
# Description

The issue #224 discusses a couple of ways to rethink when and how profiling is run. This PR implements _two_ ways, so we can see how they pan out in parallel.

* Updates the profiling workflow that runs on pushes to `develop` (so after a PR is merged) to correct a couple of minor structural issues and add a pull request to update the stats. The current implementation simply tries to push to `develop` but that isn't allowed.

* Adds a parallel implementation that (should) only run when a PR is reviewed and approved. This one then commits the updated files to the PR. I think we'll need to add the new action to the list of tests needed to merge.

So: 

**pyrealm_profiling_after_push.yaml**: An approved PR is merged, the profiling and benchmarking runs and a second PR is created to add the updates (hopefully with a warning on failed tests).

👍 Only runs once.
👎 Requires an extra PR, failing code is now **in `develop`**.   
👎 🤯  Oh god, and I suppose it is going to start of another profiling run when it gets merged, which will... etc etc.

**pyrealm_profiling_on_approve.yaml**: An approved PR is merged, the profiling and benchmarking runs and a second PR is created to add the updates (hopefully with a warning on failed tests).

👍 Stops failing code making it into `develop` without fixing/justification, single PR.
👎 Might run several times.

Fixes #224 (maybe)

## Type of change

## Key checklist

## Further checks


davidorme:
pre-commit.ci run

AmyOctoCat (approved):
LGTM, will be good to see which approach works better.

davidorme:
@AmyOctoCat - the on approval profiling script was broken so it failed its own test when you approved it. Can you reapprove so we can see if that is now fixed.

AmyOctoCat (approved):
Re-approving to trigger the profiling 👍🏻

AmyOctoCat (approved):
Approving to re-trigger profiling

davidorme:
@AmyOctoCat I wondered if something like that was needed, or maybe this: https://stackoverflow.com/a/71367604

AmyOctoCat (approved):
Re-triggering profiling step.

AmyOctoCat (approved):
Re-triggering profiling

AmyOctoCat:
> @AmyOctoCat I wondered if something like that was needed, or maybe this: https://stackoverflow.com/a/71367604

Sorry David, only just noticed your comment, that looks like it should do the job. I'm hoping the last commit I pushed achieves something similar, but if it doesn't work, that'll be next port of call...

davidorme:
It ran!

davidorme:
That worked! I'm not sure what happens with the actions here - I've seen this before and things just seem to hang for some reason.

AmyOctoCat (commented):
Apologies, I'm a bit worried the change I made to the checkout step might not be using the right branch, just looking at the logs to confirm.

AmyOctoCat (commented):
Using github.ref uses refs/remotes/pull/225/merge in this case.

From reading about it, it looks like refs/remotes/pull/<pr-number>/merge, is:
"a reference created by GitHub to keep track of what would happen if a pull request was merged. It references the merge commit between refs/pull/<number>/head and the destination branch (e.g. master).

You can think of it as a "future" commit. GitHub (as well as other Git-based collaboration platforms) use this technique to determine whether the pull request would successfully merge with all the checks passing and no merge conflicts." (from: https://stackoverflow.com/questions/63594658/git-refs-merge-vs-head-in-pull-request). 

I was confused about why it was creating a new commit hash in the logs and talking about merging, but this makes sense of that now. I think this is the desired behaviour?

PR #231, opened 2024-05-09
WIP - 230 create a draft canopy class

### LINKED ISSUE #230: Create a draft Canopy class
This new feature is to take the canopy model description added in the WIP pull request #227 and implement the creation of a Canopy instance:

Approximate specification: 

* The `Canopy` class should take a `Community` object (as in the current sketch) as an `__init__` argument.

* It should populate attributes that record:
    * the vertical heights of the canopy layers ($z^*_l$), which  will be a one dimensional array of heights in metres.
    * the vertical light extinction profile through the canopy layers, as a one dimensional array of proportion of down-welling Tlight absorbed layer $l$
    *  the leaf area within each layer for each cohort: this is a 2 dimensional array (n cohorts x n layers) of areas in metres squared.

It isn't clear to me that this needs anything public _except_ the `__init__` method? It makes sense to break up the calculations into logical subfunctions for testing and readability, but basically it is one big calculation. We could have an explicit `calculate_canopy` step?

AmyOctoCat:
# Description

An initial implementation for a canopy class that calculates the canopy layer heights for a given community. See `community_sketch.py` for an example of the code in use.

Still to do:
- light extinction profile
- leaf area within each canopy layer -> in progress. This currently uses a map implementation, but I think it would be better if this could be switched to something using 2D arrays, as under the hood the map is a python for loop so not as performant.
- would like to change the names of anything containing `mass_swd` for better readability.
- Change the import format of plant functional types to a toml file.
- Finish off testing, status as follows:
1. test_jaideeps_t_model_extension_function.py - happy paths started, need to finish happy paths and add unhappy paths.
2. test_t_model_functions.py - happy paths written, need to add unhappy paths
3. test_canopy.py - skeleton functions added for happy paths, need filling in and unhappy paths adding
4. test_community.py - skeleton functions added for happy paths, need filling in and unhappy paths adding
5. test_canopy_functions - skeleton functions added for happy paths, need filling in and unhappy paths adding

- Improve documentation.

A rough outline of how the canopy layers are calculated:
1. Data is imported, this is currently in the form of a json array of PlantFunctionalType objects and csv called `communities.csv` containing rows of cohorts. This uses the `.from_csv()` method in the communities object.  see `community_sketch.py` for an example use.
7. When the Community is initialised, the T Model Geometry and Canopy Factors from Jaideep's extension to the T Model are calculated in the constructor and saved as array properties of the Community class.
8. A Canopy class can then be created from a Community object and a canopy gap fraction.
9. This uses a calculate_layer_heights function, which is called during initialisation of the Canopy class, and provides the canopy layer heights as an array property of the Canopy class.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


davidorme (changes requested):
I think this is good but there are some structural things it would be good to look at. A couple of overarching things:

1. As noted below, I think I've given you a bad steer on the Community structure. I'm not convinced that actual `Cohort` objects are the way forward - the structure is very clear but it's likely to be less efficient. It is possible though that adding and deleting cohorts is much more efficient with lists of `Cohort` instances than subsetting and appending arrays?

2. I think I'd rather have factory methods of classes than the intermediate import classes - it just seems cleaner.

pyrealm/canopy_model/model/community_alt.py
```diff
@@ -0,0 +1,105 @@
...
+
+class CommunityTraits:
+    def __init__(self, cohort_pfts: NDArray[np.str], flora: Flora):
+        self.qm = self.get_trait_array("qm", cohort_pfts, flora)
+        ...
+
+    def get_trait_array(
+        self, trait_name: str, cohort_pfts: NDArray[np.str], flora: Flora
+    ) -> NDArray:
+        """Extract array of values across cohorts for a given trait.
```
davidorme, line 66: Actually - this is a static method isn't it. It could be a standalone function (easier to use in other use cases), or we could make `cohort_pfts` and `flora` into attributes and simplify the signature.

pyrealm/canopy_model/model/community_alt.py
```diff
@@ -0,0 +1,105 @@
...
+        self.cohort_dbh = cohort_dbh
+        self.cohort_individuals = cohort_individuals
+        self.cohort_pfts = cohort_pfts
+        self.flora = flora
+
+        # Arrays of required PFT traits for cohorts
+        self.traits = CommunityTraits(self.cohort_pfts, self.flora)
+
+        # Calculate geometries from initial dbh
+        self.tmodel_geometries = TModelGeometries(self.cohort_dbh, self.traits)
```
davidorme, line 33: We _could_ just have trait arrays and cohort geometries attributes just as part of Community, but it would be a long list and it seems cleaner and more easily tested if we encapsulate them.

pyrealm/canopy_model/model/cohort.py
```diff
@@ -0,0 +1,44 @@
...
+)
+
+
+class Cohort:
+    """Contains cohort data.
+
+    Stores cohort data. During initialisation, retrieves the detailed plant functional
+    type data, calculates the T Model Geometry of a given cohort and the canopy factors
+    from Jaideep's extension to the T Model.
+    """
```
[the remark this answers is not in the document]
davidorme: Just reviewing the PRs and we did discuss this on the meta-issue for demography (https://github.com/ImperialCollegeLondon/pyrealm/pull/227#issuecomment-2090530654), without really coming to any sort of resolution.

pyrealm/canopy_model/model/community_alt.py
AmyOctoCat: Thanks David, it's really good to see an illustration of what you meant. I'm a bit concerned that when adding and removing cohorts we'd have to be very careful to ensure that we removed the relevant entries from every set of traits, ie all the t model geometries, all the community traits, etc., since if we missed one, we'd end up with the data misaligned, which might make it a bit fragile. If we later added new attributes to a cohort, then we'd also need to be sure we were adding them into whatever function we use to remove/add them. The best case would be the arrays being different lengths and ending up with a runtime failure, but I'd also be worried about the program silently just attaching the wrong geometry to the wrong cohort etc (for example if we removed one cohort from one data set and a different one from another).

AmyOctoCat:
> I think this is good but there are some structural things it would be good to look at. A couple of overarching things:
> 
>     1. As noted below, I think I've given you a bad steer on the Community structure. I'm not convinced that actual `Cohort` objects are the way forward - the structure is very clear but it's likely to be less efficient. It is possible though that adding and deleting cohorts is much more efficient with lists of `Cohort` instances than subsetting and appending arrays?
> 
>     2. I think I'd rather have factory methods of classes than the intermediate import classes - it just seems cleaner.

Hi David,

Thank you very much for the feedback, these were things I was going round in circles a bit myself, so it's great to have the input. 

Regarding the first point I think I've circled round on the arrays vs objects quite a few times over and still don't have a strong feeling. I think going for the OO approach might make it easier to read and easier to reason about when it comes to adding and removing cohorts, but I can see the efficiency argument for using arrays.

Regarding the second point I also went in and out on this. JSON dataclasses did not play very nicely with passing the flora object into the nested cohorts within the community which was why I ended up with the separate import classes (I thought this was also a bit clearer from the point of view of easily seeing which fields were needed in the input data). Once there were separate import classes, the logic looked quite complicated and I thought it deserved it's own class (was going for the single responsibility principle from SOLID), but will have a review to see if there's a better way. Maybe adding extra functionality to the generated from_json method that json_dataclass creates in the cohort class is possible that would allow us to pass the flora in there.

Sorry that's all very inconclusive! Hopefully some more ideas might percolate over the space of the weekend.

pyrealm/canopy_model/model/community_alt.py
davidorme: That is a clear advantage of the OO `Cohort` model - and also of using some kind of record array. Having said that, we could also write a fairly defensive `remove_cohort(index)` method to these classes.

pyrealm/canopy_model/model/canopy.py
```diff
@@ -0,0 +1,157 @@
...
+        :param fG: community gap fraction
+        :param A: community area
+        :param l: layer index
+        :param z: height
+        """
+
+        community_projected_area_at_z = self.calculate_community_projected_area_at_z(z)
+
+        # Return the difference between the projected area and the available space
+        return community_projected_area_at_z - (A * l) * (1 - fG)
```
vgro, line 69:
```suggestion
        return (
            community_projected_area_at_z
            - (community_area * layer_index) * (1 - gap_fraction)
        )
```

davidorme:
I've reviewed the files and all of @AmyOctoCat's work in here has been merged in through the incremental PRs linked in above.

PR #237, opened 2024-05-22
236 develop pyrealmcoresolar function radiation calc library

### LINKED ISSUE #236: Develop pyrealm.core.solar function radiation calc library
Proposal to breakout individual solar radiation calc functions from `pyrealm.splash.solar` and make these available in `pyrealm.core.solar`

**Alternatives**
Alternative would be to try and find an existing Python library to provide similar functionality and then write custom functions for any missing features such as calculation of ppfd and calculation of radiation for sub-daily intervals..
Seeing as we already have much of the code we need in `Pyrealm`, we might as well reuse it.

**Additional context**
This is in support of implementing the two-leaf canopy model

j-emberton:
# Description

Refactored `pyreal.splash.solar` to move core calculations to `pyrealm.core.solar`.
Functionality of Splash module is unchanged
Change is to facilitate development of two-leaf canopy model

Fixes #236 

## Type of change

## Key checklist

## Further checks


davidorme:
@j-emberton That all looks good. In the past, I wouldn't have tended to use `pytest.parameterise` when there is only one parameter set, but this makes it easier to extend and better isolates/identifies the inputs, so is probably better practice. 

One awkward thing in the API are the functions `calc_sunset_hour_angle` and `calc_daily_solar_radiation` taking those intermediate variables `ru` and `rv` rather than directly accepting `delta` and `lat`. There's good reason for that in `splash` usage - `ru` and `rv` are used repeatedly in different calculations so calculating them once is more efficient, but if someone is using them more directly, then it is a bit obscure. Having said that, the API does get a bit clumsy if you either provide `delta` and `lat` (and create `ru` and `rv` internally) or provide `ru` and `rv`...

davidorme:
Oh - wait... Will post a suggestion.

pyrealm/core/solar.py
```diff
@@ -23,6 +23,137 @@
...
+
+    Args:
+        ru: dimensionless parameter
+        rv: dimensionless parameter
+        k_pir: constant rad to degrees conversion, degrees/rad
+    """
+
+    angle = np.arccos(-1.0 * np.clip(ru / rv, -1.0, 1.0)) / k_pir
+
+    return angle
```
davidorme, line 103:
```suggestion
def calc_sunset_hour_angle(delta: NDArray, lat: NDArray, k_pir: float) -> NDArray:
    """Calculate sunset hour angle.

    This function calculates the sunset hour angle using Eq3.22, Stine & Geyer (2001)

    Args:
        ru: dimensionless parameter
        rv: dimensionless parameter
        k_pir: constant rad to degrees conversion, degrees/rad
    """
    ru, rv = calc_lat_delta_intermediates(delta, lat)
    
    return _calc_sunset_hour_angle_from_ru_rv(ru, rv, k_pir)
    
def _calc_sunset_hour_angle_from_ru_rv(ru: NDArray, rv: NDArray, k_pir: float) -> NDArray:
    """Calculate sunset hour angle from intermediates.
    
    This function calculates the sunset hour angle using Eq3.22, Stine & Geyer (2001)
    Args:
        ru: dimensionless parameter
        rv: dimensionless parameter
        k_pir: constant rad to degrees conversion, degrees/rad
    """

    return np.arccos(-1.0 * np.clip(ru / rv, -1.0, 1.0)) / k_pir
```

j-emberton:
@davidorme this library seems to be working now and (at least locally) all tests pass.

I incorporated your proposed change to calculate `ru` and `rv` from within `calc_hour_angle`. However, if we apply this throughout `splash.solar`, then `ru` and `rv` get calculated multiple times instead of just once adding computational cost.

If we compile these functions with cython or number it probably doesn't matter too much unless the arrays are huge.

What would you prefer?

davidorme:
I guess I was thinking that we extend that same structure to the other functions that use `ru` and `rv`. So individual users of the stand alone functions might provide `delta` and `latitude`, but all of those functions are backed by a short cut method that accepts `ru` and `rv`. 

I'm not sure how complex that looks - I think my approach here is that users see functions with straightforward inputs and don't have to manual calculate those intermediate inputs, but those functions can then wrap the calculation of the intermediate values and an internal calculation, which we can use more directly inside routines.

j-emberton:
One way around it would be to put these library functions inside a class and then have self.ru and self.rv generated upon creating a class instance.

public methods would then have access to these variables by default.

What I'll do for now is go ahead and implement as per your suggestion. We can then profile the code and see if this is an issue or not. If it is, we have a good place to refactor for performance improvements as we have tests ready to go.

j-emberton:
One point I haven't put lots of thought into is whether we want more extensive documentation updates now we've added more functionality to the `core.solar` module

vgro (approved):
LGTM

j-emberton:
@davidorme 
Screenshot 2024-07-18 at 10 02 17

If I add some user guidance on how to access the functions in this library, where should I put it. Do we want it to go under core utilities, or create a new solar utilities section?

j-emberton:
moving back to draft while I add a few more bits

davidorme:
> If I add some user guidance on how to access the functions in this library, where should I put it. Do we want it to go under core utilities, or create a new solar utilities section?

Definitely a new `core.solar` section.

j-emberton:
On top of what was in there before I've added:
new functions:
- calc_ppfd - wrapper to calculate pfd from elementary inputs
- calc_declination - function to calculate solar declination
- calc_beta_angle_from_lat_dec_hour - function to calculate solar elevation

Tests for the above functions

Added detail like examples and latex maths to some of the docstrings to function as public facing documentation (where needed)

j-emberton:
Once we get a couple more papers into the bibliography we can add proper citations for the implemented maths

pyrealm/core/solar.py
```diff
@@ -1,26 +1,584 @@
...
+
+    This function calculates distance factor using the method of Berger et al. (1993)
+
+    .. math::
+
+        dr = \left( \frac{1.0}{\left(\frac{1.0 - k_e^2}{1.0 + k_e \cos\left(\deg2rad(\nu)\right)}\right)} \right)^2
+
+    Args:
+        nu          : heliocentric true anomaly (degrees)
+        k_e         : Solar eccentricity
```
davidorme, line 24:
```suggestion
    This function calculates the distance factor :math:`dr` using the method of
    :cite:t:`berger:1993a`.

    .. math::

        dr = \left( 1.0 \mathbin{/}
               \left(\frac{1.0 - k_e^2}
                          {1.0 + k_e \cos\left(\nu \cdot \pi \mathbin{/} 180)\right)}
               \right)
             \right)^2

    Args:
        nu          : heliocentric true anomaly (:math:`\nu`, degrees)
        k_e         : Solar eccentricity (:math:`k_e`)
```

This is exceptionally picky:
* I think the latex is easier to read with some line breaks but it also means we can drop the `noqa`. 
* I've also avoided the triple stack division - the symbols get tiny when it is rendered
* I've replaced `\deg2rad` with the actual conversion.
* I've annotated the text to identify the symbols in the equation.

pyrealm/core/solar.py
```diff
@@ -1,26 +1,584 @@
...
+    julian_day: NDArray,
+    n_days: NDArray,
+    k_e: float,
+    k_eps: float,
+    k_pir: float,
+    k_d: float,
+    k_Gsc: float,
+    k_fFec: float,
+    k_alb_vis: float,
+    k_c: float,
```
davidorme, line 262: I'd be tempted here to use `const: CoreConst` just to make it easier to use and then pass the individual constants down to the sub-functions.

pyrealm/core/solar.py
```diff
@@ -1,26 +1,584 @@
...
+def calc_daytime_net_radiation(
+    hn: NDArray,
+    k_pir: float,
+    rnl: NDArray,
+    delta: NDArray,
+    latitude: NDArray,
+    k_alb_sw: float,
+    tau: NDArray,
+    k_Gsc: float,
+    dr: NDArray,
```
davidorme, line 448: Maybe order these to put the constants at the end? This might be another helper function that could take `const`?

pyrealm/splash/solar.py
```diff
@@ -113,78 +118,76 @@ def __post_init__(
...
-                self.ru * self.core_const.k_pir * self.hs
-                + self.rv * np.sin(np.deg2rad(self.hs))
-            )
+        self.ra_d = calc_daily_solar_radiation(
+            self.core_const.k_Gsc,
+            self.dr,
+            self.core_const.k_pir,
+            self.hs,
+            self.delta,
+            lat,
```
davidorme, line 134: I've a strong preference (that is probably wildly inconsistently applied) for naming arguments to function calls and not relying on argument order.

davidorme (dismissed):
This looks great. Only minor things:

* Some docstring formatting issues. I've pushed a commit to include the solar docs in the API section, so we can review how they render. I've made some general suggestions on one of the docstrings that apply across the module.
* Using named arguments - I'm not brilliant at practicing this, but with long lists of args, it's much more robust.
* I think the helper wrapper functions could take a packaged `CoreConst` object to make them more usable.

j-emberton:
Thankyou for the feedback @davidorme. Much appreciated. I thought we had agreed to make all the variables explicit but I may have got my wires crossed. In any case I'm happy to swap in the core const and adjust as required. I'll do it for all the library functions I create for consistency. It makes sense since those objects will usually be available a priori and don't require any bespoke setup.

pyrealm/splash/solar.py
j-emberton: I'll update these functions for sure. Would you like this applied as a general rule?

davidorme:
I don't think you got your wires crossed - it's a grey area and we don't have a consistent approach.  It's difficult to come up with a hard and fast rule and we're not systematic about this, but I think it's something like "If a function is passing constants on to lower level functions, then passing in constants classes is probably easier for the user and more compact."

pyrealm/splash/solar.py
davidorme: Yeah, I think so.

pyrealm/core/solar.py
```diff
@@ -1,26 +1,584 @@
...
+) -> NDArray:
+    r"""Calculates nightime net radiation, J/m^2.
+
+    .. math::
+
+        secs\_d = 86400  \quad \text{ \# seconds in one solar day}
```
j-emberton, line 521: @davidorme should we move seconds in one solar day into core constants?

davidorme: Happy to! Although it seems obvious, pinning it to a clear definition is better.

pyrealm/core/solar.py
```diff
@@ -670,14 +670,9 @@ def calc_solar_elevation(site_obs_data: LocationDateTime) -> NDArray:
...
-                            time-specific information for the observation site,
-                            including
-            latitude_rad: site latitude in radians
-            longitude   : site longitude in decimal degrees
-            UTC offset  : Offset from UTC for site local meridian, decimal degrees
-            local time  : decimal local time of observations at site,
-            Julian days : Julian day of observations at site
+        site_obs_data: A :class:`~pyrealm.core.calendar.LocationDateTime` instance
+            containing the location and time-specific information for the observation
+            site.
```
davidorme, line 675: This avoids repetition of the class structure by pointing to the specific docs (but also, the resulting rendering of this RST structure was all kinds of nasty).

pyrealm/core/solar.py
```diff
@@ -87,3 +657,252 @@ def calc_heliocentric_longitudes(
...
+        >>> latitude = -35.058333
+        >>> longitude = 147.34167
+        >>> year_date_time = np.array([np.datetime64("1995-10-25T10:30")])
+        >>> UTC_offset = 150.0
+        >>> ldt = LocationDateTime(latitude = latitude, longitude = longitude, year_date_time = year_date_time, UTC_offset = UTC_offset)
+        >>> # Run solar elevation calculation
+        >>> calc_solar_elevation(ldt)
+        array([1.0615713])
+
+    """  # noqa: E501
```
MarionBWeinzierl, line 697: Did you want to remove this and wrap line 692 around?

tests/unit/core/test_solar.py
```diff
@@ -24,3 +344,128 @@ def test_calc_heliocentric_longitudes(day, n_day, expected):
...
+
+
+@pytest.mark.parametrize(
+    argnames="latitude, declination, hour_angle, expected",
+    argvalues=[
+        (np.array([-0.61]), np.array([-0.23]), np.array([-0.37]), np.array([1.0647289]))
+    ],
+)
+def test_elevation_from_lat_dec_hn(latitude, declination, hour_angle, expected):
+    """TBD."""
```
MarionBWeinzierl, line 465: What is missing here?

tests/unit/core/test_solar.py
```diff
@@ -24,3 +344,128 @@ def test_calc_heliocentric_longitudes(day, n_day, expected):
...
+
+    assert np.allclose(result, expected)
+
+
+@pytest.mark.parametrize(
+    argnames="julian_day, expected",
+    argvalues=[(np.array([298]), 5.11261928)],
+)
+def test_calc_day_angle(julian_day, expected):
+    """TBD."""
```
MarionBWeinzierl, line 407: What is missing here?

MarionBWeinzierl (commented):
I went through the code and have some minor comments.

j-emberton:
This should be ready to go now

pyrealm/core/calendar.py
```diff
@@ -118,3 +118,78 @@ def __repr__(self) -> str:
...
+    """The offset from Coordinated Universal Time (UTC) for the location."""
+    year_date_time: np.ndarray
+    """An array of np.datetime64 values corresponding to observations at the 
+    location."""
+    julian_days: np.ndarray = field(init=False)
+    """An array of Julian day of the year numbers calculated from the
+    ``year_date_time``."""
+    local_time: np.ndarray = field(init=False)
+    """An array of Julian day of the year numbers calculated from the 
+    ``year_date_time``."""
```
davidorme, line 164: The docstring is wrong here 😄 

Also, it might be worth changing `LocationDateTime.local_time`  to `LocationDateTime.decimal_time` or `LocationDateTime.local_decimal_time` to make it clear that its a different representation.

pyrealm/core/calendar.py
```diff
@@ -118,3 +118,78 @@ def __repr__(self) -> str:
...
+    radians, the Julian days from the date-time information, and a decimal
+    representation of the local time.
+
+    Example:
+        >>> import numpy as np
+        >>> ldt = LocationDateTime(
+        ...     latitude=-35.058333,
+        ...     longitude=147.34167,
+        ...     UTC_offset=10,
+        ...     year_date_time=np.array([np.datetime64("2024-08-12T10:30")])
```
davidorme, line 138: It isn't clear from this example and docs that the `year_date_time` is assumed to be local time. Is UTC optional?

pyrealm/core/solar.py
```diff
@@ -1,26 +1,600 @@
...
+
+    Args:
+        dr          : dimensionless distance factor
+        ru          : dimensionless variable substitute
+        rv          : dimensionless variable substitute
+        hs          : local hour angle, degrees
+        const       : CoreConst object containing core constants, including:
+            k_pir   : radians to degrees conversion, degrees/rad
+            k_secs_d: seconds in one solar day
+            k_Gsc   : planetary radiation constant, W/m^2
```
davidorme, line 189: Odd indenting on these rows.

I guess as a style, I've not tended to use spaces to align the argument descriptions.

pyrealm/core/solar.py
```diff
@@ -1,26 +1,600 @@
...
+
+    ppfd = calc_ppfd_from_tau_ra_d(
+        tau=tau, ra_d=ra_d, k_fFEC=const.k_fFEC, k_alb_vis=const.k_alb_vis
+    )
+
+    return ppfd
+
+
+def calc_rnl(sf: NDArray, tc: NDArray, k_b: float, k_A: float) -> NDArray:
+    r"""Calculates net longwave radiation, rnl, W/m^2.
```
davidorme, line 334: I've a preference for more expressive function names:  `calc_net_longwave_radiation`

pyrealm/core/solar.py
```diff
@@ -1,26 +1,600 @@
...
+    return dr
+
+
+def calc_declination_angle_delta(
+    lambda_: NDArray, k_eps: float, k_pir: float
+) -> NDArray:
+    r"""Calculates declination angle delta.
+
+    This function calculates the solar declination angle delta using
+    the method of :cite:t:'Woolf:1968'.
```
davidorme, line 47:
```suggestion
    This function calculates the solar declination angle delta using the method of 
    :cite:t:`Woolf:1968`.
```

Two things there, one of which is hugely petty 😈 :
* The citation syntax needs backticks (here and elsewhere).
* The docstrings could use more of the linewidth - the project uses 88 characters. I use the Rewrap plugin for VSCode, which is great for wrapping and re-wrapping strings and comments: https://marketplace.visualstudio.com/items?itemName=stkb.rewrap

pyrealm/core/solar.py
```diff
@@ -87,3 +661,254 @@ def calc_heliocentric_longitudes(
...
 
     return (nu, lambda_)
+
+
+def calc_solar_elevation(site_obs_data: LocationDateTime) -> NDArray:
+    r"""Calculate the solar elevation angle for a specific location and times.
+
+    This function calculates the solar elevation angle, which is the angle between
+    the sun and the observer's local horizon, using the methods outlined in
+    de Pury & Farquhar (1997).
```
davidorme, line 671: Convert to `:cite:`

pyrealm/core/solar.py
```diff
@@ -1,26 +1,600 @@
...
+) -> NDArray:
+    r"""Calculate photosynthetic photon flux density, PPFD,(mol/m^2).
+
+    This function calculates the PPFD in mol/m^2 from secondary calculated
+    variables and constants.
+
+    .. math::
+
+        ppfd = (1.0 \times 10^{-6}) \cdot k\_fFEC \cdot (1.0 - k\_alb\_vis) \cdot \tau \
+            cdot ra\_d
```
davidorme, line 244:
```suggestion
        ppfd = (1.0 \times 10^{-6}) \cdot k\_fFEC \cdot (1.0 - k\_alb\_vis) \cdot \tau \
            \cdot ra\_d
```

pyrealm/core/solar.py
```diff
@@ -87,3 +661,254 @@ def calc_heliocentric_longitudes(
...
+    the latter.
+
+    .. math::
+
+    E_t = \left( 0.000075
+    + 0.001868 \cdot \cos(\Gamma)
+    - 0.032077 \cdot \sin(\Gamma)
+    - 0.014615 \cdot \cos(2\Gamma)
+    - 0.04089 \cdot \sin(2\Gamma) \right)
+    \times 229.18
```
davidorme, line 869: Needs to be indented to render as maths

```suggestion
    .. math::

        E_t = \left( 0.000075
            + 0.001868 \cdot \cos(\Gamma)
            - 0.032077 \cdot \sin(\Gamma)
            - 0.014615 \cdot \cos(2\Gamma)
            - 0.04089 \cdot \sin(2\Gamma) \right)
            \times 229.18
```

davidorme (changes requested):
This all looks good - there are a few minor (and below minor...) fixes to make.

I think a wider issue is the notation and rendering of the maths in the docstring equations. It would be good to have:

* proper notation throughout (e.g $k_{pir}$ rather than $k\\_pir$)
* definitions of all notation values in docstrings - that is going to involve either a load of repetitive stuff added to docstrings or possibly a glossary of symbols and units in the module docstring?

Not sure if doing that belongs in this PR or in a new one?

j-emberton:
Thanks for the feedback @davidorme . I'm still a noob with both latex and sphinx, but I'll fix these asap!

j-emberton:
> In many ways, I'd like to avoid both 😄

amen

pyrealm/core/solar.py
j-emberton: Have sorted this now

pyrealm/core/solar.py
j-emberton: sorted this as well

pyrealm/core/solar.py
j-emberton: Sorted

pyrealm/core/solar.py
j-emberton: sorted

pyrealm/core/calendar.py
j-emberton: This is an oversight on my part. I've removed the references to UTC offset. They were a hangover from when I thought calculation of solar elevation required knowledge of local time zone.

pyrealm/core/calendar.py
j-emberton: Done

j-emberton:
> This all looks good - there are a few minor (and below minor...) fixes to make.
> 
> I think a wider issue is the notation and rendering of the maths in the docstring equations. It would be good to have:
> 
> * proper notation throughout (e.g 
>     
>       k
>       
>         p
>         i
>         r
>       
>     
>    rather than 
>     k
>     _
>     p
>     i
>     r
>   )
> * definitions of all notation values in docstrings - that is going to involve either a load of repetitive stuff added to docstrings or possibly a glossary of symbols and units in the module docstring?
> 
> Not sure if doing that belongs in this PR or in a new one?

what do you mean by definition of all notation values here?

davidorme:
> what do you mean by definition of all notation values here?

So having something like "where $k_{pir}$ is the whatever the hell k_pir is" below the equations. I think there is so much overlap that pointing to a glossary in the module docstring would work just fine and be less repetitive.

j-emberton:
OK, I think this is all working now and docs basically look ok, although I appreciate there maybe some style tweaks required - lets do these in a followup issue if that's ok. I'd like to just get this work in the bank.

pyrealm/core/solar.py
```diff
@@ -1,26 +1,604 @@
...
+
+
+def calc_lat_delta_intermediates(
+    delta: NDArray, latitude: NDArray
+) -> tuple[NDArray, NDArray]:
+    r"""Calculates intermediate values for use in solar radiation calcs.
+
+    This function calculates :math:`ru` and :math:`rv` which are dimensionless \
+    intermediate values calculated from the solar declination angle delta and the \
+    observation latitude.
```
davidorme, line 76: This module seems to be using `\` to shown line breaks in docstrings, as in this example selection.  That isn't / shouldn't be needed.

davidorme (approved):
I agree that we should bank this functionality and worry about documentation polishing later. The docs are readable and useable now.

One oddity noted in a comment - there are a lot of line endings marked with `\`. Those don't appear in the built docs, so aren't a problem, but also aren't required and aren't used elsewhere in the project.

pyrealm/core/solar.py
j-emberton: I think I started doing this after you mentioned something a while back about doing line character limits properly. And I think I read somewhere else that it was necessary/useful. I may also just be confused! Happy to remove them just now!

omarjamil (approved):

Issue #244, opened 2024-05-29
Two leaf irradiance


j-emberton:
**Is your feature request related to a problem? Please describe.**
To implement a class capturing the two leaf irradience model and data

**Describe the solution you'd like**
Single class with multiple methods calculating various parameters but also drawing on core.solar library of useful calculations

**Describe alternatives you've considered**

**Additional context**

PR #245, opened 2024-06-04
Adding two leaf irradience model

### LINKED ISSUE #236: Develop pyrealm.core.solar function radiation calc library

j-emberton:
# Description

Adding irradience calculation class, supporting functions and constants dataclass for the two leaf canopy model implementation.

Fixes #236 and #244 

## Type of change

## Key checklist

## Further checks


pyrealm/pmodel/two_leaf_irradience.py
```diff
@@ -0,0 +1,31 @@
...
+class TwoLeafIrradience:
+    """Calculates two-leaf irradience."""
+
+    def __init__(self, beta_angle: NDArray, PPFD: NDArray, LAI: NDArray, PATM: NDArray):
+
+        self.beta_angle: NDArray = beta_angle
+
+
+def beta_angle(lat: NDArray, d: NDArray, h: NDArray) -> NDArray:
+    """Calculates solar beta angle.
```
davidorme, line 16: The reference for this is page 554 of this:
https://onlinelibrary.wiley.com/doi/epdf/10.1111/j.1365-3040.1997.00094.x

I think `beta_angle` here is (or should be) solar elevation angle - which might make it part of the solar functionality rather than an irradiance specific thing. If we move to using one of the existing solar packages, we might get a more accurate model.

pyrealm/pmodel/two_leaf_irradience.py
```diff
@@ -0,0 +1,31 @@
...
+
+    Calculates solar beta angle using Eq A13 of dePury & Farquhar (1997).
+
+    Args:
+        lat: array of latitudes (rads)
+        d: array of declinations (rads)
+        h: array of hour angle (rads)
```
davidorme, line 23: I've adopted Keith's short variable names in my rough and ready translation to Python, but we should shift to longer names (`latitude`, `declination`, `hour_angle`)?

pyrealm/pmodel/two_leaf_irradience.py
```diff
@@ -0,0 +1,31 @@
+"""To do."""
+
+import numpy as np
+from numpy.typing import NDArray
+
+
+class TwoLeafIrradience:
+    """Calculates two-leaf irradience."""
+
+    def __init__(self, beta_angle: NDArray, PPFD: NDArray, LAI: NDArray, PATM: NDArray):
```
davidorme, line 10: Although these variable names are acronyms, I think we should stick to lower case variable names: `ppfd`, `lai`, `patm` (and arguably longer - `leaf_area_index`, although I know I'm guilty of using `patm` and `tc` elsewhere. 

We could do with a sweep to standardise /check variable names, really!

davidorme (approved):
This looks great - structure is very compliant with the existing package and popping out the constants class is good. I've added some comments but there isn't any refactoring or major stuff to worry about.

vgro (approved):
Looks very good to me, I agree with David's comments. One minor point, I would suggest to start docstrings with capital letters for consistency (and potentially to avoid some warning messages).

j-emberton:
@davidorme Could you help me with a query.

I have been implementing the irradience model into a Pyrealm irradience class, and along the way separating out the individual functions. It appears that a variable defined in the .md is surplus to requirements as it is never used. I have checked dePury& Farquhar(1997) and this appears to be the case. The extraneous variable is `Ibs` calculated at line 155 in the .md. Could you confirm?

pyrealm/pmodel/two_leaf_irradience.py
j-emberton: I've added pysolar alternatives in the latest commit which calculate solar elevation directly from latitude, longitude and date time. However, they don't take array inputs natively so you need to use and enumerate operation, so there will be a speed accuracy trade.

The best pysolar solar elevation implementation includes atmospheric distortion, but they also have a faster/simpler version which is the same as the first version I wrote. Because the first version uses numpy vectorisation it may well be faster but I've included the pysolar fast version anyway.

pyrealm/pmodel/two_leaf_irradience.py
```diff
@@ -299,3 +308,24 @@ def beta_angle(latitude: NDArray, declination: NDArray, hour_angle: NDArray) ->
...
+    """
+    solar_elevations = np.zeros_like(latitudes, dtype=float)
+
+    if not fast:
+        get_beta = get_altitude
+    else:
+        get_beta = get_altitude_fast
+
+    for i, (lat, lon, date_time) in enumerate(zip(latitudes, longitudes, date_time)):
+        solar_elevations[i] = get_beta(lat, lon, date_time)
```
davidorme, line 329: I don't think we can do this - partly the speed but also it doesn't generalise to multiple dimensions

```python
In [4]: import numpy as np
In [5]: x = np.arange(1,10).reshape((3,3))
In [6]: for ix, (x1, x2) in enumerate(zip(x, x)):
   ...:     print(ix, x1, x2)
   ...: 
0 [1 2 3] [1 2 3]
1 [4 5 6] [4 5 6]
2 [7 8 9] [7 8 9]
```
If I read the `pysolar` code correctly, I think the `get_altitude_fast` is the same (or similar) to the previous beta calculation. I'd love to outsource these calculations to a more specialised and precise package but maybe not now?

https://github.com/pingswept/pysolar/blob/ed75beb5ccfe9eeb2d4a6a410f5f953214e8d19f/pysolar/solar.py#L134

j-emberton: Ah, OK. I think I made the assumption that the array would be 1D.

I just took a look at the pysolar functionality as I thought you were prompting that earlier in the discussion! No matter though.

davidorme: It would be great to use something like this -  especially if it opens up better models of atmosphere etc - but I think the ability to handle arrays is a prerequisite unfortunately. With a couple of exceptions, the code in `pyrealm` is written to handle arbitrary dimensionality, although the most usual use cases are: 1D (time series, typically), 2D (XY, or maybe a set of sites through time), 3D (XYT).

davidorme: The "exceptions" are really just - first axis must be Time because there is a lagged process involved.

pyrealm_build_data/two_leaf/Untitled.ipynb
MarionBWeinzierl, line 1: Should this get a title, or be deleted?

tests/regression/two_leaf_canopy/test_two_leaf_model.py
```diff
@@ -0,0 +1,128 @@
+"""TBC."""
```
MarionBWeinzierl, line 1: Needs a proper docstring

tests/regression/two_leaf_canopy/test_two_leaf_model.py
```diff
@@ -0,0 +1,128 @@
...
+import pytest
+
+from pyrealm.pmodel import PModel
+from pyrealm.pmodel.pmodel_environment import PModelEnvironment
+from pyrealm.pmodel.two_leaf_irradience import TwoLeafAssimilation, TwoLeafIrradience
+
+
+@pytest.fixture
+def get_data():
+    """Docstring."""
```
MarionBWeinzierl, line 14: Also needs filling in, as do those below

tests/regression/two_leaf_canopy/test_two_leaf_model.py
```diff
@@ -0,0 +1,128 @@
...
+@pytest.fixture
+def solar_elevation():
+    """Docstring."""
+    # from Keiths code 0.982121, 0.9820922045388
+    return np.array([0.98212117])
+
+
+@pytest.fixture
+def solar_irradience(solar_elevation, get_data):
+    """Test."""
```
MarionBWeinzierl, line 30: And these need better descriptions

tests/unit/two_leaf_canopy/test_two_leaf_classes.py
```diff
@@ -0,0 +1,181 @@
+"""TBC.
```
MarionBWeinzierl, line 1: Same here, need proper docstrings

MarionBWeinzierl (commented):
Just flagging a couple of dummy docstrings you might have missed.

j-emberton:
@MarionBWeinzierl Thanks for taking a look. Yes, filling in the last few bits but nearly there :)

PR #263, opened 2024-07-17
Allow P Models to accept an array of kphio values

### LINKED ISSUE #261: Allow PModel to accept an array of kphio values.

At present, the `kphio` argument to `pyrealm.pmodel.pmodel.PModel`  is expected to be a float value. If this is supplied, it is converted into a `numpy` scalar array, with the value then being used for all calculations. However, users might want to be able to supply per-observation estimates of `kphio` rather than a single global initial value.

Update `kphio` to accept scalar or array inputs and then check that array `kphio` values broadcast correctly across other array inputs.

davidorme:
# Description

This PR update the `kphio` arguments to `PModel` and `SubdailyPModel` to accept an array of `kphio` estimates. This is largely intended to allow users to test alternative models for calculating `kphio` - there is an approach that accounts for temperature and soil moisture being developed.

The PR then includes a new test suite for this feature that checks that outcomes of passing in an array of kphio values is identical to the predictions of calculating those values using a single value at a time, and that the results are not sensitive to the shape of the inputs. That is repeated for `PModel` and `SubdailyPModel`.

Fixes #261

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/pmodel/pmodel.py
+++ b/pyrealm/pmodel/pmodel.py
@@ -217,13 +217,13 @@ class PModel:
         # Set context specific defaults for kphio to match Stocker paper
         if kphio is None:
             if not self.do_ftemp_kphio:
-                self.init_kphio = np.array([0.049977])
+                self.init_kphio = np.array(0.049977)
             else:
-                self.init_kphio = np.array([0.081785])
+                self.init_kphio = np.array(0.081785)
         else:
             if isinstance(kphio, float):
                 # A single scalar global value
-                self.init_kphio = np.array([kphio])
+                self.init_kphio = np.array(kphio)
             elif isinstance(kphio, np.ndarray):
                 # An array of values, which must match the shape of the inputs to the
                 # PModelEnvironment instance.
--- a/pyrealm/pmodel/pmodel.py
+++ b/pyrealm/pmodel/pmodel.py
@@ -186,7 +186,7 @@ class PModel:
     def __init__(
         self,
         env: PModelEnvironment,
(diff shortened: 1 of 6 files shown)


j-emberton (commented):
Do we need to add a bit more documentation to guide users on how they can use this new feature? Or is it in line with what we currently have?

davidorme:
> Do we need to add a bit more documentation to guide users on how they can use this new feature? Or is it in line with what we currently have?

Great point - I've added a short paragraph to the user facing PModel docs.

j-emberton (approved):
Good addition

pyrealm/pmodel/pmodel.py
```diff
@@ -215,11 +217,18 @@ def __init__(
         # Set context specific defaults for kphio to match Stocker paper
         if kphio is None:
             if not self.do_ftemp_kphio:
-                self.init_kphio = 0.049977
+                self.init_kphio = np.array(0.049977)
```
AmyOctoCat, line 220: Could these values be pulled out into a constant rather than using magic numbers?

AmyOctoCat (approved):
I am not familiar with the theory behind this change, but looks great to me in terms of code from what I can tell. Could possibly set some constants for the magic numbers used.

pyrealm/pmodel/pmodel.py
davidorme: I've got some inbound code that changes the `kphio` part completely, so I _think_ these will disappear in the near future.

PR #267, opened 2024-07-25
Initial draft of new quantum yield module

### LINKED ISSUE #266: Additional kphio method using aridity and temperature
David Sandoval is working on a new calculation for the `kphio` parameter that includes modulation by temperature and aridity. It would be good to include this functionality for the upcoming workshop.

The bigger picture here is that the `kphio` handling in the `PModel` and `SubdailyPModel` then gets rather overloaded with arguments and settings in the signature. It is probably cleaner to break the `kphio` implementations out into a new ABC that allows new approaches to be added with disrupting the PModel API.

davidorme:
# Description

This PR implements a new approach to providing quantum yield (`kphio`) values to the `PModel` and `SubdailyPModel`.

* The `kphio` and `do_ftemp_kphio` arguments to `PModel` and `SubdailyPModel` are removed. These set a reference `kphio` value and then switched between constant or temperature modulated methods.
* They are replaced by `method_kphio` and `reference_kphio`. The method name links to a registry of subclasses of the new `pyrealm.pmodel.quantum_yield.QuantumYieldABC`, which provide different methods for calculating `kphio`.
* Each implementation just needs the definition of the `_calculate_kphio` abstract method, which has access to the `PModelEnvironment` object for the model and the shared constants definitions, as well as a flag setting whether C3 or C4 parameterisations should be used.
* Each implementation sets a default `reference_kphio`: the meanings of this reference vary with the method, so need to be documented.

In addition:
*  The sandoval method introduced different derivations of the equation for calculating enzyme kinetics. The new function `calc_modified_arrhenius_factor` implements the two different equations and this function is now used with the `QuantumYieldSandoval` and replaces the `calc_ftemp_inst_vcmax`, which was basically a $V_{cmax}$ specific version of the function.
- The `calc_ftemp_kphio` function is replaced by `QuantumYieldTemperature` has been removed.
- The tests, docstrings and user facing docs have been updated to match.

Fixes #266 

## Type of change

- [x] **Breaking change**

## Key checklist

## Further checks


pyrealm/constants/pmodel_const.py
```diff
@@ -89,6 +89,22 @@ class PModelConst(ConstantsClass):
...
+        1558.853,
+        -50.223,
+        294.804,
+        75000.0,
+    )
+    """Enzyme kinetics parameters for estimation of kphion from mean growth temperature
+    in the Sandoval method for estimation of quantum yield efficiency. Values are: the
+    intercept and slope of activation entropy as a function of the mean growth
+    temperature (J/mol/K), the deactivation energy constant (J/mol) and the activation
+    energy (J/mol). """
```
j-emberton, line 106: Can we add a citation for these constants?

pyrealm/pmodel/functions.py
```diff
@@ -47,6 +48,7 @@ def calc_ftemp_arrh(
     Args:
         tk: Temperature (in Kelvin)
         ha: Activation energy (in :math:`J \text{mol}^{-1}`)
+        tk_ref: The reference temperature for the reaction.
```
j-emberton, line 51: In Kelvin?

pyrealm/pmodel/functions.py
```diff
@@ -60,18 +62,19 @@ def calc_ftemp_arrh(
 
     Examples:
         >>> # Relative rate change from 25 to 10 degrees Celsius (percent change)
-        >>> round((1.0-calc_ftemp_arrh( 283.15, 100000)) * 100, 4)
-        np.float64(88.1991)
+        >>> np.round((1.0-calc_ftemp_arrh( 283.15, 100000)) * 100, 4)
+        array([88.1991])
```
j-emberton, line 66: Python `round` and numpy `np.round` use different rounding strategies. Python `round` uses bankers rounding whereas `np.round` uses round away from zero. This means that under certain conditions they will give different results i.e. one will round up and the other down. This probably doesn't matter too much in this instance, but its worthier keeping in mind.

pyrealm/pmodel/functions.py
```diff
@@ -194,6 +197,76 @@ def calc_ftemp_inst_vcmax(
...
+        core_const: Instance of :class:`~pyrealm.constants.core_const.CoreConst`.
+
+    PModel Parameters:
+        To: The standard reference temperature expressed in Kelvin (`T_0`, ``k_To``)
+        R: the universal gas constant (:math:`R`, ``k_R``)
+
+    Returns:
+        Values for :math:`f`
+
+    Examples:
```
j-emberton, line 238: I'm not sure why the example doesn't use the `calc_modified_arrhenius_factor` function

pyrealm/pmodel/pmodel.py
```diff
@@ -206,61 +187,41 @@ def __init__(
...
-                # A single scalar global value
-                self.init_kphio = np.array(kphio)
-            elif isinstance(kphio, np.ndarray):
-                # An array of values, which must match the shape of the inputs to the
-                # PModelEnvironment instance.
-                _ = check_input_shapes(self.env.tc, kphio)
-                self.init_kphio = kphio
-
         # -----------------------------------------------------------------------
         # Optimal ci
```
j-emberton, line 191:
```suggestion
        # Optimal chi
```

pyrealm/pmodel/functions.py
```diff
@@ -13,7 +13,8 @@
 
 def calc_ftemp_arrh(
     tk: NDArray,
-    ha: float,
+    ha: float | NDArray,
+    tk_ref: NDArray | None = None,
```
j-emberton, line 17: We are slightly inconsistent on whether or not we record optional args as optional in the doctoring description.

j-emberton:
I see you've added a csv of data related to the Sandoval k_phio approach.

Over time we're slowly increasing the size of the data stored in the build data folder. Is there a case for migrating this to a Pyrealm zenodo and downloading once for each test session? This would allow us to better manage test data provenance.

Alternatively, maybe we could just add a readme to the `pyrealm_build_data/sandoval_k_phio` folder to note where this data is sourced from for any future developers.

pyrealm/pmodel/functions.py
davidorme: Just hadn't got around to updating docstrings properly 😄

pyrealm/pmodel/functions.py
davidorme: Yeah - I agree. It's implicit in the typing that something is optional but I haven't been consistent about actual documenting it and saying what the defaults are. Reviewing this is probably a separate issue but I've fixed this one.

davidorme:
@j-emberton and @AmyOctoCat This is now fully implemented - could you review. 

We need to finalise what to do about the breaking change - we're thinking of releasing `2.0.0rc1` to give us some wiggle room on making the functionality available but not committing to the final state of a new `2.0.0`  API.

davidorme:
> Alternatively, maybe we could just add a readme to the `pyrealm_build_data/sandoval_k_phio` folder to note where this data is sourced from for any future developers.

It's de novo generated from some appropriate inputs. Having said that, at the moment we have a page in the docs about the `pyrealm_build_data` package but that's separated from the package contents. It would actually make a lot more sense to move the documentation in there into docstrings in `__init__.py` files within the `pyrealm_build_data` package and then use `autodoc` to maintain an API like documentation tree for the package.

j-emberton:
> > Alternatively, maybe we could just add a readme to the `pyrealm_build_data/sandoval_k_phio` folder to note where this data is sourced from for any future developers.
> 
> It's de novo generated from some appropriate inputs that are included in the folder.
> 
> Having said that, at the moment we have a page in the docs about the `pyrealm_build_data` package but that's separated from the package contents. It would actually make a lot more sense to move the documentation in there into docstrings in `__init__.py` files within the `pyrealm_build_data` package modules and then use `autodoc` to maintain an API like documentation tree for the package.

OK, lets not hold up this PR with doing that but perhaps add it as an issue

docs/source/users/pmodel/pmodel_details/pmodel_overview.md
```diff
@@ -26,43 +25,86 @@ reference documentation](../../../api/pmodel_api)
...
+
+Several of the approaches implemented within the P model in `pyrealm` may be estimating
+the same underlying environmental limitation of light use efficiency via different
+pathways. As an example, there are multiple approaches to incorporating effects of [soil
+moisture stress](soil_moisture) on productivity, via modulation of $\phi_0$, $m_j$ and
+the calculation of GPP penalty factors.
+
+Some combinations of methods may therefore result in multiple corrections for the same
+limitation. At present, `pyrealm` does not automatically detect such over-correction, so
+take care when selecting methods for fitting the P model.
```
j-emberton, line 100: I think we should add this as an issue. It shouldn't be too difficult to have a compatibility check for any combination of selected methods. We could also add a quick reference to the docs for users.

docs/source/users/pmodel/subdaily_details/subdaily_calculations.md
```diff
@@ -170,6 +179,8 @@ at 25°C. This is acheived by multiplying by the reciprocal of the exponential p
 the Arrhenius equation ($h^{-1}$ in {cite}`mengoli:2022a`).
 
 ```{code-cell}
+:trusted: true
+
 # Are these any of the existing values in the constants?
```
j-emberton, line 184: Does this comment need to be in the docs?

j-emberton (approved):
I've added another couple of comments but no showstoppers. I think you're good to go :)

j-emberton:
It just occurred to me that we possibly don't want to merge into develop, as that might then constrain us from further releases under the 1.x major version. Do we instead want to create a v2.0rc branch and maintain that in parallel until v2.0 release?

PR #277, opened 2024-09-05
Implementation of Flora and PlantFunctionalType

### LINKED ISSUE #276: Implementation of Flora and PlantFunctionalType
This issue is to create a PR to provide an implementation of the `Flora` and `PlantFunctionalType` classes. It is intended to isolate that particular component of @AmyOctoCat's work in progress PR #231 to allow a review and refactor of that section of the code and allow us to bring in that work in smaller chunks.

This issue should: 

* provide something close to a final definition of the classes building on @AmyOctoCat's work and review by @davidorme,
* methods to create those classes from data in files,
* testing of those code.

davidorme:
# Description

This PR provides a restructure and review of @AmyOctoCat's implementation of the Flora and PlantFunctionalType classes needed for the [demography milestone](https://github.com/ImperialCollegeLondon/pyrealm/milestone/4).

The PR isolates the implementation of those two classes in order to start merging elements of that milestone down into `develop` rather than having large PRs. The main implementations are taken from #231 and revised.

**Code changes**:

* The validation has dropped the use of `json_dataclasses` from #231 and dropped back to the parent `marshmallow` and `marshmallow_dataclasses`. The validation then happens on Python data objects, which pushes the file handling back into our code, but also makes it really easy to share that validation across different file formats.
* The calculation of the `q_m` and ` z_max_prop` traits has been pushed back all the way into `PlantFunctionalType` since they are constant within a type and are now populated automatically within `__post_init__`.
* The `Flora` class now has three factory methods for creating instances from JSON, TOML and CSV inputs. Each method handles file loading and minor data format alignment and then can use the same shared `marshmallow` validation schema. The code in #231 had a separate serialisation class, but this functionality has been replaced with the factory methods.

**Other changes**:
* The module paths in #231 have been simplified to `pyrealm.demography`. This may change but I think we had more nesting than needed.
* Tests are implemented for the `pyrealm.demography.flora` module.
* Docs should now build, including the simple canopy sketch (more or less an accidental inclusion).

Fixes #276 (issue)

## Type of change

## Key checklist

## Further checks


j-emberton (commented):
This generally looks good to me.

I'm concerned that the way it is currently defined is a little open to human error. For example, if a human is writing the pft json, it would be easy to skip a property entry by mistake with the result that this then takes a default value rather than the intended one. This wouldn't be so easy to detect or troubleshoot with the current implementation.

One thing I think we could do is create `Flora.summarise()` feature which would create a table of the pft properties (columns are properties, rows are each pft). Each property could be checked to see if it is taking a default or user specified value and if default this is highlighted some how to the user.

Another option is to have a 'strict' mode where all inputs must be defined 'a priori' where the code raises an exception if the user has failed to define a value.

This wouldn't be too hard to implement and would be a useful sanity check tool.

davidorme:
Nice catch @j-emberton - that's a really likely area for errors. I originally had the PFT dataclass without defaults, which would have prevented that failure mode, but this is effectively replacing the existing `TModelTraits` constants class.

I think it probably makes sense to enforce strict mode when using the factory methods - that's _particularly_ true of CSV, where you'd have to handle NA if you wanted to support partial definitions for some fields. Is there a canned approach to this? My immediate thought is to have a base `PlantFunctionalTypeStrict` that is used for the loading schema and then a derived `PlantFunctionalType` as now is the programmatic interface with defaults.

davidorme:
Also, I'm thinking that the PFT dataclass should have `frozen=True`.

davidorme:
@j-emberton How about the solution in the most recent pushes? I've made the decision to make the file loading strict but retained the version with defaults as a way to get that defined set of default values and get a slightly friendlier programmatic API.

j-emberton:
> @j-emberton How about the solution in the most recent pushes? I've made the decision to make the file loading strict but retained the version with defaults as a way to get that defined set of default values and get a slightly friendlier programmatic API.

I think that's a simpler fix than anything I was going to suggest

> @j-emberton How about the solution in the most recent pushes? I've made the decision to make the file loading strict but retained the version with defaults as a way to get that defined set of default values and get a slightly friendlier programmatic API.

This is a simpler solution than anything I was going to suggest and It adds some good safety rails for the user.

j-emberton:
There's apparently conflicts in the ref.bib file. I'll be good to approve once they're fixed.

davidorme:
@j-emberton Conflicts all fixed.

j-emberton (approved):
lgtm

PR #285, opened 2024-09-16
Complete implementation of T model functions

### LINKED ISSUE #284: Complete implementation of T Model functions in `demography` module and test better.

The most recent code (PR #282) include only some of the T Model functions and the testing is against a small number of hand calculated values.

Complete the transfer of T Model functions from within the to be deprecated `pyrealm.tmodel` module and then build a deeper regression test suite against the R T model implementation.

davidorme:
# Description

This PR:

* Completes the suite of functions needed for running the T Model within the new `demography` module. It is basically unpacking the current `pyrealm.tmodel.TModel.calculate_growth` method for estimating growth increments into standalone functions with proper docstrings etc. following @AmyOctoCat 's similar treatment of `pyrealm.tmodel.TModel.reset_diameters` for the allometric scaling.
* Updates the regression test datasets from the `rtmodel.r` implementation to include predictions from more than one plant functional type.
* Adds regression tests for all of the functions in `pyrealm.demography.t_model_functions` against those new regression golden outputs.

*Does not* add any user-facing documentation yet.

Fixes #284

## Type of change

## Key checklist

## Further checks


MarionBWeinzierl:
When putting in assertions, would it make sense to set a tolerance for `allclose` ? As, I assume, it can be quite different what an acceptable deviation for the expected value is in a each case.

davidorme:
> When putting in assertions, would it make sense to set a tolerance for `allclose` ? As, I assume, it can be quite different what an acceptable deviation for the expected value is in a each case.

These ones are purely deterministic and should be precise - or at least to the high default tolerance of `allclose`. They more or less should be identical to machine precision.

pyrealm/demography/t_model_functions.py
```diff
@@ -145,6 +145,260 @@ def calculate_sapwood_masses(
...
+    The net primary productivity (NPP, :math:`P_{net}`) is calculated as a plant
+    functional type specific yield proportion (:math:`y`) of the total GPP (:math:`P`)
+    for the individual minus respiration (:math:`R_m`), as the sum of the respiration
+    costs for foliage  (:math:`R_f`), fine roots  (:math:`R_r`) and sapwood
+    (:math:`R_s`).
+
+    .. math::
+        P_{net} = y (P - R_m) = y (P - W_{\cdot s} r_s - \zeta \sigma W_f r_r - P r_f)
+
+    Note that this differs from Equation 13 of :cite:t:`Li:2014bc`, which removes foliar
```
j-emberton, line 250: Is there a citation for this approach?

pyrealm/demography/t_model_functions.py
```diff
@@ -145,6 +145,260 @@ def calculate_sapwood_masses(
...
+    * the stem mass (:math:`\Delta W_s`), and 
+    * the foliar mass (:math:`\Delta W_f`). 
+        
+        
+    The stem diameter increment can be calculated using the available productivity for
+    growth and the rates of change in stem (:math:`\textrm{d}W_s / \textrm{d}t`) and
+    foliar masses (:math:`\textrm{d}W_f / \textrm{d}t`): 
+
+    .. math::
+
```
j-emberton, line 330: Is there a citation available for this approach?

j-emberton:
Looks fine to me aside from the queries on citations. Happy to approve and let you update the docstrings as you wish.

j-emberton (approved):
LGTM

pyrealm/demography/t_model_functions.py
davidorme: No - it's a "new" approach but only in the pretty trivial sense that the foliage respiration rate can be tuned.

pyrealm/demography/t_model_functions.py
davidorme: It's Lie et al. again

Issue #270, opened 2024-08-08
Improve documentation of optional arguments to callables


davidorme:
**Is your feature request related to a problem? Please describe.**

As @j-emberton pointed out over on #267, the docstrings don't do a consistent job of making it clear where arguments are optional and what happens when they aren't provided. It's clear which arguments _are_ optional from the typing, but then individual argument descriptions should consistently end with something like "Optional, defaulting to XYZ".

**Describe the solution you'd like**
Review the signatures and consistently note the optional arguments in the `Args` section of the docstrings.

PR #288, opened 2024-09-18
Implementation of the canopy model

### LINKED ISSUE #286: Implementation of the Canopy model

This issue is to review and merge the `Canopy` structure from @AmyOctoCat's PR #231 down into `develop`. PRs are now accepted for the `Flora` (PR #277) and `Community` (PR #282) objects and the T model functions are in progress (#285), so this is the next logical section of code to merge in and commit.

Take Amy's `Canopy` code out of #231 and then review and integrate it with those previous PRs.

davidorme:
# Description

[Updated to move from WIP to ready to merge].

This PR is to review, adapt and merge in @AmyOctoCat's code on the canopy object from #231. 

The main areas of change from that PR are:

* The `t_model_functions` and `canopy_functions` modules have now settled down to which functions belong in which 
 module.
* The`canopy_functions` module has now been extended to include the canopy vertical structure functions from #231, including the relative canopy radius, stem crown area and stem leaf area at given heights, along with a solver function to find the height at which a canopy layer closes. This includes:
    * Rethinking the workflow for functions to avoid repetitive calculation.
    * Removing `Community` object as an argument within the `canopy_functions` functions and replacing it with explicit arguments for the community properties for each function. This makes the core functions more flexible.
    * Adding a bunch of validation for input shapes of canopy function arguments, but making it optional for a speed-up within solvers and classes where the shapes can be assumed correct.
* The `Canopy` object has been reworked to use this new functionality to get to the same end point as in #231. 
* I've also added unit tests for the canopy functions and a really simple test of the Canopy object.

In addition, from the earlier WIP version of the PR, this code:

* is updated to drop use of `pandas` data structure in favour of simple use of `numpy` arrays throughout (see #292).
* updates the `t_model_functions` to include validation of the input argument shapes and then builds out a deeper unit test suite to a range of input combinations using predictions from the original R implementation.

Things to do later:
* Update the docstrings with better docs on the equations and then update the `canopy.md` document to use this code.
* More functionality needed in the `Canopy` object to use it to generate absorbed radiation per stem per layer and hence generate productivity estimates.

Fixes #286 (issue)

## Type of change

## Key checklist

## Further checks


pyrealm/demography/canopy.py
```diff
@@ -0,0 +1,133 @@
...
+        )
+        """Column vector of the heights of canopy layers."""
+
+        # Loop over the layers except for the final layer, which will be the partial
+        # remaining vegetation below the last closed layer.
+        starting_guess = self.max_stem_height
+        for layer in np.arange(self.n_layers - 1):
+            target_area = (layer + 1) * community.cell_area * (1 - canopy_gap_fraction)
+
+            # TODO - the solution here is predictably closer to the upper bracket, might
```
MarionBWeinzierl, line 69: Instead of putting a TODO in the code, creating an issue has a higher likelihood of this being seen and solved.

pyrealm/demography/canopy_functions.py
```diff
@@ -0,0 +1,411 @@
...
+        q_z: Relative crown radius at heights in z.
+        crown_area: Crown area for a stem
+        stem_height: Total height of a stem
+        f_g: Within crown gap fraction for each stem.
+        q_m: Canopy shape parameter ``q_m``` for each stem
+        z_max: Height of maximum canopy radius for each stem
+        validate: Boolean flag to suppress argument validation.
+    """
+
+    # TODO - could this merge with the stem crown area function? A lot of overlap, so
```
MarionBWeinzierl, line 365: Again, should be an issue

pyrealm/demography/canopy_functions.py
```diff
@@ -0,0 +1,411 @@
...
+
+
+# def calculate_gpp(cell_ppfd: NDArray, lue: NDArray) -> float:
+#     """Estimate the gross primary productivity.
+
+#     Not sure where to place this - need an array of LUE that matches to the
+
+#     """
+
+#     return 100
```
MarionBWeinzierl, line 411: Can these be removed?

tests/unit/demography/test_t_model_functions.py
```diff
@@ -1,21 +1,13 @@
...
-    calculate_heights,
-    calculate_sapwood_masses,
-    calculate_stem_masses,
-)
 
 
 def test_calculate_heights():
     """Tests happy path for calculation of heights of tree from diameter."""
+
+    from pyrealm.demography.t_model_functions import calculate_heights
```
MarionBWeinzierl, line 9: Hm, I don't know how I feel about moving imports into functions. I prefer them to be on the top of the file, for better overview. Is there a difference in performance if you draw the imports in here?

MarionBWeinzierl: (same below)

pyrealm/demography/canopy_functions.py
davidorme: They're placeholders for the next steps in this milestone, so easier for me to leave them there for now.

tests/unit/demography/test_t_model_functions.py
davidorme: That style was advice from the RSE team at Imperial. The argument is that having the imports for the **package being tested** inside the functions isolates the actual importatation as part of the test. So, if stuff goes wrong with an import, a specific related test fails, rather than a whole module of tests (or the whole test suite).

I definitely see what you mean about having an overview of which code is getting tested but I buy their argument and we've seen it be useful in practice.

pyrealm/demography/canopy.py
```diff
@@ -0,0 +1,133 @@
...
+    at which canopy layers close. This also includes
+        ground level, which will capture all remaning crown and leaf area below the last
+        closed layer.
+    """
+
+    def __init__(
+        self,
+        community: Community,
+        canopy_gap_fraction: float,
+        layer_tolerance: float = 0.001,
```
omarjamil, line 28: Perhaps worth outlining in the docstring this default value.

pyrealm/demography/canopy.py
```diff
@@ -0,0 +1,133 @@
...
+            self.layer_heights[layer] = starting_guess = solution.root
+
+        # Find relative canopy radius at the layer heights
+        self.stem_relative_radius: NDArray[np.float32] = (
+            calculate_relative_canopy_radius_at_z(
+                z=self.layer_heights,
+                stem_height=community.cohort_data["stem_height"].to_numpy(),
+                m=community.cohort_data["m"].to_numpy(),
+                n=community.cohort_data["n"].to_numpy(),
+                validate=False,
```
omarjamil, line 102: `validate` is set to `False` in these function calls. Should there be an option to set this to true via this class?

pyrealm/demography/canopy.py
```diff
@@ -0,0 +1,133 @@
...
+                    "Estimation of canopy layer closure heights failed to converge."
+                )
+
+            self.layer_heights[layer] = starting_guess = solution.root
+
+        # Find relative canopy radius at the layer heights
+        self.stem_relative_radius: NDArray[np.float32] = (
+            calculate_relative_canopy_radius_at_z(
+                z=self.layer_heights,
+                stem_height=community.cohort_data["stem_height"].to_numpy(),
```
omarjamil, line 99: Same arrays are being converted to numpy in different calls. From performance point of view it might worth doing a conversion once and using those arrays.

pyrealm/demography/canopy_functions.py
```diff
@@ -0,0 +1,411 @@
...
+    # Set Ap = Ac where z <= zm
+    A_p = np.where(z <= z_max, crown_area, A_p)
+    # Set Ap = 0 where z > H
+    A_p = np.where(z > stem_height, 0, A_p)
+
+    return A_p
+
+
+def solve_community_projected_canopy_area(
+    z: float,
```
omarjamil, line 261: Sometimes a really long argument list can be a bit cumbersome so worth considering dataclass or namedtuple for such cases. However, that comes with different overheads so a bit of a style choice.

pyrealm/demography/canopy_functions.py
```diff
@@ -0,0 +1,411 @@
...
+    elif z.ndim == 2 and z.shape[1] == 1:
+        # z is a column array, so check q_z is a matrix
+        if q_z.shape != (z.size, stem_property.size):
+            raise ValueError("Invalid shape for q_z.")
+        return
+
+    raise ValueError("Invalid shape for the z value.")
+
+
+def calculate_relative_canopy_radius_at_z(
```
omarjamil, line 167: If the inputs array can be reshaped to be broadcastable, then using numpy functions instead of python built-ins e.g. `np.power` instead of `**`can help with vectorization and improve performance.

omarjamil (approved):
Overall, this is a well-structured and thoroughly documented piece of scientific code. The use of numpy, type hinting, and input validation contributes to its robustness. My suggestions are for further improvement of the code, but not critical.

omarjamil:
I should add that I am happy to defer to James and Marion in terms of what they consider to important changes to make as they have better code and project knowledge than me.

pyrealm/demography/canopy.py
davidorme: We've actually got an issue about this (#270)  - @j-emberton pointed out that the documentation of defaults and optional values isn't consistent throughout the codebase.

davidorme: Not that this stops me fixing this specific instance 😄

pyrealm/demography/canopy.py
davidorme: I agree. This is the issue we discussed yesterday about using `pandas.DataFrame` for the `cohort_data` attribute of `Community`. I think the answer is to change that attribute - we're never making use of the pandas functionality outside of creating the `cohort_data`, so it would be much cleaner as a dictionary of `np.array`s.

pyrealm/demography/canopy.py
davidorme: I should doc this better. The validation is intended to help people use the standalone functions correctly, but here the `Canopy.__init__` should guarantee the correct inputs, and so the validation is turned off to improve run time by avoiding repeated validation within the root solver call.

If it's broken, that's a developer issue with the `__init__` code, rather than something where a user could usefully turn validation back on.

pyrealm/demography/canopy_functions.py
davidorme: This specific module has a bunch of extremely similar but not identical signatures. We could bundle everything up in one dataclass (and indeed @AmyOctoCat had these using `Community` for exactly that reason), but I'd like to retain the flexibility of using these directly.

One reason for that is that I want to build some user facing helper classes to generate canopy data outside of a Community object, so it's probably easier to keep them like this, but I will be looking at that next.

pyrealm/demography/canopy_functions.py
davidorme: I need to think about (and document) what shapes are sensible here. I believe that `M**n` is faster if `n` is a scalar but `np.power(M, n)` is faster when `n` is an array? Could get either in using this function.

omarjamil: I think np.power will be faster if you are applying the operation to an array even if the exponent is a scaler. And you can further performance gains if you can apply np.vectorize to functions. Though I feel like I need to back up this with some testing!

pyrealm/demography/canopy.py
davidorme: Comments added to code to explain this.

pyrealm/demography/canopy_functions.py
davidorme: This one is a bad idea on reflection - the crown area function will get run _a lot_ from within the solver, so keeping it as lean as possible makes a lot of sense.

j-emberton (commented):
Haven't had a in depth look at the tests yet, but generally this looks OK.
It'd be good to get some thoughts on the potential type inconsistencies for m,n and q_m in canopy_functions.py.
I would prefer to see canopy.py refactored to move the evaluations out of the constructor.

pyrealm/demography/canopy.py
```diff
@@ -0,0 +1,149 @@
...
+            closure heights (default: 0.001 metres)
+    """
+
+    def __init__(
+        self,
+        community: Community,
+        canopy_gap_fraction: float = 0.05,
+        layer_tolerance: float = 0.001,
+    ) -> None:
+        # Calculate community wide properties: total crown area and maximum height
```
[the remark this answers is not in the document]
davidorme: This code is very strongly tied to the `__init__` method. It _would_ look cleaner to declare the attributes in the `__init__` and then have a separate method for the maths, but that seems almost entirely cosmetic. And I don't mean to knock cosmetic at all - clarity matters. I just can't see how this leads to meaningful separation of concerns. I guess I'm mostly thinking testing - the `__init__` and `do_maths` would be largely inseparable for testing purposes?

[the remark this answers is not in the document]
davidorme: Done - I think it is definitely cleaner but still curious if it offers anything else in this case.

davidorme:
@j-emberton The `calculate_canopy_q_m` and `calculate_canopy_z_max_proportion` should definitely return `float | NDArray` but for the other functions, isn't the typing simply saying that _this_ function requires an array of values for `q_m`, `m`, `n` and `z_max_proportion`.  It doesn't matter what type the original function generates as long as any intermediate code packs up values into arrays for use here.

I guess the danger is that users might (not unreasonably) stick the `float` outputs of those two functions straight into the other functions, but that is intentionally an error - I don't want these functions to support float inputs.

j-emberton:
> @j-emberton The `calculate_canopy_q_m` and `calculate_canopy_z_max_proportion` should definitely return `float | NDArray` but for the other functions, isn't the typing simply saying that _this_ function requires an array of values for `q_m`, `m`, `n` and `z_max_proportion`. It doesn't matter what type the original function generates as long as any intermediate code packs up values into arrays for use here.
> 
> I guess the danger is that users might (not unreasonably) stick the `float` outputs of those two functions straight into the other functions, but that is intentionally an error - I don't want these functions to support float inputs.

Yeah, I see. I think I was being a bit overzealous here, and got into a mindset where those type inconsistencies in `calculate_canopy_q_m` would be able to propagate into the other functions. But I see that's not the case.

pyrealm/demography/canopy.py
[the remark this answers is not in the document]
j-emberton: It's definitely cleaner. I think ideally, I'd break out the root finding loop into a function to be called from `canopy_functions.py` but it's not a deal breaker at all. It would be more testable, and if you're thinking of replacing it with another root finding algorithm, more modular.

This class to me is essentially a dataclass, and where we've implemented data classes previously e.g. `calendar` we've used post-init methods to process the inputs and provide values for attributes. I like this style and it keeps things tidy.

This is all ultimately aesthetic preference. There's not really a computational implication.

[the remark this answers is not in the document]
omarjamil: There can be some computational advantages (may not be applicable here) to what James is suggesting. If you can separate out some of the `__init__` calculations into separate functions and wrap them in `@property` decorator then you can benefit from lazy evaluation.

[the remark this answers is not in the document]
davidorme: Yeah - this could very well be a dataclass. I don't think it's an issue to leave as is for the moment but I see your point about the structure. I'll leave the root finding loop as is for now. 

@omarjamil That's a good point about lazy evaluation - I don't think it applies here because we need all of the attributes calculated to move on to the next step but that it is something I'll keep in mind in future.

pyrealm/demography/canopy_functions.py
```diff
@@ -0,0 +1,415 @@
...
+        # All same height or stem specific heights - check z is either a scalar or also
+        # a row vector of the right length.
+        check_input_shapes(z, *args)
+        return
+    elif z.ndim == 2 and z.shape[1] == 1:
+        # Z is a column vector, just check stem properties.
+        check_input_shapes(*args)
+        return
+
+    raise ValueError("Invalid shape for the z value.")
```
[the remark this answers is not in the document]
davidorme: Yeah - that's more of an insult than an Exception message.

I think those validation functions probably need to be unpacked a bit - the shape requirements are intricate and we'd expect most users to interact with higher level APIs but the functions should be more helpful in their failure modes than "haha luzer".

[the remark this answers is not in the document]
davidorme: OK - that kicked off a major rewrite of the validation functions and tests. There is now a single integrated validation test rather than two separate ones, that does a much clearer job of setting out the valid combinations of shapes and providing specific guidance in failure cases. The test suite got a lot more clear too, I think.

davidorme:
@j-emberton I think with that update to the `Canopy.__init__` and the improved validation, this is now in good shape. Could you approve? Unless, of course, you don't 😥 ...

j-emberton:
> @j-emberton I think with that update to the `Canopy.__init__` and the improved validation, this is now in good shape. Could you approve? Unless, of course, you don't 😥 ...

👼 🤔 😈 

will go ahead and approve

j-emberton (approved):
Thanks for the changes

Issue #290, opened 2024-09-20
Add helper functions for PFT geometry and canopy shape.


davidorme:
We need some neat wrappers to make it easier to visualise PFT scaling without having to go through a Community object. This is partly to make the docs easier, but also to allow users to visualise the effects of the PFT settings. We need:

* A geometry method or class that takes a PFT and an array of diameter at breast heights (DBH) or stem heights and generates all the other predictions of the T model, for use in plotting allometries and scaling.
* A similar canopy method or class that takes a PFT, a stem height / DBH value and generates the vertical canopy data for plotting.

davidorme:
OK - been playing around with idea on this, and I think a nice API for users on this issue is to extend `Flora`:

```python
class Flora():
    ...
    def get_allometries(dbh: NDArray) -> dict[str, NDArray]:
        """Generate T Model predictions for PFTs in Flora for a set of DBH values."""
        ...
    def get_canopy_shape(z: NDArray) -> dict[str, NDArray]:
        """Generate canopy shape predictions for PFTs in Flora for a set of vertical height values."""
        ...

```

Those aren't methods of `PlantFunctionalType` as described above _but_  immediately makes predictions across a set of PFTs that a user is working with, which seems like 90% of the use cases. If someone genuinely does only want one PFT, then creating a single PFT flora is pretty low cost.

If the return values in the dictionary are a set of `(n_dbh, n_pfts)` arrays for different predictions then it is also a very clean plotting structure for `matplotlib`. A usage sketch would be:

``` python
import numpy as np
from matplotlib import pyplot as plt
from pyrealm.demography.flora import PlantFunctionalType, Flora

# Create flora
pft1 = PlantFunctionalType(name='pft1', h_max=20)
pft2 = PlantFunctionalType(name='pft2', h_max=30)
flora = Flora([pft1, pft2])

# Get allometries for DBH values from 0 to 1 m in 1cm increments
dbh = np.linspace(0, 1, num=101) 
allometries = flora.get_allometries(dbh=dbh)

# Plot stem_height vs dbh for both PFTs in one go.
plt(dbh, allometries['stem_height'], label = flora.pft_names)
```

[...]

PR #282, opened 2024-09-12
Implementation of Community object init and factory methods.

davidorme:
# Description

This PR brings in a Community object based of @AmyOctoCat's work in PR #230. From the features listed in #278, we have a (mostly) complete implementation with some elements postponed for later PRs:

- [x] provide something close to a final definition of the Community building on @AmyOctoCat's work and review by @davidorme,

> [!NOTE]
>   - The community object takes cohort data as arrays, merges in PFT traits for those arrays and then calculates the cored predictions of the T Model for the initialisation data.
>   - I've switched to using a `pandas.DataFrame` for storing cohort data.  I think that reverses something I told @AmyOctoCat (sorry!), but although there is overhead in using the Dataframe rather than simple arrays as attributes, it does simplify the namespace of the class a lot and it is also much cleaner to add and delete cohorts as rows rather than having to iterate over individual attributes.
  
- [ ] methods to create instances from data in files - and also handle multiple communities in files.

> [!NOTE]
>   - This **does not** implement loading multiple `Community` objects yet - might as well sense check the single case before doing that!
>   - The serialisation formats for CSV, JSON and TOML are revised
>   - Factory methods are now implemented to load from those formats.
>   - Marshmallow schemas are used to validate and post-process all of those formats into the format needed to initialise a `Community` object.

- [x] test that code.

> [!NOTE]
>  - It currently tests the T Model functions code rather lightly and it would be better to have deeper regression tests against the R implementation. I'd like to break that out into a new PR as complete tests would include some elements missing from the T model  functions at the moment and I don't want to bloat this basic Community PR.
  
- [x] provide API docs 

This PR also updates `Flora` to provide a `pandas.Dataframe` view of the flora data, which makes it super easy to merge onto the cohort data. 

There's some complexity in the serialisation formats and schemas:

- the `Community` object needs single values for `cell_area` and `cell_id` but arrays for the cohort data.
- CSV is great for rows of cohort data, but isn't structured for the single values, so I've imposed those as needing to be constant fields in the CSV. That is easy for users to maintain and feeds forward into having multiple cells in a file.
- JSON/TOML allows structured data, but editing arrays of cohort data is ugly and confusing because they're not neatly aligned in grids. So here, I've made the cohort data a list of objects, which is much easier to maintain in these formats.

The validation schemas take these two slightly different approaches, validate them and then coerce to the common arguments used for initialisation.

Fixes #278

## Type of change

- [x] New feature (non-breaking change which adds functionality)
[...]

PR #292, opened 2024-09-21
Replace use of `pandas` in `pyrealm.demography`

### LINKED ISSUE #291: Look at replacing `pandas` in `pyrealm.demography`
Using `pandas.DataFrame` and `pandas.Series` has a lot of unintended consequences for broadcasting array shapes, which is a bigger part of `demography` than anticipated. As noted here:

https://github.com/ImperialCollegeLondon/pyrealm/issues/290#issuecomment-2364225749

This is an exploratory issue to see where removing it ends up.

davidorme:
# Description

In #277, I used `pandas.Dataframe` within `Flora` to provide an array-like view onto the plant functional type data. That then led to using it for `Community.cohort_data` (#282) and also typing the inputs to the T Model functions as `pandas.Series`. That turns out to be awkward for a number of core use cases:

* See comment https://github.com/ImperialCollegeLondon/pyrealm/issues/290#issuecomment-2364225749 about the awkwardness of using this structure in creating allometry predictions for a `Flora`
* See comment https://github.com/ImperialCollegeLondon/pyrealm/pull/288#discussion_r1768449896 about the repeated casting of `Series` to `np.ndarray` when broadcasting needed.
* See issue #291 which summarises the issues and describes what this PR should do.

This PR starts to replace the usage with a pure numpy alternative. There is code in the currently open PR #288 that will also need updating, but this seems like the right way to go now.

The code:
* Replaces the use of pandas structures throughout.
* Adds a prototype validation function and unit test for that function to validate array inputs to T model functions.
* I *have not yet* added that validation to the functions - wanted to wait for feedback before moving on.

Fixes #291 (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/demography/community.py
+++ b/pyrealm/demography/community.py
@@ -7,9 +7,8 @@ files, using :mod:`marshmallow` schemas to both validate the input data and to p
 post processing to align the input formats to the initialisation arguments to the
 Community class.
 
-Internally, the cohort data in the Community class is represented as a pandas dataframe,
-which makes it possible to update cohort attributes in parallel across all cohorts but
-also provide a clean interface for adding and removing cohorts to a Community.
+Internally, the cohort data in the Community class is represented as a dictionary of
+`numpy` arrays.
 
 Worked example
 ==============
--- a/pyrealm/demography/community.py
+++ b/pyrealm/demography/community.py
@@ -354,9 +354,8 @@ class Community:
 
     # Post init properties
     number_of_cohorts: int = field(init=False)
-
-    # Dataframe of cohort data
-    cohort_data: pd.DataFrame = field(init=False)
+    cohort_data: dict[str, NDArray] = field(init=False)
+    pft_data: dict[str, NDArray] = field(init=False)
 
     def __post_init__(
         self,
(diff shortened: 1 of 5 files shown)


j-emberton (changes requested):
The implementation looks fine to me overall.

One issue to fix:
There's a straggler issue in that the community.py module docstring still makes reference to using pandas as the internal data  vehicle.

`Internally, the cohort data in the Community class is represented as a pandas dataframe,
which makes it possible to update cohort attributes in parallel across all cohorts but
also provide a clean interface for adding and removing cohorts to a Community.`

And one query:
I see the type hints for the new NDArrays are now `float32`. Is there a specific reason for this? Will Python type promotion not mean that this doesn't survive contact with any `float64` data?

davidorme:
> One issue to fix: There's a straggler issue in that the community.py module docstring still makes reference to using pandas as the internal data vehicle.

Fixed

> And one query: I see the type hints for the new NDArrays are now `float32`. Is there a specific reason for this? Will Python type promotion not mean that this doesn't survive contact with any `float64` data?

That's a good point - the typing of arrays has got cleaner and I think the earlier code was written when this was not such a transparent thing to do. I don't think we need the precision of `float64` so we use less RAM by doing this, but equally I don't know if that has speed implications (good or bad) with 64 bit architecture.

So - not sure. I guess I'd like to leave this PR as is and tackle the array typing more widely as another issue.

j-emberton (approved):
Thanks for sorting the docstring. Happy for the float32 types hints to be left in place for the time being. My gut feel is that any data read in from csv/toml etc will default to float64, and would need to be explicitly converted to float32.

PR #296, opened 2024-09-25
Add Flora helper methods

### LINKED ISSUE #290: Add helper functions for PFT geometry and canopy shape.

davidorme:
# Description

---
> [!Warning]
>  See the [discussion here](https://github.com/ImperialCollegeLondon/pyrealm/discussions/297) and below about the implementation of this functionality. ~I don't think it should be merged as is, so this PR is likely to be a placeholder from which to pick out smaller chunks for a cleaner implementation.~
> 
> ETA: After discussion with @MarionBWeinzierl , we agreed to merge this to bank the functionality - knowing the structure isn't quite right - and then add incremental PRs to introduce the better structuring.
>
> ETA 2: Tried to do it that way - but the code in here is poorly arranged across the modules and trying to fix it up one bit at a time leads to circular imports and other intertwined mess. So this merge has been reverted and we'll build up in well structured bits.
---

Following the updates discussed in #290 and implemented in  #291, this implements methods of the `Flora` object to allow users to visualise the behaviour of the `PlantFunctionalType` instances within the `Flora`:

- [x] Calculate allometric scaling of stem sizes and masses with stem DBH.
- [x] Calculate the allocation of potential GPP to growth, respiration and turnover across stem sizes.
- [x] Calculate canopy profiles across vertical stem heights

So... this PR has broken out of its cage a bit and I need to think where to go next.

The good
---

1. We now have three 'bundled' functions to get:

    * T model allometries for PFTs , given a range of stem sizes (`pyrealm.demography.t_model_functions.calculate_t_model_allometry`)
    * T model carbon allocation for PFTs, given potential GPP and stem sizes (`pyrealm.demography.t_model_functions.calculate_t_model_allocation`)
    * Crown vertical profiles for PFTs, given a representative stem size and heights.
(`pyrealm.demography.canopy_functions.calculate_canopy_profiles`)

    Each of those has a wrapping `Flora` method to generate the predictions for the set of PFTs in a `Flora` instance

2. Some initial user docs generating those profiling outputs - partly genuinely for user guidance, partly to test out how the API feels (`docs/source/users/demography/crown.md` and `docs/source/users/demography/flora.md`)
3. A couple of extra functions to help with T model allometry (`calculate_dbh_from_height`) and crown profiles (`calculate crown radius`).
4. Some test refactoring to use shared resources.

The bad
---

The gaps are:

* Testing is incomplete
* The docstrings are incomplete

The ugly
---

Coding this reveals a lot of repeated structures that I think would be better refactored into well defined data classes.

I think we have:

* `Flora`: stops being a dictionary of PFT instances. Those instances are almost never used directly, but the `Flora.data` dictionary of trait arrays is used _all the time_. It would be better as a data class with those traits as attributes. 
* `get_t_model_allometries` returns an `Allometries` dataclass rather than a dict of variables.
* `get_t_model_allocation` returns an `Allocation` dataclass rather than a dict of variables.
* `get_canopy_profiles` returns an `CrownProfiles` dataclass rather than a dict of variables.

Those structures can then be passed in as arguments more easily and are better defined and give clearer typing. I don't think we should change all of the function args - the low level functions should retain specific variable arguments, but I think those new 'bundled' functions can act as an interface from the dataclasses to the specific sets of functions. For example:

```python
def get_t_model_allometries(flora: Flora, dbh: NDArray) -> Allometries:
    ...

def get_t_model_allocations(flora: Flora, stems: Allometries, potential_gpp: NDArray) -> Allocations:
    ...

def get_canopy_profiles(flora: Flora, stems: Allometries, z: NDArray) -> CrownProfiles:
    ...
```

I think that's cleaner and makes it easy to reuse the same code within `Canopy`. @omarjamil was suggesting something similar on one of the earlier PRs, but those structures seem more compelling to me now. There is a cost: ultimately the `Canopy` model will need to add and remove cohorts and having multiple dataclasses gives more attributes to update to keep cohorts aligned. But that seems relatively minor.

So I think the specific review I'm after here is:

* does the new functionality and API look good in a kind of hand wavey general fashion.
* does that refactor sound like a better way to go than the current dicts of variables.

If so - it doesn't seem sane to try and do that all in one PR. 

Thoughts?

Fixes #290 

## Type of change

## Key checklist

## Further checks


davidorme:
See https://github.com/ImperialCollegeLondon/pyrealm/discussions/297

MarionBWeinzierl (approved):
As agreed I did not do an in-depth review, but from looking over it it looks fine. Improvements on the structure will be done in add-on PRs -- see also https://github.com/ImperialCollegeLondon/pyrealm/discussions/297

PR #300, opened 2024-09-27
Update traits data structures

### LINKED ISSUE #298: Update traits data structures
Following the comment in merging PR #290 and in discussion here:
https://github.com/ImperialCollegeLondon/pyrealm/discussions/297

This PR is to implement trait data classes with consistent trait attributes for handling PFTs, Flora and StemTraits. These all share the same set of attribute but differ in the typing and details. The initial signature sketchs are below and this will also provide a `Flora.get_stem_traits()` method to generate a `StemTraits` instance from a `Flora` - that's basically currently part of `Community` and would be better placed in `Flora`.

```python
@dataclass(frozen=True)
class PlantFunctionalTypeStrict:
    """PFT with all traits explicitly declared - mostly used for deserialising."""
    name: str
    a_hd: float
    ...
    f_g: float

@dataclass(frozen=True)
class PlantFunctionalType(PlantFunctionalTypeStrict):
    """PFT with default values, used as a programatic convenience."""
    a_hd: float = 116.0
    ...
    f_g: float = 0.05

@dataclass(frozen=True)
class Flora():
    """
    * Is constructed from a list of PFTs.
    * Guarantees unique PFT names
    * Otherwise has all the same trait attributes as PFTS but as row arrays.
    * Also has a bonus extra dictionary onto the original PFT instances
    """
    pfts: InitVar[list[PlantFunctionalTypeStrict]]
    name: NDArray[np.str_] = field(post_init=True)
    a_hd: NDArray[np.float32] = field(post_init=True)
    ...
    f_g: NDArray[np.float32]  = field(post_init=True)
    pft_dict: dict[str, PlantFunctionalTypeStrict] = field(post_init=True)

@dataclass()
class StemTraits():
    """
    * Core data class for Communities - holds the stem traits for each cohort
    * Provides all the same trait attributes as row arrays.
    """
    name: NDArray[np.str_] 
    a_hd: NDArray[np.float32]
    ...
    f_g: NDArray[np.float32]
```

davidorme:
# Description

This PR implements the updated trait data structures requested in #298. It update tests and things using the trait data structures.

It forms part of a sequence re-implementing the functionality set implemented in #296 but on less shaky structural grounds and in smaller more easily reviewed sections.

Fixes #298 (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/demography/flora.py
+++ b/pyrealm/demography/flora.py
@@ -21,7 +21,7 @@ import json
 import sys
 from collections import Counter
 from collections.abc import Sequence
-from dataclasses import dataclass, field, fields
+from dataclasses import InitVar, dataclass, field, fields
 from pathlib import Path
 
 import marshmallow_dataclass
@@ -192,16 +192,18 @@ in gaps from the default values.
 """
 
 
-class Flora(dict[str, type[PlantFunctionalTypeStrict]]):
-    """Defines the flora used in a ``virtual_ecosystem`` model.
+@dataclass(frozen=True)
+class Flora:
+    """A dataclass providing trait data on collection of plant functional types.
 
-    The flora is the set of plant functional types used within a particular simulation
-    and this class provides dictionary-like access to a defined set of
+    A flora provides trait data on the complete collection of plant functional types
+    that will be used within a particular simulation. The dataclass provides access
+    to trait attributes as row arrays across those plant functional types.
+
+    The class is created using a list of
     :class:`~pyrealm.demography.flora.PlantFunctionalType` or
-    :class:`~pyrealm.demography.flora.PlantFunctionalTypeStrict` instances.
-
-    Instances of this class should not be altered during model fitting, at least until
-    the point where plant evolution is included in the modelling process.
+    :class:`~pyrealm.demography.flora.PlantFunctionalTypeStrict` instances, which must
+    have unique names.
 
     Args:
         pfts: A sequence of ``PlantFunctionalType`` or ``PlantFunctionalTypeStrict``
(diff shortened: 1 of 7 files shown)


pyrealm/demography/flora.py
```diff
@@ -308,3 +359,94 @@ def from_csv(cls, path: Path) -> Flora:
...
+        # matching the pft_names and pass that into the StemTraits constructor.
+
+        return StemTraits(
+            **{trt: getattr(self, trt)[pft_index] for trt in self.trait_attrs}
+        )
+
+
+@dataclass()
+class StemTraits:
+    """A dataclass for stem traits.
```
j-emberton, line 391: Would it make sense to make `Flora` inherit `StemTraits`, or have both inherit from an ABC, so that this list of  traits is only defined in one place?

davidorme: Ha. I did look at all sorts of things here - Protocols and ABC and inheritance, but these classes are basically intended to deal with two different constructor signatures, even if they share nearly all the attributes. That seems to make any kind of inheritance more finicky than just duplicating stuff. I may have missed it - I'd love a way to share that set of attributes cleanly across the two PFT traits, Flora and Stem Traits.

Right now though, I'd like to define that common API on the attributes and get the next PR in to close out the functionality hanging in #296 - and if the API looks good, I'm all for circling back to improve the implementation later.

j-emberton (approved):
Happy to approve to keep the demography work stream moving. I'm wary of accumulating technical debt (i.e. not coming up with an improved implementation of the stem traits that avoids duplication) but appreciate we need to get the remaining stuff in as well.

PR #303, opened 2024-09-28
Minor updates to T Model functions and testing

### LINKED ISSUE #302: Minor T Model updates

The code developed in #296 added a new T Model function (`calculate_dbh_from_height`) and reworked the test structures included for the `t_model_functions` module to use the new Flora interface. This issue is to isolate those changes before bringing in the rest of the functionality in #296.

davidorme:
This PR is to:

* Adds a T Model function to reverse calculate DBH from stem height.
* Moves the R T model data into `tests/unit/demography/conftest.py` so it can be shared and splits it into a dataframe of expected values and a new Flora object created from the PFT definitions.
* Updates the unit testing for the T Model functions to use these new structures.

Fixes #302 (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/demography/t_model_functions.py
+++ b/pyrealm/demography/t_model_functions.py
@@ -78,6 +78,56 @@ def calculate_heights(
     return h_max * (1 - np.exp(-a_hd * dbh / h_max))
 
 
+def calculate_dbh_from_height(
+    h_max: NDArray[np.float32],
+    a_hd: NDArray[np.float32],
+    stem_height: NDArray[np.float32],
+    validate: bool = True,
+) -> NDArray[np.float32]:
+    r"""Calculate diameter at breast height from stem height under the T Model.
+
+    This function inverts the normal calculation of stem height (:math:`H`) from
+    diameter at breast height (DBH, :math:`D`) in the T Model (see
+    :meth:`~pyrealm.demography.t_model_functions.calculate_heights`). This is a helper
+    function to allow users to convert known stem heights for a plant functional type,
+    with maximum height (:math:`H_{m}`) and initial slope of the height/diameter
+    relationship (:math:`a`) into the expected DBH values.
+
+    .. math::
+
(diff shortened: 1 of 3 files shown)


pyrealm/demography/t_model_functions.py
```diff
@@ -78,6 +78,56 @@ def calculate_heights(
...
+    """
+
+    if validate:
+        _validate_t_model_args(pft_args=[h_max, a_hd], size_args=[stem_height])
+
+    # The equation here blows up in a couple of ways:
+    # - H > h_max leads to negative logs which generates np.nan with an invalid value
+    #   warning. The np.nan here is what we want to happen, so the warning needs
+    #   suppressing.
+    # - H = h_max generates a divide by zero which returns inf with a warning. Here the
```
j-emberton, line 122: I'm not quite sure why its desirable for the np.nan to be generated.

Would it make sense to add another optional argument to allow the user to an alternative behaviour where the code saturates any stem height at the max height for that PFT?

There could be three alternative behaviours desired the function which could be chosen by the user:
1) Allow np.nan to be calculated and suppress errors (current implementation)
2) Allow np.nan to be calculated and raise errors
3) Force saturation of stem height instead of np.nan

This is well outside my field of expertise so feel free to push back :)

j-emberton: Aside from this query, code looks fine. Happy to approve

davidorme: In the existing function for the  DBH to H calculation is fine, $H \to H_{max}$ as $D \to \inf$. This new function inverts that calculation from H to DBH, and so when $H >= H_{max}$, $D$ is genuinely infinite or undefined and so there isn't a sensible asymptotic $D$ to saturate to. I did initially think there was some thing we could do, but we do genuinely need to set `np.nan` for impossible trees.

j-emberton (approved):

Issue #315, opened 2024-10-02
Calculate light extinction/interception


davidorme:
The next step in the demography module is to implement the calculation of light extinction through the canopy. There are a few scientific details to check but that shouldn't change the main form of the implementation.

The PR needs to take a Canopy object and:

* Calculate the leaf area per canopy layer per cohort.
* Calculate the light extinction profile through the leaf layers.
* Generate a scaling for canopy top incident photosynthetic flux density that allows the flux to be partitioned among stems within canopy layers

Issue #317, opened 2024-10-03
Review validation and return shapes from demography objects


davidorme:
At the moment, the validation of dimension inputs and checking hasn't been completely worked through the `demography` module. There's a fair bit of variation in what shapes get returned by different combinations of inputs at present and it might be worth rationalising those options.

Issue #295, opened 2024-09-24
Validate sign of inputs to T Model functions.


davidorme:
At the moment, there's nothing stopping people giving negative heights, dbh etc to the T Model functions. We probably should add this within the validation of the inputs to the functions of that module.

I'm not quite sure what level of user error we want to guard against though. Stuff that blows up the code, sure, but there is a runtime and code bloat cost to guarding against everything.

PR #328, opened 2024-10-11
Revise and document functionality of canopy and crown

### LINKED ISSUE #316: Update documentation for canopy and crown
Ongoing work on the canopy model (#315) is nearly ready to add as a PR. The new functionality on that issue will then need an update on the documentation for the crown and canopy modules. That documentation is currently in review (#313), so this is all stacking up a bit, but I wanted to keep a big pile of doc changes from bulking out the review of the code for #315.

The docs will need to be updated to:

* Help users understand the connection between crown radius and projected area better.
* Add a simple helper function to provide some more easily plottable structures for crowns.
* Extend the canopy description to talk about the light extinction and community crown area more generally before introducing the PPA model.

davidorme:
# Description

This issue:

* Revises the crown and canopy modules. The main changes are
  * Move a canopy function from `crown` into `canopy`: `solve_canopy_area_filling_height`
  * Add a plotting helper function to `crown`: `get_crown_xy`. I'm not 100% convinced about the API of this yet, but think it's useful.
  * Revised the internals of the `canopy` model to try and generalise how it works and fixed some bugs.

* Updates the user facing documentation for the crown and canopy modules. 
* Updates the existing tests to the new structures. Doesn't add any tests as yet.

**Some things to note**:

* I've made the open PR #318 outdated. That PR is currently hanging with errors that I think are fixed in here and this PR includes all the code from #318. That turned out to be necessary because I needed to fix the code to I write the docs.

* I've updated the way in which the metadata in the markdown notebooks is handled: `jupyter` itself can add metadata to notebooks and code cells. Because we are using Myst Markdown through `jupytext`, that has its own filtering  mechanism that controls what actually gets preserved in the file. That's now caused all of our markdown files to get minor changes (hence the **47 files changed** 😱 ). 

  There are really only a few files that matter: the crown and canopy docs (`docs/source/users/demography/...`, the crown and canopy modules (`pyrealm/demography/...`) and the minor rearrangements in the tests.

---
Things to do (maybe not in this PR)

- [ ] There are a couple of areas in the code that I haven't resolved. These are described in TODO boxes here in the [built_docs](https://pyrealm.readthedocs.io/en/316-update-documentation-for-canopy-and-crown/users/demography/canopy.html) for this PR.
- [ ] There are some missing tests.
- [ ] The docs need updating to explain the things I haven't resolved, as and when they get resolved.

Will fix both #316 and #315

## Type of change

## Key checklist

## Further checks


pyrealm/demography/canopy.py
```diff
@@ -37,130 +191,136 @@ class Canopy:
...
+            self.heights = fit_perfect_plasticity_approximation(
+                community=community,
+                canopy_gap_fraction=canopy_gap_fraction,
+                max_stem_height=self.max_stem_height,
+                solver_tolerance=solver_tolerance,
+            )
 
         self._calculate_canopy(community=community)
 
     def _calculate_canopy(self, community: Community) -> None:
```
j-emberton, line 266: I would prefer to see the individual operations in `_calculate_canopy()` broken out into separate methods for testability

j-emberton:
Is this still draft PR? Aside from a little (suggested) restructuring in `canopy` it looks good to go

davidorme:
That's tricky - this isn't complete or final. 

* There's a section on the light allocation that is wrong because it ought to vary with crown gap fraction and doesn't.
* I'm not 100% sold on the workflow yet.
* There is #295 and #317 kicking around
* There isn't enough testing.

But it feels like it's just going to keep sprawling, so I'm tempted to address the issues and bank this and add bit size issues to fix the above. Thoughts?

pyrealm/demography/canopy.py
davidorme: I can see why.

The reason I haven't is because there are two trivial calculations and then a chain of calculations that build on each other. It _could_ be split into `calculate_leaf_area_index` and `calculate_light_transmission` but both of those would be much cleaner as methods of `Canopy` rather than standalone functions as the signatures are a bit unwieldy if we don't pass in the content of `self`.

If we do that then we have:

```python
class Canopy:
    ...
    def _calculate_canopy(self, ...):
        ...
       self._calculate_leaf_area_index()
       self._calculate_light_transmission()

    def _calculate_leaf_area_index(self, ...):
        ...

    def _calculate_light_transmission(self, ...):
        ...
```

That's still all chained together - it's not immediately clear to me that there is a huge advantage. Did you mean to break them out into standalone functions?

davidorme: @j-emberton I think I'll leave it as is because I'm going to be revisiting this section in #337.

j-emberton: Ok. Just reviewing all this now.

pyrealm/demography/crown.py
```diff
@@ -452,3 +400,114 @@ def __repr__(self) -> str:
...
+
+    The data are returned as a list with one entry per stem. The default value for each
+    entry a tuple of two arrays (height, attribute values) but the `as_xy=True` option
+    will return an `(N, 2)` dimensioned XY array suitable for use with
+    {class}`~matplotlib.patches.Polygon`.
+
+    Args:
+        crown_profile: A crown profile instance
+        stem_allometry: The stem allometry instance used to create the crown profile
+        attr: The crown profile attribute to plot
```
j-emberton, line 442: Can we add a simple link to the list of valid attributes?

j-emberton (approved):
This looks ok to me.

I 've added another comment regarding helping people understand what the list of valid str attributes that can be passed but its not a deal breaker.

PR #335, opened 2024-10-18
Updating and systematising NDArray typing

### LINKED ISSUE #334: Standardise NDArray typing
At the moment, we have rather variable typing of `NDArray` variables.

We’ve have some variables typed as `NDArray[[np.int]]` and `NDArray[np.str_]` but then the `demography` module has been using `NDArray[np.float32]` and much of the rest of the codebase has bare `NDArray`.

The `np.float32` in the `demography` wasn't very well thought through. The rationale is simply to reduce memory requirements and (maybe?) run faster, but because it is so piecemeal it likely isn't effective at either due to type promotion etc. We have #304 to think about addressing this more systematically, but for now we should just type `NDArray` and do so consistently.

davidorme:
# Description

This PR makes the `NDArray` typing across the project systematic:

* Changes `NDArray[np.float32]` to `NDArray[np.float64]`
* Changes bare `NDArray` to the appropriate type (99% `NDArray[np.float64]`, 1% `NDArray[np.datetime64]`)
* Updates a couple of `.astype(np.float32)`
* Imports `numpy` in a couple of modules as needed

Fixes #334

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -96,6 +96,7 @@ nitpick_ignore = [
     ("py:class", "numpy._typing._array_like._ScalarType_co"),
     ("py:class", "numpy._typing._generic_alias.ScalarType"),
     ("py:class", "numpy.float32"),
+    ("py:class", "numpy.float64"),
     ("py:class", "numpy.int64"),
     ("py:class", "numpy.timedelta64"),
     ("py:class", "numpy.bool_"),
--- a/pyrealm/constants/core_const.py
+++ b/pyrealm/constants/core_const.py
@@ -114,7 +114,7 @@ class CoreConst(ConstantsClass):
     :cite:t:`berger:1978a`."""
 
     # Hygro constants
-    magnus_coef: NDArray[np.float32] = field(
+    magnus_coef: NDArray[np.float64] = field(
         default_factory=lambda: np.array((611.2, 17.62, 243.12))
     )
     """Three coefficients of the Magnus equation for saturated vapour pressure,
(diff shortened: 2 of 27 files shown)


j-emberton:
Are we happy with the idea that people now will be instructed not to pass ints into functions doing ordinary arithmetic or array ops? For most functions, integers will work fine.

davidorme:
I think that depends on what we mean by people? Since the typing is static, this change doesn't actually stop _users_ from passing in integer arrays to those arguments (or indeed _anything_ that isn't explicitly blocked by downstream input checking!), but it does indicate to developers that something is expecting to work with floats. We could expand the typing to make it explicit that these are "numeric" rather than implicitly allowing integers but I _think_ that only affects developers.

j-emberton:
This is incredibly pernickety.

Do we care if we pass values as ints into a function that requires a float. Examples of this in the code are at line  92 in `pmodel_environment.py` where we call `bounds_checker` and pass in `-25` and `80` as ints, when the function args are type hinted as floats.

j-emberton:
> I think that depends on what we mean by people? Since the typing is static, this change doesn't actually stop _users_ from passing in integer arrays to those arguments (or indeed _anything_ that isn't explicitly blocked by downstream input checking!), but it does indicate to developers that something is expecting to work with floats. We could expand the typing to make it explicit that these are "numeric" rather than implicitly allowing integers but I _think_ that only affects developers.

It might make sense to just have np.number rather than float64. This is tidy for developers and docs, plus allows use of decimal, fractions etc.

davidorme:
I don't think we do care if `int` gets passed in. We don't test for it and they'll get converted (unless I guess someone passes in really long integers that can't be represented in `np.float64`?).  I guess I'm inclined to leave it as is for now.

j-emberton (approved):
lgtm

PR #338, opened 2024-10-21
Shared base classes for demography classes

### LINKED ISSUE #314: Shared base classes for demography classes.
The various demography dataclasses (`Flora`, `StemTraits`, `StemAllometry`, `StemAllocation`, `Canopy`) require some shared functionality:

* Convert data to `pandas.Dataframe`: this is more a utility function than anything mission critical, but it is an obvious way that users might want to handle the data. It would have to cope with unstacking 2D arrays (or insist that users select a single PFT/Cohort to export, but that seems restrictive). Doing this would require synchronizing the name of the `xyz_attrs` class variables in those classes to a shared `var_attrs` name. For the case with only row arrays, it should be as simple as defining a base model with a `to_pandas` method that does: 

```python
def to_pandas(self):

    return pd.DataFrame({k: getattr(self, k) for k in self.var_attrs})
```

* A baseclass that provides methods to combine the `var_attrs` from identical classes and drop indices across classes. The use case here is for recruitment and mortality in demography. The sketch below uses StemTraits, but this would extend to other data classes.

```python
existing_stems = StemTraits(...)
new_stems = StemTraits(...)
existing_stems.add(new_stems)
existing_stems.drop([0,4,7])
```

Not all classes that could support `pandas` conversion should support add and drop, so I think we're looking at a couple of base classes and then multiple inheritance: `StemTraits(ToPandas, CohortBehaviour)`.

davidorme:
# Description

> [!note]
> I probably should have done this as two PRs - it's not too bloaty, but there is a clear division between the two classes here. 

This PR:

* Introduces the `demography.core` module and defines the `PandasExporter` and `CohortMethods` abstract base classes. These two ABCs share a common class attribute (`array_attrs`) which is used to define which fields to export to `pandas.DataFrame` and which fields need to be kept synchronised as cohorts are added or removed.
* Adds some low level tests of some basic subclasses in `tests.unit.demography.test_core.py` to show that the ABCs work independently and with multiple inheritance.

* Applies the `PandasExporter` to many of the `demography` classes:
    * They  inherit the new `to_pandas()` method from the base class.
    * Updates documentation to use the method.
    * Adds simple checks to existing tests to make sure the method works on all subclasses.

* Applies the `CohortMethods` subclass to the `Cohorts`, `StemTraits` and `StemAllometry` classes. These are the three main dataclasses within `Community`, where the functionality to synchronise the addition and deletion of cohorts is required.
    * Adds tests for each classes  that the inherited `add_cohort_data` and `drop_cohort_data` methods are working as expected.

* Adds `Community.add_cohorts` and `Community.drop_cohorts` methods, which wrap the individual dataclass methods to update the instance.

Fixes #314 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/api/demography_api.md
+++ b/docs/source/api/demography_api.md
@@ -30,6 +30,14 @@ language_info:
     :members:
 ```
 
+## The {mod}`~pyrealm.demography.core` module
+
+```{eval-rst}
+.. automodule:: pyrealm.demography.core
+    :autosummary:
+    :members:
+```
+
 ## The {mod}`~pyrealm.demography.flora` module
 
 ```{eval-rst}
--- a/docs/source/users/demography/crown.md
+++ b/docs/source/users/demography/crown.md
@@ -274,9 +274,9 @@ above calculated at each height $z$:
 crown_profiles
 ```
 
-The {meth}`~pyrealm.demography.crown.CrownProfile.to_pandas` method can be used to
-extract the data into a table, with the separate stems identified by the column index
-field.
+The {meth}`~pyrealm.demography.core.PandasExporter.to_pandas()` method of the
+{meth}`~pyrealm.demography.crown.CrownProfile` class can be used to extract the data
+into a table, with the separate stems identified by the column index field.
 
 ```{code-cell} ipython3
 crown_profiles.to_pandas()
--- a/docs/source/users/demography/crown.md
+++ b/docs/source/users/demography/crown.md
@@ -218,12 +218,10 @@ flora = Flora([narrow_pft, medium_pft, wide_pft])
 flora
 ```
 
-The Flora object can also be used to show a table of canopy variables:
+The key canopy variables from the flora are:
 
 ```{code-cell} ipython3
-# TODO - add a Flora.to_pandas() method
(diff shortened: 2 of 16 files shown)


davidorme:
Hmm. Apparently `sphinx` does not generate links to inherited methods: https://github.com/sphinx-doc/sphinx/issues/10699

j-emberton:
Hi @davidorme , is there anything I should specifically focus on thin this review that has previously been reviewed in the merged branches?

davidorme:
Not sure I understand the question, @j-emberton? The base classes are new to this branch - unless I've managed to muck up working on two branches in parallel.

davidorme:
So `pyrealm.demography.core` is the new functionality and then everything else is just rolling that functionality out to the classes that need it.

j-emberton:
Sorry I think my brains out of gear this morning. I was looking at the diff thinking that I'd seen a bunch of the code before, and couldn't quite work out what was going on.

j-emberton (approved):
lgtm

MarionBWeinzierl (approved):
This looks sensible to me. Cannot find much to comment on.

omarjamil (approved):

PR #341, opened 2024-10-23
Group internal attributes in canopy into data classes

### LINKED ISSUE #337: Group internal attributes in Canopy into cohort and community objects.
Not so sure about this one, but `Canopy` has some attributes that are per cohort (stem leaf area, etc.) and some that are for the whole canopy. So some attributes are always `(n_heights, n_cohorts)` and some are always `(n_heights,)` in shape. If we group those into specific dataclasses, then those can inherit the `PandasExporter` class and it's also a bit more structured.

davidorme:
# Description

This PR revises the internal structure of the `Canopy` class, grouping the attributes under two new dataclasses: `CohortCanopyData` and `CommunityCanopyData`, which  use `__post_init__` methods to populate the attributes. This:

* gives a cleaner internal structure 
* groups the attributes that share the same dimensions (cohort is `(n_heights, n_cohorts)`, community is just `(n_heights,)`) 
* allows us to get these dataclasses to inherit methods with shared functionality such as export to pandas (upcoming #338).
* simplifies the internals of `Canopy._calculate_canopy` by bundling cohort and community calculations. This is something that @j-emberton requested in PR #328.
* adds a new test class to validate the calculations in these two dataclasses, which is a finer grained check than the previous implementation.

The calculation order here goes:

1. main cohort attributes on light extinction
2. whole community light extinction and fraction absorbed
3. divide the whole community absorption among cohorts and stems

The first iteration had: `CohortCanopyData`, `CommunityCanopyData` and then `CohortCanopyData.allocate_fapar` method, but that leaves some canopy attributes hanging until the allocation method is called. The code now generates the `CommunityCanopyData` _within_ `CohortCanopyData` so it all happens in a single pass. Within `Canopy` the community data is moved up into a top level attribute for ease of access.

```python
self.cohort_data = CohortCanopyData(...)
self.community_data = self.cohort_data.community_data
```

Fixes #337

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/users/demography/canopy.md
+++ b/docs/source/users/demography/canopy.md
@@ -139,7 +139,7 @@ should equal the whole canopy $f_{abs}$ calculated using the simple Beer-Lambert
 equation and the PFT trait values.
 
 ```{code-cell} ipython3
-print(simple_canopy.extinction_profile[-1])
+print(simple_canopy.community_data.extinction_profile[-1])
 ```
 
 ```{code-cell} ipython3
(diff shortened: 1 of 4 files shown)


j-emberton (approved):
seems reasonable to me

davidorme:
@j-emberton I'm just looking at #314 and that alternative arrangement of calculating everything within `CohortCanopyData` seems cleaner now - because it won't leave the `cohort_fapar` and `stem_fapar` attributes hanging undefined. That just seems like an error waiting to happen (and also makes the `to_pandas` method ugly because not all of the variables that could possibly by exported are definitely there in an instance. 

I could define those attrs with a default or get the pandas exporter to check that each exported attribute actually exists on an instance, but it seems easier just to go with that alternative.

Any reasons not to?

j-emberton:
> @j-emberton I'm just looking at #314 and that alternative arrangement of calculating everything within `CohortCanopyData` seems cleaner now:
> 
> ```python
> self.cohort_data = CohortCanopyData(...)
> self.community_data = self.cohort_data.community_data
> ```
> 
> This is mostly because it won't leave the `cohort_fapar` and `stem_fapar` attributes hanging undefined. That just seems like an error waiting to happen (and also makes the `to_pandas` method ugly because not all of the variables that could possibly by exported are definitely there in an instance.
> 
> I could define those attrs with a default or get the pandas exporter to check that each exported attribute actually exists on an instance, but it seems easier just to go with that alternative.
> 
> Any reasons not to? I've pushed the updated implementation here and the specific changes are: [a82c6ea...0ba5f11](https://github.com/ImperialCollegeLondon/pyrealm/compare/a82c6ea...0ba5f11)

Yeah I can't see any reason not to do this. Happy for you to keep going.

PR #342, opened 2024-10-25
Review validation and shapes in demography

### LINKED ISSUE #317: Review validation and return shapes from demography objects

davidorme:
# Description

This PR:

* Creates some new validation tools in `demography.core`:
    *  `_validate_demography_array_arguments` which is intended to replace `demography.t_model_functions._validate_t_model_args` and `demography.crown._validate_z_qz_args`. These two overlap a lot. The function will check that:
        *  trait arguments (the columns in demography data arrays) and 'size' arguments like height and stem diameter (the rows in demography data arrays) are congruent.
        * that additional "at size" arguments are then congruent with the trait and size data. An example use case here is allocating GPP values to existing allometries for stems at a given size - users need to provide congruently shaped GPP inputs.
    * `_enforce_2D`, which will be used to ensure a common return dimensionality from functions: various legal inputs can give rise to 0, 1 or 2D results from calculations and it seems easier to fix the single return array.
    * Adds tests for these tools

* Updates validation and shapes in demography:
    *  `StemTraits` is now validated when created directly, but `Flora.get_stem_traits()` bypasses validation because the inputs are guaranteed correct.
    * The functions in `t_model_functions` now use  `_validate_demography_array_arguments` and have new validation flag arguments. The `StemAllometry` and `StemAllocation` classes now also optionally validate the inputs _once_ and then the inputs should be congruent so the individual function validation is skipped.
    * The same validation has then been rolled out across the other modules: `crown`, `community` and `canopy`. The tests are updated to match and the docs issues are fixed.

### Dimensionality and validation

There's an issue with differing approaches to validation here.

1.  Most of the validation is about checking the array shapes of inputs are congruent with each other. The `_enforce_2D` function is about ensuring that the output dimensions are consistent with the sizes vs stems array size model. 
2. The first drafts of this PR (and the functions it is replacing/overhauling) use `check_input_shapes` within the validation. This is a `pyrealm.core` function that was setup for use with the `PModel` parts of the code and it insists that arrays are not simply broadcastable, but are **identically shaped** or scalar. 
3. This actually leads to one of the features of the current broader `pyrealm` API that I dislike and would like to get rid of: users _have_ to broadcast inputs to get shapes to match to get things to run. I suspect we could relax `check_input_shapes` and simply apply `np.broadcast_shapes` as a validation tool. The existing code should then handle broadcasting correctly and users would not need to pre-broadcast data. They _would_ still have to know which axis is which and allowing `xarray` inputs would make this easier.
4. But - specifically if we use `check_input_shapes` here - then the internals of the demography are going to start insisting on identical shapes where values are actually sanely broadcastable. And that ends up with insisting that all the initial demography inputs are identically shaped - which is a significant usability issue and makes for uglier and more convoluted code.

So - I'm going to try and get the `demography` working with the less constraining 'broadcastable' checking. I think that's something we should try and adopt more widely, but there is a difference in approach for now.

Fixes #317

## Type of change

## Key checklist

## Further checks


pyrealm/demography/t_model_functions.py
```diff
@@ -79,8 +84,11 @@ def calculate_heights(
 
     if validate:
         _validate_t_model_args(pft_args=[h_max, a_hd], size_args=[dbh])
+        _validate_demography_array_arguments(
+            trait_args={"h_max": h_max, "a_hd": a_hd}, size_args={"dbh": dbh}
+        )
 
-    return h_max * (1 - np.exp(-a_hd * dbh / h_max))
+    return _enforce_2D(h_max * (1 - np.exp(-a_hd * dbh / h_max)))
```
davidorme, line 91: @j-emberton and @MarionBWeinzierl

I just wanted to pause here and get a sanity check before rolling this out further. Basically, there are two axes here: traits of a cohort (columns) and some measure of size for which something is being calculated. All of the functions accept arrays with varying dimensions, and the calculations can give rise to output arrays with 0, 1 or 2 dimensions. The calculations all work, but then the response dimensionality is unpredictable. My plan here is to promote all outputs to (n_traits, n_sizes), so that there is consistent outputs.

Does that seem sane?

j-emberton: Yes I think the utility of having a consistent shape so that all downstream calcs can be simplified to assume this shape is reasonable.

MarionBWeinzierl (commented):
Overall, it looks sensible to me. I agree that the `_enforce_2D` is not ideal, but I cannot think of a more elegant way of doing this right now.

MarionBWeinzierl (changes requested):
I think the error messages in two of the tests might not be complete, see above.

tests/unit/demography/test_t_model_functions.py
```diff
@@ -727,6 +661,54 @@ def test_calculate_dbh_from_height_edge_cases():
...
+        pytest.param(np.array(1), does_not_raise(), None, id="pass_0D"),
+        pytest.param(np.ones(1), does_not_raise(), None, id="pass_1D_scalar"),
+        pytest.param(np.ones(3), does_not_raise(), None, id="pass_1D_row"),
+        pytest.param(np.ones((1, 3)), does_not_raise(), None, id="pass_2D_row"),
+        pytest.param(np.ones((4, 1)), does_not_raise(), None, id="pass_2D_col"),
+        pytest.param(np.ones((4, 3)), does_not_raise(), None, id="pass_2D_full"),
+        pytest.param(
+            np.ones(4),
+            pytest.raises(ValueError),
+            "The array shapes of the trait (3,) and size (4,) arguments",
```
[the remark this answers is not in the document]
davidorme: It is but its enough to confirm the expected raise and the test uses `startswith`. I guess I'm just trying to avoid extra lines in the test parameters.

MarionBWeinzierl (approved):

PR #356, opened 2024-11-12
Manual/semi-automatic performance regression checking

### LINKED ISSUE #256: Develop manual performance regression checking
After #242 , it was decided to disable the automatic profiling in the GH Actions. Manual profiling can still be done running `poetry run /usr/bin/time -v pytest -m "profiling" --profile-svg`.

What we want now is the ability to compare the results of two profiling runs of different commits in order to check whether performance has deteriorated (or improved) through code changes.

MarionBWeinzierl:
# Description

This branch includes code to run profiling on PyRealm, and check whether performance has degraded between

Fixes #256 

## Type of change

## Key checklist

## Further checks


MarionBWeinzierl:
The scripts are currently lacking any error checking etc, but before I do more, here are some questions @davidorme : 

1. Is it ok to have this as a bash script calling Python, or would you rather have a Python script (potentially calling some bash for the git parts)?
2. This script uses the existing pytest-profile functionality. As you mentioned on Slack, we could also just call cProfile. I am wondering whether this changes something in the precision of the results.
3. It seems that there is a bit of a fluctuation we would need to account for with  a tolerance. Running the current script on the same two commit hash three times gives, for example, the following results:

image
image
image

MarionBWeinzierl:
Talking about overheads.... I noticed that the most expensive parts in the list were those associated with running pytest etc:
image

I had assumed, when I created the simplified version, that these parts would be kind of similar in each run and therefore would not need excluding, but I was wrong. So I reinstantiated the `exclude` parameter to make sure those are thrown out. This is what I get now, running the same thing three times:
image
image
image

Given that the two commits that I am comparing are quite similar, I would argue that we might still want a, say, 5% tolerance or so.

MarionBWeinzierl:
Do we want this in the CI (falling over if the codes get 5% slower), or as a manual thing?

CONTRIBUTING.md
```diff
@@ -99,13 +99,8 @@ details on the workflow and process.
 Even if the code works as expected and passes all our tests, it can still be slow! We
 use code profiling to work out where time is spent when using `pyrealm` and identify
 where we can improve performance. We also use benchmarking between `pyrealm` versions to
-make sure that changes to the code aren't making it slower. This is run automatically
-when new code is pulled to the `develop` or `main` branches but can also be used to do
-local profiling and benchmarking.
-
-See the [profiling and benchmarking
-page](https://pyrealm.readthedocs.io/en/latest/development/profiling_and_benchmarking.md)
```
MarionBWeinzierl, line 107: That link doe not work. Looking at https://github.com/ImperialCollegeLondon/pyrealm/blob/develop/docs/source/development/profiling_and_benchmarking.md , a lot of that information is also no longer valid. We will need to adapt that when we have decided how to proceed (i.e., keep the old benchmarking code, if and when to run the new code automatically, etc.)

profiling/simple_benchmarking.py
MarionBWeinzierl, line 1: @davidorme would you prefer me to use this to replace the old `run_benchmarking.py` , or leave the old script in and keep this as `simple_benchmarking.py`? That decision will also influence how the above-mentioned documentation is extended or adapted/rewritten.

CONTRIBUTING.md
```diff
@@ -99,13 +99,16 @@ details on the workflow and process.
...
-for more details.
+make sure that changes to the code aren't making it slower.
+
+This is currently runmanually using the `performance_regression_checking.sh` script in
```
j-emberton, line 104:
```suggestion
This is currently run manually using the `performance_regression_checking.sh` script in
```

davidorme:
Something is up with the profiling in the CI. The profiling YAML is re-running the whole of the `pyrealm_ci.yaml` under the `test` job. I think we can probably simply add the profiling job as a new job under the `docs` job  `pyrealm_ci.yaml`, which avoids having to import the previous workflow and ensures that the previous testing runs?

Also, the purpose of this job is to check that the profiling script works? We only need it to run so we can drop all of the graphviz stuff from the job and make it faster and cleaner.

davidorme:
I think that's cleaner? It's not really doing a different workflow, it's adding a new CI test to the standard test and build suite, so it seems reasonable to just add it there.

With that change, I think we can delete:

```
.github/workflows/pyrealm_profiling_after_push.yaml
.github/workflows/pyrealm_profiling_on_approve.yaml
.github/workflows/pyrealm_profiling_without_benchmarking.yaml
```

CONTRIBUTING.md
```diff
@@ -99,13 +99,25 @@ details on the workflow and process.
...
+This is currently run manually using the `performance_regression_checking.sh` script in
+the `profiling` directory. When this bash script is run without an argument, the current
+`HEAD` will be compared to the `origin/develop` branch. Alternatively, the two commits
+to be compared can be provided as parameters with `-n` for the "new" commit (the current
+HEAD, or code you have changed), and `-o` for the "old" commit (the baseline code you
+want to compare the perfomance against). The code will fail with an error message if the
+new performance is more than 5% slower than the baseline, otherwise it will succeed with
+a message indicating whether the new code is faster or has similar performance.
+
+To look deeper into how much time is spent in each function, the command
```
davidorme, line 113: Just to make it clear that this needs to be run from within `profiling` and within a `poetry` shell to make sure that the required python packages are present.

````suggestion
a message indicating whether the new code is faster or has similar performance.

```sh
cd profiling
poetry run performance_regression_checking.sh
```

To look deeper into how much time is spent in each function, the command
````

davidorme (approved):
LGTM - this works for me and I think it is a much saner approach. We need to work on the docs and possibly moving things into Python and adding functionality, but this gets us back on track.

MarionBWeinzierl:
I have created #358 as a follow-up regarding the old code and documentation. I am merging this in now, as incremental improvement.

PR #379, opened 2024-12-18
Added mm/m2 to moles, tests, fixed docs

### LINKED ISSUE #378: Add a converter function from precipitation (mm/m2) to moles.
For the phenology models, we will need a converter function from water in mm/m2 to moles of water. That should account for temperature and pressure.

While we are on this, the documentation for the `calc_density_h2o` functions reports that they return units of g/cm3. But they clearly don't:

```python
In [4]: water.calc_density_h2o_chen(tc=20, p=101246)
Out[4]: np.float64(998.250462136224)
```
The density of **Osmium** is approximately 22.59 g/cm3 and water should be ~ 1 g/cm3. The units of the return values are in g/L or equivalently Kg/m3. This hasn't been a problem because both functions that use the calculation do actually expect Kg/m3.

davidorme:
# Description

This adds `convert_mm_water_to_moles`, tests it and fixes some incorrect unit reporting in the `calc_density_h2o`docs.

Fixes #378

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/core/water.py
+++ b/pyrealm/core/water.py
@@ -336,6 +336,9 @@ def convert_water_mm_to_moles(
         np.float64(55.417)
     """
 
+    # Check inputs, return shape not used
+    _ = check_input_shapes(water_mm, tc, patm)
+
     # Calculate density at given temperature and pressure in g/cm3
     water_density = calc_density_h2o(tc=tc, patm=patm, core_const=core_const) / 1000
     # Hence molar volume as mol/cm3 or equivalently mol/mL
--- a/pyrealm/core/water.py
+++ b/pyrealm/core/water.py
@@ -30,7 +30,7 @@ def calc_density_h2o_chen(
             equations.
 
     Returns:
-        Water density as a float in (g cm^-3)
+        Water density in kg m3
 
     Raises:
         ValueError: if the inputs have incompatible shapes.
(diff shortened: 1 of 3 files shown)


pyrealm/core/water.py
```diff
@@ -148,7 +148,7 @@ def calc_density_h2o(
             functions are numerically unstable.
 
     Returns:
-        Water density as a float in (g cm^-3)
+        Water density in kg m3.
```
MarionBWeinzierl: Is this also kg/m^3 ?

MarionBWeinzierl (commented):
Looks fine in general, just one question

pyrealm/core/water.py
davidorme: Not sure I understand? This is a wrapper around one of the _fisher or _chen methods, both of which return kg m3, so this does too.

MarionBWeinzierl: OK. I just thought as it says it returns a density it should be km/m3, as for example in line 120

davidorme: Oh - wait. You mean I should be using `kg/m3` or  something like `kg m^-3`, rather than what I have? If so, yup - sloppy on my part.

The typo above for `km/m3` had me completely baffled for a bit 🤣

MarionBWeinzierl: Sorry, yes, kg/m3 I meant! Sorry for the typo

MarionBWeinzierl:
```suggestion
        Water density in kg/m^3.
```

MarionBWeinzierl (approved):

PR #382, opened 2025-01-06
Replacing `np.allclose`

### LINKED ISSUE #380: Replace usage of `np.allclose` in testing with `assert_allclose`
I was expecting these two commands to be equivalent:
 
```python
In [3]: import numpy as np
   ...: from numpy.testing import assert_allclose
   ...: a = np.array([[2,2]])
   ...: b = np.array([[2,2], [2,2]])
   ...: a.shape == b.shape
   ...: False
   ...: assert np.allclose(a, b)  # DOES NOT RAISE!
   ...: assert_allclose(a, b)
   ...: 
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
...

(shapes (1, 2), (2, 2) mismatch)
 ACTUAL: array([[2, 2]])
 DESIRED: array([[2, 2],
       [2, 2]])
```
 
Turns out `np.allclose` broadcasts the inputs before testing, so this does not guarantee the inputs are the correct shape. This is  a good reason to prefer `np.testing.assert_allclose(a, b)` over `assert np.allclose(a, b)`. I can sort of see why it does it - the broadcasting is such a fundamental part of numpy that it happens first? - but it is counterintuitive.

I've used `np.allclose` a lot in the tests. I doubt we've got any issues, but it's tempting fate.

davidorme:
# Description

This PR is to replace the use of `assert np.allclose` in testing with `np.testing.assert_allclose`. The testing version checks that shapes are equal, where `np.allclose` allows broadcasting before testing values (see #380).

- [x] Replace calls in `test_*.py` files and import `np.testing.assert_allclose`.
- [x] "Fix" issues where previously passing tests now fail due to differences in default permitted tolerances. This PR does not change any of the package code, so no calculations differ: it is simply differences in the permitted tolerance and we accepted the values before.
- [x] Look at the new fails where the shape is _not_ as expected and `np.allclose` let it pass.

> [!IMPORTANT]
> * All of the shape issues turn out to have been sloppy specification of the expected values. The code was working as intended but the shapes of the test values were incorrect.
> * I've rebased this onto the branch for #383 rather than `develop` - this isn't something that is a bug exactly in 1.0.0.

Fixes #380 (issue)

## Type of change

## Key checklist

## Further checks


tests/regression/pmodel/test_pmodel.py
```diff
@@ -525,11 +526,11 @@ def test_jmax_limitation(
     expected_lue = (
         (0.05 * ftemp_kphio) * optchi.mj * jmax.f_v * xf * env.core_const.k_c_molmass
     )
-    assert np.allclose(expected_lue, expected["lue"], equal_nan=True)
+    assert_allclose(expected_lue, expected["lue"], equal_nan=True, rtol=1e6)
```
MarionBWeinzierl: Does the tolerance need to be `1e-06` as above? Seems to be rather larger otherwise.

MarionBWeinzierl (commented):
Seems fine, just checking about one of the tolerances

tests/regression/pmodel/test_pmodel.py
davidorme: The defaults are `rtol=1e-07, atol=0` so this passes under stricter conditions.

MarionBWeinzierl: But that is `1e6=1,000,000`. I believe there is is a minus missing.

davidorme: Sorry - my bad. Was looking at the wrong branch and didn't understand. Yes - this is wrong and now updated to `rtol=1e-5`

MarionBWeinzierl (approved):

PR #389, opened 2025-01-18
Implementation of Arrhenius methods

### LINKED ISSUE #384: Change the calculation of Jmax25 and Vcmax25 in PModel to the simple form
Currently, the implementation of vcmax25 in the PModel uses the so called "peaked" Arrhenius factor (via `pyrealm.pmodel.functions.calc_modified_arrhenius_factor`), following the implementation of `rpmodel`.

This:
* is not the canonical version of the PModel calculations, and
* is inconsistent with the scaling used in the SubdailyPModel, which uses the simple form implemented in `pyrealm.pmodel.functions.calc_ftemp_arrh`,
* and (possibly?) does something really peculiar by setting $t_{growth}$ equal to $T_{leaf}$ temperatures, which results in no actual peak in the form of the relationship.

We need to:

* Change `PModel` to calculate the factors in the same way as `SubdailyPModel`.
* Add a `PModel.jmax25` attribute (currently missing)
* Remove any regression testing against `rpmodel` values for `vcmax25`. We do not want to try and replicate that behaviour and the other tests check against the key values (`vcmax`, `jmax` GPP).

davidorme:
# Description

This PR fixes #384, but also goes quite a bit further in trying to future proof the API of the`PModel` and `SubdailyPModel` a bit. The basic story is:

* The `PModel` currently follows `rpmodel` in using a modified Arrhenius equation parameterised by Kattge and Knorr (2007). This is inconsistent with `SubdailyPModel`, is implemented a bit oddly (`t_growth` should *not* be the same as the observed temperature) and there's no general agreement that using this form is better than the simple form at present.

* So, the simple equation (currently in `calc_ftemp_arrh`) should be used **but it is highly likely** that we're going to want to be alter the Arrhenius scaling in the near future.

So this PR:

* Renames the existing Arrhenius scaling functions (`calc_ftemp_arrh` -> `calculate_simple_arrhenius_factor` and `calc_modified_arrhenius_factor` -> `calculate_kattge_knorr_arrhenius_factor`) and tidies the function implementation up a bit.
* Adds a new `pyrealm.pmodel.arrhenius` module that implements the `ArrheniusFactorABC` and
`ARRHENIUS_METHOD_REGISTRY`. This ABC is really just a wrapper around the standalone functions that is tailored to run the calculations from within the context of a `PModel` or `SubdailyPModel` environment and the registry allows the implementations to be accessed via a short method name.
* Implemented Subclasses set a class `method` attribute, which defines that short name.  This name is used to:
   * automatically register the subclass using `__init_subclass__`,
   * expose that method option through the registry via the new `method_arrhenius` argument to PModel, and
   * to look up the required coefficients in the PModelConsts.
* When an instance is created, it sets up commonly used variable (`tk` and `tkref`) and then the `calculate_arrhenius_factor` method can be called with the coefficients for a given enzyme system. The base method checks the coefficients automatically and then calls the private `_calculation_method` abstract method defined by the subclass.

The PR then also:

* Fixes up the tests for this and adds some simple testing of the new ABC.
* Adds a brand new test that checks that the PModel and SubdailyPModel give the same outputs for fixed data. Previously, they did not.
* Removes the recent implementation of the different modes for the Kattge Knorr Arrhenius scaling. The suggested use of "J1942" over "M2002" turns out to have been a bit of a storm in a teacup and there is general agreement that "M2002" should always be used.

I think the altered PModel API is sensible - even if the internals change a bit, the signatures should be stable.

Fixes #384

## Type of change

## Key checklist

## Further checks


pyrealm/constants/pmodel_const.py
```diff
@@ -117,6 +117,33 @@ class PModelConst(ConstantsClass):
...
+            kattge_knorr=dict(
+                entropy_intercept=659.70,
+                entropy_slope=-0.75,
+                ha=49884,
+                hd=200000,
+            ),
+        )
+    )
+    """Coefficients of Arrhenius factor scaling for :math:`J_{max}`."""
+
```
MarionBWeinzierl: I am sorry to be a bit stupid (Python is not my first language), but I am confused about this. Can you explain to me what the field in all those dicts and lambdas does?

davidorme: It's a peculiarity of the `dataclass` API. Default dataclass attributes get created from the class definition and so the value of mutable defaults ends up being **shared and edited across all instances of the dataclass**. So it forbids mutable defaults using the simple `attr: type = val` system and instead you have to wrap the default value in a lambda through the `field` constructor, which gives finer control over the attribute.

https://docs.python.org/3/library/dataclasses.html#mutable-default-values

MarionBWeinzierl: Ah. Oh dear. :laughing:

MarionBWeinzierl:
> But... right now does this seem like a sensible approach?

From what I understand this looks sensible, and much cleaner.

pyrealm/constants/pmodel_const.py
davidorme: @MarionBWeinzierl - other than that odd syntax, does this seem sensible?

pyrealm/pmodel/functions.py
```diff
@@ -469,8 +488,11 @@ def calc_kp_c4(
     _ = check_input_shapes(tc, patm)
 
     # conversion to Kelvin
```
MarionBWeinzierl: This comment does not fit here anymore, as the next line does more. Either move it to line 493, or delete it?

MarionBWeinzierl (approved):
This looks good to me, just one minor complaint about an outdated comment.

PR #393, opened 2025-01-20
Convert JmaxLimitation to a registry of subclasses

### LINKED ISSUE #392: Convert JMaxLimitation from a class with methods to a registry of JMaxLimitationABC subclasses.
The `optchi`, `quantum_yield` and `arrhenius` (see #389) method arguments to PModel and SubdailyPModel all follow an extendable registry structure. `JmaxLimitation` is a more inflexible class with baked in methods for different approaches. 

It would be more flexible and consistent to adopt the same registry based approach.

davidorme:
# Description

As noted in #392, the current implementation of calculating JMaxLimitation is a single, slightly clumsily structured, class that provides an internal lookup to named methods that calculate different implementations of the limitation terms. This PR converts that structure into the same registry based approach used with optimal chi, the quantum yield calculation (and also see the draft PR for Arrhenius calculations #389, which does the same).

The PR:

* Defines an ABC for the core methods and attributes.
* Adds three subclasses implementing the existing `wang17`, `smith19` and `none` approaches.

This uses a dataclass as the ABC, mostly because this implementation allows the subclass data attributes to be extended, before automatically running the calculation method via the `__post_init__` method. We _could_ achieve the same with a standard class by amending `__call__` to autorun the method (e.g. see https://stackoverflow.com/a/76993404/3401916). 

It also:

* Updates tests to pass
* Updates docs
* Adds a description to CHANGES.md (but not the migration guide - will do that all at once before 2.0.0)

Fixes #392 (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,17 @@ worked through. The changes below are provisional.
   changed: the arguments `kphio` and `do_ftemp_kphio` have been replaced by
   `method_kphio` and `reference_kphio`.
 
+- The implementation of $J_{max}$ and $V_{cmax}$ limitation has been updated to provide
+  a more flexible and expandable system.
+  
+  The changes are mostly internal, but there are two **breaking changes**:
+
+  - The PModel option `method_jmaxlim = 'c4'` has been removed - it only ever generated
+    an instruction to use the settings `method_optchi='c4'` and
+    `method_jmaxlim='simple'`  to duplicate the `rpmodel` argument `method_jmaxlim='c4'`.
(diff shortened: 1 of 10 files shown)


MarionBWeinzierl (approved):
This looks good to me. I assume the failing docs will be fixed after the commit?

davidorme:
> This looks good to me. I assume the failing docs will be fixed after the commit?

I forgot to `make clean` to force the doc build to check all files, and only fixed the API page. Should be good now.

PR #400, opened 2025-01-27
Adding `convert_water_mm_to_moles` and `calculate_water_molar_volume`

### LINKED ISSUE #399: Add `convert_water_mm_to_moles`
We recently added `pyrealm.core.water.convert_water_mm_to_moles` and it would be useful to have the reverse operation too.

davidorme:
# Description

This PR:

* Adds `convert_water_mm_to_moles` as the reciprocal of `convert_water_moles_to_mm`
* Adds `calculate_water_molar_volume` which otherwise is duplicate within those two functions.
* Adds tests - simple benchmarks for all three functions against real world approximate values and then extends the existing tests to check that the round trip through both functions works as expected.

Fixes #399

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/unit/core/test_water.py
+++ b/tests/unit/core/test_water.py
@@ -7,6 +7,7 @@ check the size of outputs and that the results meet a simple benchmark value.
 
 import numpy as np
 import pytest
+from numpy.testing import assert_allclose
 
 
 @pytest.mark.parametrize(argnames="shape", argvalues=[(1,), (6, 9), (4, 7, 3)])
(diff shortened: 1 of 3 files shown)


tests/unit/core/test_water.py
```diff
@@ -80,6 +81,46 @@ def test_calc_viscosity_h20_matrix(shape):
...
+def test_calculate_water_molar_volume():
+    """Simple sense check that molar volume at standard conditions ~= molar mass."""
+
+    from pyrealm.constants import CoreConst
+    from pyrealm.core.water import calculate_water_molar_volume
+
+    assert_allclose(
+        calculate_water_molar_volume(tc=0, patm=101325),
+        CoreConst.k_water_molmass,
+        atol=0.01,
```
MarionBWeinzierl: Is there a reason you are using atol here and rtol elsewhere?

MarionBWeinzierl (approved):
LGTM, just one question about tolerances in the tests

tests/unit/core/test_water.py
davidorme: No 😄

Issue #390, opened 2025-01-18
Centralise data inputs to PModel in PModelEnvironment


davidorme:
A set of linked changes to `PModelEnvironment`

* Following the proposed API sketch in #394, the suggestion for `PModelEnvironment` is that with the new `PModel` API _all_ of the forcing variables can be set in the `PModelEnvironment`, so `fapar` and `ppfd` are set here. This makes it really easy to switch from standard to subdaily P Models.
* Free up the definition of extra variables to the `PModelEnvironment`. At the moment, downstream methods can require extra variables to calculate a method. The method registry setup makes this easily extendable, but at the moment these extra variables are hardcoded into the `PModelEnvironment` setup. We should change this to add _any_ variables with the right dimensions via `kwargs` - this does lose specific variable bounds checking, but that is a relatively small price.
* Centralise calculation of `tk` in `PModelEnvironment`. Many of the calculations in the PModelEnvironment require scaling factors calculated using temperature in Kelvin. At the moment, all of these functions take temperature in °C and then convert internally. It would be much saner to calculate once when creating a PModelEnvironment instance and then store as an attribute and pass `tk` to functions not `tc`.

davidorme:
Closed by #412

PR #401, opened 2025-01-27
Aligning PModel and Subdaily PModel

### LINKED ISSUE #385: Consider the removal of the PModel.estimate_productivity method
At the moment, `PModel` has a separate method for scaling light use efficiency calculations up to GPP:

```python
model = PModel(....)
model.estimate_productivity(fapar=..., PPFD=...)
```

That saves a few calculations if only LUE is required. But, the same can't be applied to the calculation of `SubdailyPModel` and it would be better to have the two models have closer signatures. So, I'm consulting, but I think we lose this method and align the signatures so that both use:

```python
model = PModel(..., fapar=..., PPFD=...)
```

We can then add `getters` for some of the properties so that they are calculated on demand to keep the memory footprint and calculation speed lower.

### LINKED ISSUE #386: Align the attributes of the PModel and SubdailyPModel implementations
We currently have completely inconsistent naming for the attributes of the `PModel` and `SubdailyPModel`. The table below isn't yet complete - I think the thing to do is probably make the `__init__` methods logic free and have a complete listing of attributes and docstrings there, that then calls a private `_fit_model` method to isolate the logic.

ETA: Actually, I think we can likely create a base class (or have Subdaily inherit from Standard) to further reinforce the shared components. I think it might also make sense to move some of the `Subdaily` signature into a new `AcclimationModel` class that gathers all of the timing and scaling infrastructure. An instance of `AcclimationModel` isn't really independent of a given Subdaily model (n_dates has to match), but I think it's a cleaner API.

See also #385 and #384 which overlap considerably. Probably close those two first - a bit around the houses but reducing the number of things being changed in one PR. This one is a big one.

Partial table of attributes:

| Use | Current PModel | Current SubdailyPModel |
|--------|--------|--------|
| `lue` | `lue` | ??? |
| `iwue` | `iwue` | ??? |
| `vcmax` | `vcmax` | `subdaily_vcmax` | 
| `vcmax25` | `vcmax25` | `subdaily_vcmax25` | 
| `vcmax25_daily_optimal` | NA | `vcmax25_opt` |
| `vcmax25_daily_realised` | NA | `vcmax25_real` |
| `jmax` | `jmax` | `subdaily_jmax` | 
| `jmax25` | Not currently exposed | `subdaily_jmax25` | 
| `jmax25_daily_optimal` | NA | `jmax25_opt` |
| `jmax25_daily_realised` | NA | `jmax25_real` |
| `xi` | `self.pmodel_acclim.optchi.xi` - add a getter? | `subdaily_xi` |
| `xi_daily_optimal` | NA | `self.pmodel_acclim.optchi.xi` Maybe add a getter? |
| `xi_daily_realised` | NA | `xi_real` |
| `gpp` | `gpp` | `gpp` |
| `gs` | `gs` | ??? |
| `ppfd` | `ppfd` | ??? |
| `fapar` | `fapar` | ??? |
| `rd` | `rd` | ??? | 
| `Ac` | Not currently exposed | `subdaily_Ac` |
| `Ac` | Not currently exposed | `subdaily_Aj` |
| `J` | Not currently exposed | `Not currently exposed |

davidorme:
# Description

This is a draft of a new structure for the PModel and SubdailyPModel:

* It centralises a lot of the shared attributes and docstrings in the `PModelABC`, which includes a single abstract class `_fit_model`.
* Each of the `PModel` and `SubdailyPModel` subclasses then:
    * Defines the specific `_fit_model` for the class.
    * Defines an `__init__` method that:
        * calls `super.__init__(...)` to initialise the superclass attributes
        * adds any specific attributes for the subclass, and
        * lastly, calls `self._fit_model()`

The result aligns the internal attribute names and shares a lot of docstrings and boilerplate. It also dumps the separate `PModel.estimate_productivity` step and `fapar` and `ppfd` are now taken directly from the `PModelEnvronment` (see PR #390 - currently still open, but nearly in, I think).

* The old implementation has been replaced in tests and all tests pass (except in one currently skipped test, which hasn't worked since the early days - we should delete it).

Things still to do but probably not in this PR.

* Strip out the old implementations - this PR is already bulky, with more changes coming so I would prefer to leave them in place - they are still used in the docs currently, until more of the 2.0.0 API changes have merged and then remove them in one go.

* Not all attributes defined on each model are populated. I need to chase down the science on these and update them, but I think for the moment I will just leave them as unpopulated.

* It would be nice to use `@cachedproperty` on some of these attributes to keep overheads down. For example, `PModel.rd` might only be calculated on request. This seems like a separate  PR, and shouldn't break the API, so can be post `2.0.0`. 

Fixes #386 
Fixes #385

## Type of change

## Key checklist

## Further checks


j-emberton:
Hi @davidorme.

Just taken a look at this. Overall I think it looks really good and is a big improvement on what we had before.

Areas for improvement:
- I feel the constructor of the `PModelABC` base class is a bit cluttered and contains a lot of similar repeated operations when evaluating and retrieving different methods and models. There's potential to rationalise these operations into class methods and make them more testable. But this isn't a showstopper.
- Do we want to add explicit unit checking? This could be as simple as checking feasible physical values for inputs such as ppfd or something more complex such as using [Pint](https://pint.readthedocs.io/en/stable/) to enforce unit consistency when writing a longer analysis script that uses multiple Pyrealm classes.

davidorme:
> Areas for improvement:
> 
> * I feel the constructor of the `PModelABC` base class is a bit cluttered and contains a lot of similar repeated operations when evaluating and retrieving different methods and models. There's potential to rationalise these operations into class methods and make them more testable. But this isn't a showstopper.

Yup - I agree that could be packaged more cleanly. I'll have a look at this but might park some of this for later polishing. I think the big picture shape of the new API and keeping the tests working is key on this PR.

> * Do we want to add explicit unit checking? This could be as simple as checking feasible physical values for inputs such as ppfd or something more complex such as using [Pint](https://pint.readthedocs.io/en/stable/) to enforce unit consistency when writing a longer analysis script that uses multiple Pyrealm classes.

On the bounds checking - yes, but not here. Once this is done, PPFD and FAPAR will move into PModelEnvironment, and that   adds bounds checks already (no issue yet, but see #394)

j-emberton:
Hi @davidorme , is this ready to go now?

davidorme:
@j-emberton Nope - this is going to sprawl a lot more, but there are a couple of incoming PRs that will make this switch cleaner.

j-emberton:
> @j-emberton Nope - this is going to sprawl a lot more, but there are a couple of incoming PRs that will make this switch cleaner.

ah cool. I thought I saw a refreshed request for review come in so thought I'd double check.

davidorme:
@j-emberton - Just to refresh on these two questions:

> * I feel the constructor of the `PModelABC` base class is a bit cluttered and contains a lot of similar repeated operations when evaluating and retrieving different methods and models. There's potential to rationalise these operations into class methods and make them more testable. But this isn't a showstopper.

I've done that - it doesn't really save a lot of lines but the structure and the separation of  attribute typing and population does look cleaner to me.

> * Do we want to add explicit unit checking? This could be as simple as checking feasible physical values for inputs such as ppfd or something more complex such as using [Pint](https://pint.readthedocs.io/en/stable/) to enforce unit consistency when writing a longer analysis script that uses multiple Pyrealm classes.

All the forcing variables inputs are now via `PModelEnvironment` (pending merge of #390) and are now checked via the new `BoundsChecker` class (merged from #409)

j-emberton (dismissed):
Thanks for the updates @davidorme . This looks good

j-emberton:
Sorry. I got ahead of myself. Have dismissed my approval

pyrealm/pmodel/__init__.py
```diff
@@ -46,6 +46,8 @@
     calculate_simple_arrhenius_factor,
 )
 from pyrealm.pmodel.isotopes import CalcCarbonIsotopes
+
+# from pyrealm.pmodel.new_pmodel import PModelNew, SubdailyPModelNew
```
MarionBWeinzierl, line 50: Rather delete than comment out

pyrealm/pmodel/__init__.py
```diff
@@ -62,6 +64,8 @@
     "PModel",
     "PModelEnvironment",
     "SubdailyPModel",
+    # "PModelNew",
```
MarionBWeinzierl, line 67: Same, rather delete

tests/regression/phenology/test_phenology_gpp.py
```diff
@@ -9,14 +9,18 @@
...
+    from pyrealm.pmodel.new_pmodel import PModelNew, SubdailyPModelNew
 
     # Calculate the PModel photosynthetic environment
     env = PModelEnvironment(
         tc=de_gri_half_hourly_data["TA_F"].to_numpy(),
         vpd=de_gri_half_hourly_data["VPD_F"].to_numpy(),
         co2=de_gri_half_hourly_data["CO2_F_MDS"].to_numpy(),
         patm=de_gri_half_hourly_data["PA_F"].to_numpy(),
+        fapar=np.ones(de_gri_half_hourly_data.shape[0]),
+        # ppfd=de_gri_half_hourly_data["PPFD"].to_numpy(),
```
MarionBWeinzierl, line 22: Can this line be deleted?

MarionBWeinzierl (commented):
I am not too keen on the "new" in the class names. Do we need to keep the old version?

Other than that, just a couple of tidying-up comments

davidorme:
> I am not too keen on the "new" in the class names. Do we need to keep the old version?

I'm replacing them side by side to make it easier to keep track of the two parallel versions. Once the new functionality in the parent branch is complete, I will rename these en-masse and strip the old implementations.

pyrealm/pmodel/__init__.py
davidorme: These are as a placeholder to remind me when I strip out the old implementations. Because this isn't merging into `develop` directly, there are a few loose ends which I'll tidy up before that branch merges down.

pyrealm/pmodel/__init__.py
davidorme: Here too.

tests/regression/phenology/test_phenology_gpp.py
davidorme: OK - here part of the problem is that I'm not sure which input variable should be used - this test isn't functional yet.

MarionBWeinzierl (approved):
Ok, if we see this as an "under-development" version I think it's ok.

Issue #383, opened 2025-01-13
[META] Release 2.0.0


davidorme:
This is a meta issue to keep track of the sub-issues required for the final release of version 2.0.0:

PR #403, opened 2025-01-28
Annual fAPAR_max and LAI_max

### LINKED ISSUE #348: Calculation of annual fAPAR and LAI values from a fitted PModel and precipitation data
This is the first step in implementing #347 and the description below is taken from the draft implementation document in the meta issue.

### Theory

The predicted fAPARmax / LAImax mainly has two roles:

1. is to calculate a parameter m which is the ratio of steady-state LAI to steady-state GPP; 
2. is to limit the prediction of LAI timeseries, predicted LAI ≤ LAImax.

The equations to calculate fAPARmax (Eqn 1) and LAImax (Eqn 2) are:

(Eqn. 1) $f_{APAR_{max}}= min\left[ (1 – z/(kA_0)), [c_a  (1–\chi)/1.6 D] [f_0  P/A_0]\right]$
(Eqn. 2) $LAI_{max}=  –(1/k) ln (1 –f_{APAR_{opt}})$

Of those terms, the following are new constants across the calculations. There aren't too many, so I think these are parameters to the calculation but we might want to bundle them into a new constants class (`MaxFAPARConst`?)

* $z$ accounts for the costs of building and maintaining leaves and the total below-ground allocation required to support the nutrient demand of those leaves, and the globally fitted values for z was 12.227 mol m–2 year–1. 
* $k$ is the light extinction coefficient (set at 0.5). 
* $f_0$ is the ratio of annual total transpiration of annual total precipitation, which is an empirical function (Eqn 3) of the climatic Aridity Index (AI).

(Eqn 3)  $f_0=0.65 e^{-b ln^2 \left(\frac{AI}{1.9}\right)  }$ 

where 0.65 is the maximum value, 1.9 is the AI at which this maximum occurs, and b = 0.604169. For AI , see below.

The following need to be be extracted from a P Model

* $A_0$ is the annual sum of potential GPP. Potential GPP would be achieved if fAPAR = 1. (mol m-2 year-1). 
* $c_a$ is the ambient CO2 partial pressure (Pa). 
* $\chi$ is the annual mean ratio of leaf-internal CO2 partial pressure (Pa) to ca during the growing season (>0℃).  
* $D$ is the annual mean vapour pressure deficit (VPD, Pa) during the growing season (>0℃).

And lastly, precipitation data and aridity data is needed:

* P is the annual total precipitation (mol   m–2 year–1),
*AI is a climatological estimate of local aridity index, typically calculated as total PET over total precipitation for an appropriate period, typically at least 20 years.

An Example to predict fAPARmax:
```
z is 12.227 mol m–2 year–1; 
k is 0.5; 
A0 is 200 mol m–2 year–1; 
ca is 40 Pa;
χ is 0.8; 
D is 3500 Pa; 
P is 51888 mol m–2 year–1; 
f0 is 0.62.
fAPAR_max=min⁡{1 –12.227/(0.5×200),[37×(1–0.8)/(1.6×3500)][0.62×51888/200]}  
                  = min {0.87, 0.21}
                  = 0.21
```
### Implementation

This is a draft workflow for implementing the calculation

- [ ] A user fits a P Model - this could be the standard or subdaily model. The model must cover at least one year and should be fitted with a $fAPAR$ fixed at 1 to give predictions of potential GPP ($A_0$ above). This could be fitted to a single site or to arrays for multiple sites.
- [ ] The user needs to provide a sequence of datetimes for each observation, which have to map onto one of the axes of the input data. For now, we follow the model used in the `SubdailyPModel` and `SPLASH` of insisting this is the first axis of the arrays.
- [ ] Get the growing seasons from the datetimes and temperatures (see #352).
- [ ] Using the PModel, the datetimes and the growing season, we can then extract annual values for annual mean $c_a, \chi, D$.
- [ ] We then need to take user provided precipitation to get annual precipitation.

So, I think the signature will look like this:

```python
class MaximumFAPAR:

    def __init__(
        self, 
        pmodel: PModel, 
        datetimes: NPArray[np.datetime64],
        growing_season: NPArray[np.bool_],
        precipitation: NPArray[np.datetime64],
        aridity_index: float,
        additional_constants: ...
    ):
        ...
```

The requirement for precipitation and aridity index ties this quite closely to the SPLASH module. The demonstration implementation should probably use SPLASH to set up the scenario.

MarionBWeinzierl:
# Description

Implements the computation of fAPAR_max and LAI_max, and provides examples usages of the API (including test data).

Fixes #348 .

## Type of change

## Key checklist

## Further checks


MarionBWeinzierl:
@davidorme , could you have a look and sense-check what's there? As you suggested in #407 (I hope I understood that right) I used `Scaler` directly, rather than extracting any functionality.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,208 @@
...
+        array)
+        datetimes: Datetimes of the measurements as np.datetime62 arrays.
+        growing_season: Bool array of days, indicating whether they are ain growing
+        season or not.
+        method: Either "total" (sum all values of the year) or "mean" (take the mean
+        of all values of the year)
+    """
+
+    # Extract years from datetimes
+    all_years = [np.datetime64(i, "Y") for i in datetimes]
```
davidorme:
```suggestion
    all_years = datetimes.astype("datetime64[Y]")
```

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,208 @@
+"""Class to compute the fAPAR_max and annual peak Leaf Area Index (LAI)."""
+
+import numpy as np
+from numpy.ma.core import zeros
```
davidorme: Why not just use `np.zeros`? I think `ma` is the masked array module and that has significant performance issues.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,208 @@
...
+        self,
+        annual_total_potential_gpp: NDArray[np.float64],
+        annual_mean_ca: NDArray[np.float64],
+        annual_mean_chi: NDArray[np.float64],
+        annual_mean_vpd: NDArray[np.float64],
+        annual_total_precip: NDArray[np.float64],
+        aridity_index: NDArray[np.float64],
+        z: float = 12.227,
+        k: float = 0.5,
+    ) -> None:
```
davidorme: We can bundle these constants (and the hardcoded ones below) into a `pyrealm.constants` class - probably a new one for `PhenologyConsts`?

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,208 @@
...
+        self.energylim = -9999 * np.ones(np.shape(fapar_waterlim))
+        self.annual_precip_molar = annual_total_precip
+
+        for i in range(len(fapar_waterlim)):
+            if fapar_waterlim[i] < fapar_energylim[i]:
+                self.faparmax[i] = fapar_waterlim[i]
+                self.energylim[i] = False
+            else:
+                self.faparmax[i] = fapar_energylim[i]
+                self.energylim[i] = True
```
davidorme:
```suggestion
        self.fapar_max = np.minimum(fapar_waterlim, fapar_energylim)
        """Maximum fapar given water or energy limitation."""
        self.energy_limited = fapar_energylim  < fapar_waterlim
        """Is fapar_max limited by water or energy."""
        self.annual_precip_molar = annual_total_precip
        """The annual precipitation in moles."""
```

E.g.

```python
In [8]: fapar_waterlim = np.arange(10)
In [10]: fapar_waterlim
Out[10]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
In [12]: fapar_energylim = np.arange(9, -1, -1)
In [13]: fapar_energylim
Out[13]: array([9, 8, 7, 6, 5, 4, 3, 2, 1, 0])

In [15]: np.minimum(fapar_waterlim, fapar_energylim)
Out[15]: array([0, 1, 2, 3, 4, 4, 3, 2, 1, 0])
In [16]: fapar_waterlim < fapar_energylim
Out[16]: array([ True,  True,  True,  True,  True, False, False, False, False,
       False])
```

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,208 @@
...
+            pmodel: pyrealm.pmodel.PModel
+            growing_season: Bool array indicating which times are within growing
+              season by some definition and implementation.
+            datetimes: Array of datetimes to consider.
+            precip: Precipitation for given datetimes.
+            aridity_index: Climatological estimate of local aridity index.
+        """
+
+        annual_total_potential_gpp = get_annual(
+            pmodel.gpp, datetimes, growing_season, "total"
```
davidorme: I think this need to check we are dealing with whole years in the input?

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,208 @@
...
+    if method == "total":
+        for i in range(len(years)):
+            annual_x[i] = sum(
+                daily_x[growing_season & (years_by_day == years[i])].astype(np.int64)
+            )
+    elif method == "mean":
+        for i in range(len(years)):
+            annual_x[i] = np.mean(
+                daily_x[growing_season & (years_by_day == years[i])].astype(np.int64)
+            )
```
davidorme: I think we can avoid this loop by reshaping the inputs? 

Oh no, we can't. Leap years. Hang on - does SubdailyScalar handle leap years? Must do? Need to check on this.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,208 @@
...
+        fapar_energylim = 1.0 - z / (k * annual_total_potential_gpp)
+        fapar_waterlim = (
+            f_0
+            * annual_total_precip
+            * annual_mean_ca
+            * (1 - annual_mean_chi)
+            / (1.6 * annual_mean_vpd * annual_total_potential_gpp)
+        )
+
+        self.faparmax = -9999 * np.ones(np.shape(fapar_waterlim))
```
davidorme: Generally, I'd use `np.nan` in cases like this. Here though,  I think we can make this more efficient - see suggestion below.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,208 @@
...
+              growing season (>0℃). [Pa]
+            annual_total_precip: Aka P, annual total precipitation. [mol m^{-2}
+             year^{-1}]
+            aridity_index: Aka AI, climatological estimate of local aridity index.
+            z: accounts for the costs of building and maintaining leaves and the total
+              below-ground allocation required to support the nutrient demand of those
+              leaves. [mol m^{-2} year^{-1}]
+            k: Light extinction coefficient.
+        """
+
```
davidorme, line 195: We should `_check_shapes` on the inputs. All should be equal (although aridity_index can be scalar, that use case is permitted by `_check_shape`).

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,208 @@
...
+    growing_season: NDArray[np.bool],
+    method: str,
+) -> NDArray:
+    """Computes an array of the annual total or mean of an entity x given datetimes.
+
+    Args:
+        x: Array of values to be converted to annual values. Should be either daily (
+        same datetimes as growing_season) or subdaily (same datetimes as datetimes
+        array)
+        datetimes: Datetimes of the measurements as np.datetime62 arrays.
```
davidorme:
```suggestion
        datetimes: Datetimes of the measurements as np.datetime64 arrays.
```

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,208 @@
...
+        method: Either "total" (sum all values of the year) or "mean" (take the mean
+        of all values of the year)
+    """
+
+    # Extract years from datetimes
+    all_years = [np.datetime64(i, "Y") for i in datetimes]
+
+    # Create scaler object to handle conversion between scales
+    scaler = SubdailyScaler(datetimes)
+    scaler.set_nearest(np.timedelta64(12, "h"))
```
davidorme: We can't use the `SubdailyScalar` on data that isn't sampled at subdaily frequency. The code explicitly checks for the spacing of observations within a day and breaks with coarser data (daily, monthly). I think we can still use the same approach if we nest this within the `if` statements below and solve it for the daily+ mode in another way.

davidorme: We should add a failure mode to SubdailyScalar for that case!

davidorme (changes requested):
Some optimisations and a couple of issues. I think we're on the right track here. I did wonder about making the class `__init__` accept daily/subdaily observations and moving the resampling in the `from_Pmodel` into there, but on balance I think it is right to have the straightforward usage as `__init__` and then use factory methods to handle other core inputs.

pyrealm/phenology/fapar_limitation.py
davidorme: I was mis-remembering this - SubdailyScalar reshapes evenly spaced (e.g.) hourly data into days by hours so is unaffected by leap years.

pyrealm/phenology/fapar_limitation.py
MarionBWeinzierl: That happens when you use an API that automatically adds includes, and forget the "np" part....

pyrealm/phenology/fapar_limitation.py
MarionBWeinzierl: OK, I am drawing z and k out into a constant class. I don't think f_0 should go in there, though, as it's a function of the aridity index. And b seems to be just an empirical constant? Does it have a meaning? Otherwise I'd keep it with f_0.

davidorme: All of those parameters with  the `f_0` calculation (`b`, 0.65 and 1.9) are empirical values setting the shape of the relationship and users might very well want to change them. I'd bundle them in the constants class as a tuple of three values as `f_0_coefficients` or similar.

pyrealm/phenology/fapar_limitation.py
MarionBWeinzierl: I opened issue #416 for this.

pyrealm/phenology/fapar_limitation.py
MarionBWeinzierl: OK

pyrealm/phenology/fapar_limitation.py
MarionBWeinzierl: Wondering whether this should be something similar as in `__init__` of scaler, or something much more simple (like checking that `growing_season` can be divided by 365 or 366).

pyrealm/phenology/fapar_limitation.py
MarionBWeinzierl: So we're leaving the loops for now?

pyrealm/constants/phenology_const.py
```diff
@@ -0,0 +1,26 @@
...
+       below-ground allocation required to support the nutrient demand of those
+       leaves. [mol m^{-2} year^{-1}]."""
+    z: float = 12.227
+
+    """Light extinction coefficient."""
+    k: float = 0.5
+
+    """Coefficients to calculate f_0, the ratio of annual total transpiration of annual 
+    total precipitation."""
+    f0_coefficients = (0.65, 0.604169, 1.9)
```
davidorme:
```suggestion
    z: float = 12.227
    """z accounts for the costs of building and maintaining leaves and the total
       below-ground allocation required to support the nutrient demand of those
       leaves. [mol m^{-2} year^{-1}]."""

    k: float = 0.5
    """Light extinction coefficient."""

    f0_coefficients = (0.65, 0.604169, 1.9)
    """Coefficients to calculate f_0, the ratio of annual total transpiration of annual 
    total precipitation."""
```

The standard for python attributes is docstring after declaration: https://peps.python.org/pep-0257/

pyrealm/phenology/fapar_limitation.py
davidorme: Yeah - this was tying me in knots as well. I think it needs to be something similar to the scaler init. We want to:

* Check that the timing is regular - although that is a pain when we want to support e.g. weekly or monthly inputs
* Check that there are the same number of observations per year - although that is a pain for leap years at daily precision.

davidorme (commented):
More thoughts - not fully formed!

pyrealm/phenology/fapar_limitation.py
davidorme: I think so - if it gives the right answers then we can always come back and optimise it.

pyrealm/phenology/fapar_limitation.py
MarionBWeinzierl: OK, I implemented a check for the datetimes. I am sure I have missed something...

Probably a good time to implement tests.

MarionBWeinzierl: It currently assumes daily or subdaily inputs

MarionBWeinzierl:
I have marked this as ready for review now, although there is still no integration test - I am trying to figure out whether there is something easier than the whole LAI example.

What do you think about merging in what is there and create a separate issue for the regression test @davidorme ?

MarionBWeinzierl:
I created issue #436 to implement regression tests.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,274 @@
...
+        self._check_shapes()
+
+        self.phenology_const = PhenologyConst()
+
+        #  f_0 is the ratio of annual total transpiration of annual total
+        #  precipitation, which is an empirical function of the climatic Aridity Index
+        #  (AI).
+        a = self.phenology_const.f0_coefficients[0]
+        b = self.phenology_const.f0_coefficients[1]
+        c = self.phenology_const.f0_coefficients[2]
```
davidorme: Python can unpack when assigning:

```suggestion
        a, b, c = self.phenology_const.f0_coefficients
```

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,274 @@
...
+        self.annual_total_potential_gpp = annual_total_potential_gpp
+        self.annual_mean_ca = annual_mean_ca
+        self.annual_mean_chi = annual_mean_chi
+        self.annual_mean_vpd = annual_mean_vpd
+        self.annual_total_precip = annual_total_precip
+        self.aridity_index = aridity_index
+
+        self._check_shapes()
+
+        self.phenology_const = PhenologyConst()
```
davidorme: These need docstrings - that is going to overlap with the `Args:` but I think we do need to document them as attributes of the object.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,274 @@
...
+    return precip_molar
+
+
+class FaparLimitation:
+    r"""FaparLimitation class to compute fAPAR_max and LAI.
+
+    This class takes the annual total potential GPP and precipitation, the annual
+    mean CA, Chi and VPD, as well as the aridity index and some constants to compute
+    the annual peak fractional absorbed photosynthetically active radiation (
+    fAPAR_max) and annual peak Leaf Area Index (LAI).
```
davidorme: It would be good to add a `.. TODO::` section in this docstring just to highlight to early users where the rough edges are:

* daily or subdaily timescales at present, not monthly etc
* daily conditions taken as noon values - might need to relax this.
* users currently need to provide own growing season definition in `from_pmodel`.

davidorme (changes requested):
@MarionBWeinzierl Some minor doc stuff and simple syntax changes. Then I think this is  great as something to build regression tests and further development around.

pyrealm/phenology/fapar_limitation.py
MarionBWeinzierl: Just to make sure: do you want those as `TODO:` or as `.. todo::` or as `.. todolist::` ?

davidorme: I meant as a `sphinx` `.. TODO::` directive (https://www.sphinx-doc.org/en/master/usage/extensions/todo.html) as it does nice styling.

Although - not sure I've configured that extension in this repo!

davidorme: @MarionBWeinzierl Are you happy for me to push a commit to this turning on that extension and styling it?

MarionBWeinzierl: Yes, of course. I had searched in the repo, but I couldn't find any other place where the sphinx directive was used, that's why I was unsure whether that's the right one to use here.

davidorme:
@MarionBWeinzierl I'm just wondering about what we want to do about `2.0.0` and this feature. My gut feeling is that we maintain this as a separate feature branch for now, merging #436 down into here for example, until after  the `2.0.0` meta branch is merged in #383. 

We don't want too many branches hanging around but it feels like it is worth keeping this partial implementation out of the `2.0.0` release candidates?

MarionBWeinzierl:
> @MarionBWeinzierl I'm just wondering about what we want to do about `2.0.0` and this feature. My gut feeling is that we maintain this as a separate feature branch for now, merging #436 down into here for example, until after the `2.0.0` meta branch is merged in #383.
> 
> We don't want too many branches hanging around but it feels like it is worth keeping this partial implementation out of the `2.0.0` release candidates?

Hm, I am not exactly enthusiastic about the idea, as it will mean more effort to make sure it does not diverge even more from the main implementation. We have merged in other not-quite-complete implementations, building things up incrementally. How is this part different?

davidorme:
I agree it is a change in how we've handled feature branches. I want to get all of the stuff in #383 down into `develop` soon and that will contain breaking changes for this feature. 

I guess I've also treated new features very much as a firehose of things that are immediately public, which  is one reason we've ended up with a big refactor for `2.0.0`. I guess I'd like `2.0.0` to be a checkpoint of what we already have to then build on?

MarionBWeinzierl:
Hm, GPP seems to be a bit dodgy in the class methods mentioned above. Looking at the annual test data and the test case for the faparlim constructor the annual total should be in the 200-300 area. However, when I am using the `from...`  class methods I am getting something like
```
array([ 746667.04952215,  815773.5100894 ,  743855.75473144,
       1001033.19042252,  855252.26212993,  881269.47573939,
        854679.73058753,  975973.12057888,  946219.67858962,
        834095.85222542, 1017449.87481279])
```
And without soil moisture correction it is
```
array([1348083.36464206, 1431984.61776091, 1437939.2231294 ,
       1734271.60568739, 1660070.42586098, 1675788.71195759,
       1480716.92616792, 1806653.76254044, 1771372.61237148,
       1474244.2391465 , 1767223.78820262])
```
For the soil moisture correction,  I am also not sure whether `daily_data["PMod_gpp_smstress"]`  or `daily_data["soilm_stress"]`  should be used -- but that does not make the crucial difference.

When I am testing the constructor directly, I am using `annual_data["ann_total_A0"]`  which seems fine.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,415 @@
...
+    - 1 mm/m2 = 1000000 mm3 = 1000 mL = 1 L
+    - Molar mass of water = 18g
+    - Assuming density = 1 (but really temp varying), molar volume = 18mL
+    - So 1 mm/m2 = 1000 / 18 ~ 55.55 mols/m2
+    """
+
+    water_mm_to_mol = 1000 / 18
+    precip_molar = precip_mm * water_mm_to_mol
+
+    return precip_molar
```
davidorme: I think this can go? There is now a temperature and pressure sensitive version in `core.water`

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,415 @@
...
+        """Annual mean ratio of leaf-internal CO2 partial pressure to c_a during the 
+        growing season (>0℃). [Pa]"""
+        self.annual_mean_vpd = annual_mean_vpd
+        """Annual mean vapour pressure deficit (VPD) during the growing season [Pa]"""
+        self.annual_total_precip = annual_total_precip
+        """Annual total precipitation. [mol m^{-2} year^{-1}]"""
+        self.aridity_index = aridity_index
+        """Climatological estimate of local aridity index."""
+
+        self._check_shapes()
```
davidorme, line 212: Can we import and use `core.utils.check_shapes` here instead?

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,415 @@
...
+            precip: Precipitation for given datetimes.
+            aridity_index: Climatological estimate of local aridity index.
+            gpp_penalty_factor: Penalty factor to be applied to pmodel.gpp
+        """
+
+        check_datetimes(datetimes)
+
+        annual_total_potential_gpp = get_annual(
+            pmodel.gpp * gpp_penalty_factor, datetimes, growing_season, "total"
+        )
```
davidorme: A few things here:
* GPP (and precipitation below) shouldn't use the growing season filtering - they use the actual totals across all observations.
* GPP in the PModel is reported in µg C m-2 s-1, but this model requires an annual estimate of mol C m-2 year-1.  I think the easiest approach here is to calculate the mean of the GPP values within the year and then scale it up by the number of seconds in the year and then divide by the molar mass of Carbon (which  will be in `PModel.core_const.k_c_molmass`.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,415 @@
...
+        annual_mean_ca = get_annual(pmodel.env.ca, datetimes, growing_season, "mean")
+        annual_mean_chi = get_annual(
+            pmodel.optchi.chi, datetimes, growing_season, "mean"
+        )
+        annual_mean_vpd = get_annual(pmodel.env.vpd, datetimes, growing_season, "mean")
+        annual_total_precip = get_annual(precip, datetimes, growing_season, "total")
```
davidorme: As noted above - no growing season filtering here.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,415 @@
...
+
+    @classmethod
+    def from_pmodel(
+        cls,
+        pmodel: PModel,
+        growing_season: NDArray[np.bool],
+        datetimes: NDArray[np.datetime64],
+        precip: NDArray[np.float64],
+        aridity_index: NDArray[np.float64],
+        gpp_penalty_factor: NDArray[np.float64],
```
davidorme: The penalty factor should be optional.

davidorme (changes requested):
I can see what the issue is with the GPP - I've checked the values for `ca`, `chi` and `vpd` and they are correct, but GPP and precipitation have a couple of issues. Note also that the fortnightly example data doesn't apply a soil moisture penalty, so we should leave that out of the test.

I think the code will be more efficient if we switch over to the `AnnualValueCalculator` class (since it calculate the indexing once), but at the moment that only handles the subdaily case so perhaps we get the tests passing with the current functions and then swap it over.

pyrealm/phenology/fapar_limitation.py
MarionBWeinzierl: Just to double-check, this should be totals, not mean values, for GPP, right?

MarionBWeinzierl:
After @davidorme implemented the `AnnualValueCalculator` #492 and this was incorporated in the example and the tests. The `from_pmodel` and `from_subdailypmodel` tests still fail as `lai` and `fapar` are off by a factor of 10. We might need to check the input data, but I think there might still be a fundamental problem with these two constructors.

davidorme:
I've checked the internally calculated values with  `from_pmodel` against the annual values from the reference implementation (`pyrealm_build_data/phenology/fortnightly_example/annual_data.csv`) and there are a couple of issues. 

1. The chi, ca and VPD calculations need to set `within_growing_season=True`. Once that is done, they match. 
2. The GPP needs rescaling - the PModel provides µgC m-2 -s-1 and needs to be mol C m2 for each observation. So that is `pmod.gpp * duration_in_seconds * 1e-6 / pmod.core_const.k_c_molmass`. Those values can then be fed into the AVC.

The tricky bit there is the duration of the observation, but AVC can provide that as an attribute. I've got those changes from testing - are you happy for me to push them?

MarionBWeinzierl:
> I've checked the internally calculated values with `from_pmodel` against the annual values from the reference implementation (`pyrealm_build_data/phenology/fortnightly_example/annual_data.csv`) and there are a couple of issues.
> 
>     1. The chi, ca and VPD calculations need to set `within_growing_season=True`. Once that is done, they match.
> 
>     2. The GPP needs rescaling - the PModel provides µgC m-2 -s-1 and needs to be mol C m2 for each observation. So that is `pmod.gpp * duration_in_seconds * 1e-6 / pmod.core_const.k_c_molmass`. Those values can then be fed into the AVC.
> 
> 
> The tricky bit there is the duration of the observation, but AVC can provide that as an attribute. I've got those changes from testing - are you happy for me to push them?

Yes, happy for you to push what you have.

davidorme:
Ugh. Found an unpleasant detail in the reference implementation. Using `xarray.Dataset.resample(time="2W")` uses a two week sampling window _but forcibly aligns to actual weeks using Monday to Sunday_. So because 2004-01-01 is a Thursday, the first observation is the resampled mean from only the first four days. I'll need to update that.

davidorme:
OK - I've updated `from_pmodel` and it passes. I think the logic should transfer across to `from_subdailypmodel`. Actually - looking at those two methods - I think `from_pmodel` works for both `PModel` and `SubdailyPModel`. They are now basically identical?

MarionBWeinzierl:
Ah, interesting. I always thought they were very similar. If they can be merged (or are already the same that would be great.

davidorme:
OK - that looks like it 99% works. Something is up with the first of the resulting fAPAR values - it passes with `rtol=1e-3` but without that first value it passes at `rtol=1e-6` as expected. So some minor glitch in there, but I think we can indeed merge those two methods.

MarionBWeinzierl:
> OK - that looks like it 99% works. Something is up with the first of the resulting fAPAR values - it passes with `rtol=1e-3` but without that first value it passes at `rtol=1e-6` as expected. So some minor glitch in there, but I think we can indeed merge those two methods.

OK, I will have a look at merging the two methods

MarionBWeinzierl:
> > OK - that looks like it 99% works. Something is up with the first of the resulting fAPAR values - it passes with `rtol=1e-3` but without that first value it passes at `rtol=1e-6` as expected. So some minor glitch in there, but I think we can indeed merge those two methods.
> 
> OK, I will have a look at merging the two methods

Actually, the I could just delete the subdaily one. there was just one "correction" in addition which made things worse again.

davidorme:
> Actually, the I could just delete the subdaily one.

I think so - we can make the one method accept both PModel and SubdailyPModel. I still haven't figured out why the first value is off - the rest of the sequence is correct!

davidorme:
OK  - the difference arises in the calculation of the first value of the total potential annual GPP. There is a difference in the sequence of calculation of the annual totals between the regression implementation and the class. One converts total GPP per timestep and sums. The other takes average values and then scales up. This _should_ be equivalent and seems to be after year 1, but for some reason isn't. My gut instinct is that the initial np.nan values are causing the problem - will try and dig into it because one approach is pr - 

OH OF COURSE. The mean value "fills" the gap from the np.nan values and the sum does not. Right. Will think what the correct approach is.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,305 @@
...
+    # year_remainder = len(unique_dates) % 365
+    # check that we have the right number of leap days
+    #    if year_remainder > no_leapdays:
+    #        raise ValueError("Datetimes do not cover full years.")
+
+    if obs_per_date > 1:
+        # subdaily
+
+        # Check that the number of observations per day is evenly divisible by the
+        # number of seconds in a day (should already have led to differing date counts).
```
sjavis:
```suggestion
        # Check that the number of seconds in a day is evenly divisible by the
        # number of observations per day (should already have led to differing date counts).
```

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,305 @@
...
+    return subdaily_x
+
+
+class FaparLimitation:
+    r"""FaparLimitation class to compute fAPAR_max and LAI.
+
+    This class takes the annual total potential GPP and precipitation, the annual mean
+    CA, Chi and VPD during the growing season, as well as the aridity index and some
+    constants to compute the annual peak fractional absorbed photosynthetically active
+    radiation ( fAPAR_max) and annual peak Leaf Area Index (LAI).
```
sjavis:
```suggestion
    radiation (fAPAR_max) and annual peak Leaf Area Index (LAI).
```

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,305 @@
...
+        annual_mean_ca = avc.get_annual_means(pmodel.env.ca, within_growing_season=True)
+        annual_mean_chi = avc.get_annual_means(
+            pmodel.optchi.chi, within_growing_season=True
+        )
+        annual_mean_vpd = avc.get_annual_means(
+            pmodel.env.vpd, within_growing_season=True
+        )
+
+        # Calculate total annual precipitation
+        annual_total_precip = avc.get_annual_totals(precip)
```
sjavis, line 306: Total GPP gets scaled up for missing values. Is this not needed for precipitation?

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,305 @@
...
+        self._check_shapes()
+
+        # Constants used for phenology computations
+        self.phenology_const = PhenologyConst()
+
+        #  f_0 is the ratio of annual total transpiration of annual total
+        #  precipitation, which is an empirical function of the climatic Aridity Index
+        #  (AI).
+        a, b, c = self.phenology_const.f0_coefficients
+        f_0 = a * np.exp(-b * np.log(aridity_index / c) ** 2)
```
sjavis, line 225: This will give a runtime warning if aridity_index is 0. Perhaps use np.where to set f_0 to zero if aridity_index is zero?

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,305 @@
...
+            * (1 - annual_mean_chi)
+            / (1.6 * annual_mean_vpd * annual_total_potential_gpp)
+        )
+
+        self.fapar_max = np.minimum(fapar_waterlim, fapar_energylim)
+        """Maximum fapar given water or energy limitation."""
+        self.energy_limited = fapar_energylim < fapar_waterlim
+        """Is fapar_max limited by water or energy."""
+        self.annual_precip_molar = annual_total_precip
+        """The annual precipitation in moles."""
```
sjavis:
```suggestion
        """Maximum fapar given water or energy limitation for each year."""
        self.energy_limited = fapar_energylim < fapar_waterlim
        """Is fapar_max limited by water or energy for each year."""
        self.annual_precip_molar = annual_total_precip
        """The annual precipitation in moles for each year."""
```
I think it isn't clear that these are annual results and not overall totals.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -0,0 +1,305 @@
...
+    years = np.unique(all_years)
+
+    # Compute annual totals or means, only taking into account the days which are in
+    # growing season.
+
+    annual_x = np.zeros(len(years))
+
+    if method == "total":
+        for i in range(len(years)):
+            annual_x[i] = sum(
```
sjavis:
```suggestion
            annual_x[i] = np.sum(
```

sjavis (commented):
I just have a few relatively minor comments.

pyrealm/phenology/fapar_limitation.py
MarionBWeinzierl: @davidorme Can the aridity index be zero? If not we might need to add an assertion.

davidorme: I don't think this can reasonably be zero - we should add a test and raise when it is not >0.

pyrealm/phenology/fapar_limitation.py
MarionBWeinzierl: I created #509 to track this. This can be implemented after this PR is merged.

pyrealm/phenology/fapar_limitation.py
MarionBWeinzierl: I added that check

davidorme (approved):
As discussed, some changes need to be made, but I'm happy to merge this as an experimental feature and then break out those new changes into smaller issues.

PR #404, opened 2025-01-28
Make `summarize_attrs` use non-mutable tuple as `attrs` argument

### LINKED ISSUE #402: Update args of the summarize method to accept a tuple of values not a list
In #401 , I'm adding class attributes to set the details of data attributes to be included in summarize. Currently, those are expected to be presented as a list, which is a mutable default. 

We could fix that by using `typing.ClassVar` in the classes, but really the decision to make it a list in the first place was the mistake, so we should get rid of that.

https://github.com/ImperialCollegeLondon/pyrealm/blob/df515aa2573d8327269a7434ce15b410ed40fb18/pyrealm/core/utilities.py#L113-L116

davidorme:
# Description

This PR updates the `attrs` argument to `pyrealm.core.utilities.summarize_attrs`. It used to accept values as lists of attribute names, either with or without units:

```python
["attr_name_1", "attr_name_2"]
[("attr_name_1", "m2"), ("attr_name_2", "kg")]
```
But that mutable list argument makes it awkward to use with e.g. class attributes etc. So it is accepts a tuple of values and the option to just specify names has been removed as it was never used. 

```python
(("attr_name_1", "m2"), ("attr_name_2", "kg"))
```

All the use cases are now updated to match.

Fixes #402

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/core/utilities.py
+++ b/pyrealm/core/utilities.py
@@ -111,7 +111,10 @@ def check_input_shapes(*args: float | int | np.generic | np.ndarray | None) -> t
 
 
 def summarize_attrs(
-    obj: object, attrs: list, dp: int = 2, repr_head: bool = True
+    obj: object,
+    attrs: tuple[tuple[str, str], ...],
+    dp: int = 2,
+    repr_head: bool = True,
 ) -> None:
     """Print a summary table of object attributes.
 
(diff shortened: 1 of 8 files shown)


MarionBWeinzierl (approved):
Looks good

PR #407, opened 2025-01-31
Move scaler.py to core

### LINKED ISSUE #406: Move daily means functions etc to core
As  discussed in #405 , [get_window_values](https://github.com/ImperialCollegeLondon/pyrealm/blob/88ed9a14b18f0e4ddf3302baa88ec7164c479a86/pyrealm/pmodel/scaler.py#L332) and [ get_daily_means](https://github.com/ImperialCollegeLondon/pyrealm/blob/88ed9a14b18f0e4ddf3302baa88ec7164c479a86/pyrealm/pmodel/scaler.py#L370) should be taken out of `scaler.py` and made available as core functions.

MarionBWeinzierl:
# Description

Instead of taking just the daily means functions out of `scaler`, this PR moves the whole class to `core`, to make it functionality available elsewhere in the code.

Fixes #406

## Type of change

## Key checklist


davidorme:
Hmmm. Not sure how I feel about this.

That's partly because this is going to change with the move to an `AcclimationModel` API (see #394), but also I think the correct way to do this is probably to separate out the core of the wrapping and subsetting engine in here from the SubdailyPModel specific components. That's more of a job, but I guess I think that rather than moving the very SubdailyPModel specific implementation into `core` now  it would be better to import it from its current location for the meantime. Then the usage can help guide the breaking out the underlying shared functionality into `core`?

PR #409, opened 2025-02-03
New BoundsChecker implementation

### LINKED ISSUE #408: Replace bounds checker functionality with something more adaptable
At the moment, `bound_checker` is used within code (PModel, Splash) to check input variables are within bounds. However, the bounds themselves are hardcoded and the bounds checking is restricted to named arguments. With the proposed API changes in #394, `PModelEnvironment` loses the ability to check bounds on additional variables.

Making the bounds checker into a customizable class with sensible defaults means that the bounds are no longer hard-coded and can be adapted by users to new inputs.

davidorme:
# Description

This PR replaces `core.utilities.bounds_checker` with a user-modifiable and extendable `BoundsChecker` class. This does add one more argument to some class signatures but we plan to reorganise signatures (#394). 

The main advantages are that:
* It removes hard-coded bounds and allows users to modify them
* It can be simply extended to add bounds checking to new optional in e.g. `PModelEnvironment`

The implementation:
* Defines a `Bounds` dataclass used to define bounds for a particular variable.
* Defines a `BoundChecker` class with default bounds for core variables that acts like a library for bounds checking.
* The main use case is e.g. `BoundsChecker().check("tc", np.array([10, 1000])`, which will check that these alleged temperature data fall within the configured bounds.

Fixes #408

## Type of change

## Key checklist

## Further checks


MarionBWeinzierl (commented):
This looks, in principle, good. I am just starting to wonder, if everything is made a class, won't there, at some point, be a significant runtime penalty? But maybe it's an outdated clichee (or was never true in the first place for sequential codes) that OOP makes things slower, runtime-wise.

davidorme:
I worry about that a bit. This PR is a small class that is basically just used to apply some numpy functions to potentially large arrays. I doubt it is significantly slower than the original, if at all. I think there is a minor cost to accessing by attribute, but it's a relatively small number of attributes that are accessed once per variable in the whole of an analysis.

MarionBWeinzierl:
I agree that this one instance will will not make a bit difference, it was more a general observation about the way the code develops.

MarionBWeinzierl (approved):

pyrealm/core/bounds.py
```diff
@@ -0,0 +1,169 @@
...
+
+    The class provides a library of  {class}`~pyrealm.core.bounds.Bounds` instances for
+    core variables, keyed by the
+    {attr}`Bounds.var_name<pyrealm.core.bounds.Bounds.var_name>` attribute. The table is
+    populated from default values when a ``BoundsChecker`` instance is created but can
+    be updated and extended by assigning new ``Bounds`` instances to existing or new
+    variable name keys using the ``update`` method.
+    """
+
+    # TODO - think about these argument names - some unnecessarily terse.
```
j-emberton, line 70: What plan do you have for renaming these vars?

davidorme: Good question - I think once the API structural changes have been pushed through, then review the function signatures for variable name and synchronise them across the code base. So a polishing step once we've put it all back together.

j-emberton (approved):
Looks good

PR #412, opened 2025-02-04
New PModelEnvironment implementation

### LINKED ISSUE #390: Centralise data inputs to PModel in PModelEnvironment

davidorme:
# Description

This PR updates the definition of `PModelEnvironment` to the proposed structure in the version 2.0.0 sketch in #394 

* The fAPAR and PPFD forcing variables are now included in the list of arguments,
* All of the other directly defined arguments for optional variables are removed (theta, rootzonestress, mean_growth_temperature, aridity_index) and instead `__init__` treats all `kwargs` variables as potential additional variables.
* The class uses the new `BoundsChecker` to validate inputs.
* The changes add `tk` as an internal attribute.

It also updates all of the test and doc usages of `PModelEnvironment`. It *does not* remove usage of fAPAR and PPFD outside of `PModelEnvironment` - these will become redundant when #401 is merged, addressing #386 and #385, so I'll tidy up on that branch.

> [!TIP]
> There are a _lot_ of files changed because of all the test and docstring usage of this core class, but the main thing to review is the new `pmodel_environment.py` file.

> [!IMPORTANT]
> Having had to update mutiple test and doc files to set `fapar=1, ppfd=1`, I think it makes sense to have these two variables default to one.
> 
> 90% of the time, users will want to set them, but some use cases don't need `fapar` and `ppfd` and it is annoying to _have_ to do so. Can you see any downsides to this tweak?

Fixes #390 

## Type of change

## Key checklist

## Further checks


pyrealm/pmodel/pmodel_environment.py
```diff
@@ -124,6 +149,12 @@ def __init__(
                 "zero or explicitly set to np.nan"
             )
 
+        # Internal calculations
+        self.tk = self.tc + self.core_const.k_CtoK
+        """The temperature at which to estimate photosynthesis in Kelvin (K)"""
+
+        # TODO - swap functions over to using tk as needed
```
MarionBWeinzierl, line 165: Does this need to be a Github issue?

pyrealm/pmodel/quantum_yield.py
```diff
@@ -321,11 +323,14 @@ def _calculate_kphio(self) -> None:
...
 
+        aridity_index = getattr(self.env, "aridity_index")
+        mean_growth_temperature = getattr(self.env, "mean_growth_temperature")
+
         # Calculate enzyme kinetics
         a_ent, b_ent, Hd_base, Ha = self.env.pmodel_const.sandoval_kinetics
         # Calculate activation entropy as a linear function of
         # mean growth temperature, J/mol/K
-        deltaS = a_ent + b_ent * self.env.mean_growth_temperature
+        deltaS = a_ent + b_ent * mean_growth_temperature
```
MarionBWeinzierl: You are not usually using camel case - might be more consistent to use snake case (`delta_s`)

MarionBWeinzierl (commented):
Two small suggestions

pyrealm/pmodel/pmodel_environment.py
davidorme: Added as #413

MarionBWeinzierl (approved):

pyrealm/pmodel/pmodel_environment.py
```diff
@@ -70,6 +72,13 @@ class PModelEnvironment:
       growing degree days (°C), also used in
       :meth:`~pyrealm.pmodel.quantum_yield.QuantumYieldSandoval`.
 
+    .. note::
```
MarionBWeinzierl, line 75: Should there also be a warning output, or is it enough to be here in the docs?

davidorme: I think a warning would be intrusive. I'll need to revise the user docs anyway so will also add a similar note on those pages.

j-emberton (approved):

PR #418, opened 2025-02-11
Develop the AcclimationModel class

### LINKED ISSUE #415: Develop the new AcclimationModel class
This issue is to restructure the inputs to the SubdailyPModel, so that all of the arguments to do with the acclimation window can be  logically linked within a new `AcclimationModel` class, that is largely an extension of the existing `SubdailyScaler`.

The target signature, as suggested in #394 is something like:

```python
class AcclimationModel:

    def __init__(
        self,
        datetimes: NDArray[np.datetime64],
        alpha: float = 1 / 15,
        allow_holdover: bool = False,
        allow_partial_data: bool = False,
        fill_kind: str = "previous",
        previous_realised: tuple[NDArray, NDArray, NDArray] | None = None,
    )
```

davidorme:
# Description

Another major piece of surgery to the PModel implementation to  add the new `AcclimationModel` class to the 2.0.0 branch. The PR:

* adds the `AcclimationModel` in `pyrealm/pmodel/acclimation.py`. This is very similar to the `SubdailyScaler`, but moves most of the methods arguments to instance attributes so that the model can be defined once and then used consistently and with shorter signatures.
* moves the `pyrealm.pmodel.subdaily.memory_effect` to `pyrealm.core.utilities.calculate_exponential_weighted_average` for wider use and then wraps it in `AcclimationModel.apply_acclimation` for this specific use case.
* adds a new test file for the model in `tests/unit/pmodel/test_acclimation_model.py`

One thing I haven't done is to include the `previous_values/initial_values` in this. These are arrays that need to match to dimensions of the `PModelEnvironment` data of the `SubdailyPModel`, so there is no real way to validate the inputs within this - which only knows about the date axis of that environment. It makes more sense then to keep that argument in the  `SubdailyPModel` signature.

> [!important]
> This is another partial contribution to the 2.0.0 branch. Things left to do after merging:
> * I've again left the original implementations in place. 
> * I haven't updated `SubdailyPModelNew` to use this yet. 
> * I haven't checked the docs build.
> 
> My main aim here is to get a review of the new proposed structure. 

Fixes #415 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,7 +18,8 @@ worked through. The changes below are provisional.
   **Breaking change**: Need to specify `fapar` and `ppfd` in `PModelEnvironment`.
 
 - The `bounds_checker` function has been retired and replaced with the `BoundsChecker`
-  class, which provides more flexible and user-configurable bounds checking.
+  class, which provides more flexible and user-configurable bounds checking. This
+  functionality is used within other classes and does not introduce breaking changes.
 
 - A new system for providing alternative calculations of quantum yield ($\phi_0$) in the
   P Model, using the new `pyrealm.pmodel.quantum_yield` module. This module now provides
@@ -53,6 +54,21 @@ worked through. The changes below are provisional.
   settings, the reported values for $V_{cmax25}$ and $J_{max25}$ using `PModel` will
   change between v1 and v2, with the shift from `kattge_knorr` to `simple` as the
   default factors.
+
+- Many of the arguments to `SubdailyPModel` have been brought together into a new
+  `AcclimationModel` class. This replaces `SubdailyScaler` and bundles all of the
+  settings for acclimation into a single class. The following **breaking change**:
+  
+  - `SubdailyScaler` has been replaced with `AcclimationModel`.
+
+- The legacy implementation `SubdailyPModel_JAMES` has been deprecated. This
+  implementation duplicated the original Mengoli et al JAMES code. This was largely a
+  proof of concept implementation, misses some key parts of the acclimation model and
+  the internal calculations are sufficiently different that there is a high maintenance
+  cost to updating it the new API in version 2.0.0.
+
+  - The `fill_from` argument to `fill_daily_to_subdaily` was only required for
+    `SubdailyPModel_JAMES` and so this has also been deprecated.
 
 - The functions `calc_ftemp_kphio` and `calc_ftemp_inst_vcmax` provided narrow use cases
   with code duplication. They have been replaced by two broader Arrhenius functions:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,15 +57,19 @@ worked through. The changes below are provisional.
 
 - Many of the arguments to `SubdailyPModel` have been brought together into a new
   `AcclimationModel` class. This replaces `SubdailyScaler` and bundles all of the
-  settings for acclimation into a single class. The following **breaking change**:
+  settings for acclimation into a single class. The following is therefore a **breaking
+  change**:
   
-  - `SubdailyScaler` has been replaced with `AcclimationModel`.
+  - `SubdailyScaler` has been replaced with `AcclimationModel`, and the following
+    arguments to `SubdailyPModel` are now arguments to `AcclimationModel`: `alpha`,
+    `allow_holdover`, `allow_partial_data`, `update_point`, `fill_kind` (as
+    `fill_method`).
 
 - The legacy implementation `SubdailyPModel_JAMES` has been deprecated. This
   implementation duplicated the original Mengoli et al JAMES code. This was largely a
   proof of concept implementation, misses some key parts of the acclimation model and
   the internal calculations are sufficiently different that there is a high maintenance
-  cost to updating it the new API in version 2.0.0.
+  cost to updating it to the new API in version 2.0.0.
 
   - The `fill_from` argument to `fill_daily_to_subdaily` was only required for
     `SubdailyPModel_JAMES` and so this has also been deprecated.
--- a/pyrealm/pmodel/acclimation.py
+++ b/pyrealm/pmodel/acclimation.py
@@ -518,7 +518,7 @@ class AcclimationModel:
     def fill_daily_to_subdaily(
         self,
         values: NDArray[np.float64],
-        initial_values: NDArray[np.float64],
+        previous_values: NDArray[np.float64] | None = None,
     ) -> NDArray[np.float64]:
         """Resample daily variables onto the subdaily time scale.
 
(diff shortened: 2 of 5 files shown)


CHANGES.md
```diff
@@ -54,6 +55,21 @@ worked through. The changes below are provisional.
   change between v1 and v2, with the shift from `kattge_knorr` to `simple` as the
   default factors.
 
+- Many of the arguments to `SubdailyPModel` have been brought together into a new
+  `AcclimationModel` class. This replaces `SubdailyScaler` and bundles all of the
+  settings for acclimation into a single class. The following **breaking change**:
```
MarionBWeinzierl: Should the last sentence say something like "the following is a breaking change" or "the following breaking change was introduced" or so? It's not a sentence like that.

CHANGES.md
```diff
@@ -54,6 +55,21 @@ worked through. The changes below are provisional.
...
+  
+  - `SubdailyScaler` has been replaced with `AcclimationModel`.
+
+- The legacy implementation `SubdailyPModel_JAMES` has been deprecated. This
+  implementation duplicated the original Mengoli et al JAMES code. This was largely a
+  proof of concept implementation, misses some key parts of the acclimation model and
+  the internal calculations are sufficiently different that there is a high maintenance
+  cost to updating it the new API in version 2.0.0.
```
MarionBWeinzierl:
```suggestion
  cost to updating it to the new API in version 2.0.0.
```

pyrealm/pmodel/acclimation.py
```diff
@@ -0,0 +1,663 @@
+"""The :mod:`~pyrealm.pmodel.acclimation` module provides the
+:class:`~pyrealm.pmodel.acclimation.AcclimationModel` class, which is a required input
+to the :class:`~pyrealm.pmodel.new_pmodel.SubdailyPModelNew` class for fitting the P
+Model at subdaily time scales.
+"""  # noqa: D205
+
+import numpy as np
+from numpy.typing import NDArray
+from scipy.interpolate import interp1d  # type: ignore
```
MarionBWeinzierl, line 9: Do we still want to use the deprecated interp1d, or look into using a newer alternative?

pyrealm/pmodel/acclimation.py
```diff
@@ -0,0 +1,663 @@
...
+    :class:`~pyrealm.pmodel.new_pmodel.SubdailyPModelNew`. These datetimes need to be:
+
+    * sampled at a subdaily frequency,
+    * be strictly increasing,
+    * be evenly spaced, using a spacing that evenly divides a day.
+
+    The model can handle incomplete days at the start and end of the time series and
+    will internally pad the datetimes to complete days in order to ensure correct
+    sampling. The values in ``datetimes`` are assumed to be the precise times of the
+    observations. If the datetimes are represent the start or end of a sampling time
```
MarionBWeinzierl:
```suggestion
    observations. If the datetimes represent the start or end of a sampling time
```

pyrealm/pmodel/acclimation.py
```diff
@@ -0,0 +1,663 @@
...
+    span, then they should first be converted to a reasonable choice of observation
+    time, such as the midpoint of the timespan.
+
+    An acclimation window must then be set, defining a period of the day representing
+    the environmental conditions that plants will acclimate to. This will typically by
```
MarionBWeinzierl:
```suggestion
    the environmental conditions that plants will acclimate to. This will typically be
```

pyrealm/pmodel/acclimation.py
```diff
@@ -0,0 +1,663 @@
...
+        allow_partial_data: See
+            :meth:`~pyrealm.pmodel.acclimation.AcclimationModel.get_daily_means`
+        alpha: See
+            :meth:`~pyrealm.pmodel.acclimation.AcclimationModel.apply_acclimation`
+        allow_holdover: See
+            :meth:`~pyrealm.pmodel.acclimation.AcclimationModel.apply_acclimation`
+        update_point: See
+            :meth:`~pyrealm.pmodel.acclimation.AcclimationModel.fill_daily_to_subdaily`
+        fill_method: See
+            :meth:`~pyrealm.pmodel.acclimation.AcclimationModel.fill_daily_to_subdaily`
```
MarionBWeinzierl: I don't like the "see xxx" way here, in this case I'd rather duplicate the description. Once because then I don't need to scroll around, and second as the signature of the other method might change but this one might stay as it is, and then you loose the description.

pyrealm/pmodel/acclimation.py
```diff
@@ -0,0 +1,663 @@
...
+        self.method = "Include array"
+        self._set_sampling_times()
+
+    def set_nearest(self, time: np.timedelta64) -> None:
+        """Set the acclimation conditions to a single value closest to a target time.
+
+        This method finds the daily observation time closest to a value provided as a
+        :class:`~numpy.timedelta64` value since midnight. If the provided time is
+        exactly between two observation times, the earlier observation will be used.
+        The resulting single observation will then as the daily sample.
```
MarionBWeinzierl:
```suggestion
        The resulting single observation will then be used as the daily sample.
```

pyrealm/pmodel/acclimation.py
```diff
@@ -0,0 +1,663 @@
...
+        series used to create the instance.
+
+        Two attributes of the ``AcclimationModel`` class are used to tune the behaviour
+        of this method:
+
+        * The ``update_point`` attribute sets the point on the subdaily scale at which
+          the plant updates acclimating variables to the new daily realised value. The
+          default is ``max`` - the plant starts to acclimate to a new value at the end
+          of the acclimation window - but this can also be set to ``mean`` to start
+          acclimation from the middle of the acclimation  window. Note that this setting
```
MarionBWeinzierl:
```suggestion
          acclimation from the middle of the acclimation window. Note that this setting
```

tests/unit/core/test_exponential_moving_average.py
```diff
@@ -0,0 +1,142 @@
...
+            + np.arange(16).reshape(4, 4),
+            id="3D",
+        ),
+    ],
+)
+@pytest.mark.parametrize(argnames="alpha", argvalues=(0.0, 0.2, 0.4, 0.6, 0.8, 1.0))
+def test_exponential_moving_average(inputs, alpha):
+    """Test the exponential_moving_average.
+
+    This uses matrix maths to calculate expected by calculating the coefficients of the
```
MarionBWeinzierl:
```suggestion
    This uses matrix maths to calculate expected values by calculating the coefficients of the
```

MarionBWeinzierl (changes requested):
Just being a bit nit-picky here

pyrealm/pmodel/acclimation.py
davidorme: Oh good catch. I'll break this out into it's own issue.

pyrealm/pmodel/acclimation.py
davidorme: Fair. These are a little odd because they _aren't_ arguments to those other functions, they are internal settings that affect specific methods. But I see your point about the explanation. I will also leave the `see xxx` so that the internal usage is clear.

davidorme:
@MarionBWeinzierl Both nitpicky and excellent 😄 I've fixed them all - I've left the comment about the `see xxx` arguments unresolved. Do my changes to that seem better?

MarionBWeinzierl (approved):

PR #430, opened 2025-02-18
Prototype of @experimental

### LINKED ISSUE #429: Look at `experimental` decorator.
I would prefer to review and explicitly mark all experimental features in 2.0.0. It would be easier to do that if we had a decorator to add to classes, methods and functions.

The decorator should:

* generate a warning when the feature is used, and
* extend the docstrings to add an admonition.

[ETA]

This turns out to be hard to do well - handling the decoration needs to preserve a lot of information about the decorated objects in order to avoid autodoc losing information. 

* The docstring extension is reasonable - this can be handled by the decorator or by adding an attribute to the decorated object that triggers a sphinx extension to add text. I don't think there's much to call here, except that handling it in the decorator seems more compact than having to add sphinx extensions.

* Preserving the signature is much trickier. The `Deprecated` package comes _very_ close and provides a framework we can customise to provide an `experimental` decorator via the `deprecated.sphinx` module, but at the moment the `autodoc` output replaces the actual class `__init__`  signatures with `(*args, **kwargs)`, which won't do.

There's an open issue on that - if this gets fixed then I think we can implement this:

https://github.com/laurent-laporte-pro/deprecated/issues/86

There are various partial experiments in the branch.

davidorme:
# Description

This PR:

* Introduces the new `pyrealm.core.experimental` module, which  provides the `@experimental` decorator and `ExperimentalFeatureWarning2` (so called because we already have `pyrealm.ExperimentalFeatureWarning`, although this would be replaced if we get this working.

Tests show it works, but there is an issue with building the docs - somehow this interferes with `autodoc` in a way that breaks the links to experimental objects - if you ignore errors when building, the build succeeds and it looks good but the link in the side menu is broken and the links to the object fail:

See here:

https://pyrealm.readthedocs.io/en/429-experimental-decorator/api/pmodel_api.html#module-pyrealm.pmodel.quantum_yield

The `QuantumYieldSandoval()` appears differently in the menu and the attributes are not autogenerated.

Fixes #429

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/core/experimental.py
+++ b/pyrealm/core/experimental.py
@@ -1,0 +1,87 @@
+"""This module provides a simple decorator for use with experimental features of the
+pyrealm codebase. It automatically adds an experimental admonition to the docstring of
+decorated objects and causes those function to generate an experimental warning when
+called. The structure is largely taken from  the `deprecated` package, released under
+the Apache 2.0 licence.
+"""  # noqa: D205
+
+# Original licence of deprecation package:
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import functools
+import textwrap
+import warnings
+from collections.abc import Callable
+from typing import Any
+
+
+class ExperimentalFeatureWarning2(Warning):
+    """Warn about experimental features.
+
+    This is just a simple wrapper on the base Warning to issue clearer warnings about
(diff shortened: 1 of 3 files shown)


davidorme:
Ohhhh. It's also listed as a **function** at the top of the page. Something bad is happening.

It may be that there is a better `sphinx` side way to do this as @j-emberton suggested. Might be worth implementing both approaches in this PR and comparing side by side.

MarionBWeinzierl:
> Ohhhh. It's also listed as a **function** at the top of the page. Something bad is happening.
> 
> It may be that there is a better `sphinx` side way to do this as @j-emberton suggested. Might be worth implementing both approaches in this PR and comparing side by side.

Do you mean this? https://deprecated.readthedocs.io/en/latest/sphinx_deco.html

davidorme:
I've basically followed the internals out of:

https://github.com/briancurtin/deprecation/blob/master/deprecation.py
https://pypi.org/project/deprecation/

The actual docstring manipulation works but something interferes with the autodoc handling. I'll have a look at the internals of that Deprecated package and see if it is different

davidorme:
This turns out to be hard to do well. See notes on issue #429. This PR may be re-opened in future.

PR #432, opened 2025-02-19
Update default values for phi0

### LINKED ISSUE #414: Fix the default `kphio` values
At present, the quantum yield calculations use Beni's default settings for his calibrated calculations with the soil moisture correction:

https://github.com/ImperialCollegeLondon/pyrealm/blob/247d35e90a35c915020e2faa07ab7fce5f98384c/pyrealm/pmodel/quantum_yield.py#L220

https://github.com/ImperialCollegeLondon/pyrealm/blob/247d35e90a35c915020e2faa07ab7fce5f98384c/pyrealm/pmodel/quantum_yield.py#L243

We need to set these to 1/8 and then update the soil moisture correction docs to explicitly say that you need to set those values to replicate the predictions of the Stocker calculations.

davidorme:
# Description

The different `QuantumYieldABC` implementations had different default values for `phi_0` - but these choices reflected particular analysis decisions of the original implementations. There is now a single default.

This PR:

* removes the `default_reference_kphio` class attribute from `QuantumYieldABC` which provided class specific defaults and replaces it with a shared global default value set in the `PModelEnvironment.pmodel_consts`. 
* That constant value can be overridden globally through the constants settings, but also particular models can be run with a different setting by using `reference_kphio` with `PModel` or `SubdailyPModel` as before - this is passed into the `QuantumYieldABC` to override the global default.
* The soil moisture functions are most affected here.  The core `QuantumYield` classes  functions were explicitly parameterised with `phi_0` settings to match these soil moisture functions. That lead to confusing differences in predictions - which is the motivation for this change - and so the docs for these now need state explicitly that the models need to be set up correctly to duplicate those analyses.

Fixes #414

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -50,9 +50,18 @@ worked through. The changes below are provisional.
   supports the previous fixed and temperature dependent $\phi_0$ approaches but also
   David Sandoval's extension for estimating the impact of water stress on $\phi_0$.
 
-  **Breaking change** The signatures of the `PModel` and `SubdailyPModel` classes have
-  changed: the arguments `kphio` and `do_ftemp_kphio` have been replaced by
-  `method_kphio` and `reference_kphio`.
+  **Breaking changes**:
+  
+  - The signatures of the `PModel` and `SubdailyPModel` classes have
+    changed: the arguments `kphio` and `do_ftemp_kphio` have been replaced by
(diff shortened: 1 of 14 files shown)


MarionBWeinzierl (approved):
Looks much cleaner this way.

PR #435, opened 2025-02-19
Rework PModelConsts and usage in functions

### LINKED ISSUE #391: Review the constants bundling in `pyrealm.constants`
Many of the constants in `pyrealm.constants` subclasses are presented as single float values, even when they are logical groups of coefficients from a single source. They would be more clearly documented and organised as tuples or dicts of values. For example:

https://github.com/ImperialCollegeLondon/pyrealm/blob/531447d76dbd73ceb2608aa55b86914c0ae9aa75/pyrealm/constants/pmodel_const.py#L185-L193

I think this would be better as:

```python
# Mengoli 2023 soil moisture stress
soilm_mengoli: dict[str, float] = field(
        default_factory=lambda: dict(y_a=0.62, y_b=-0.45, psi_a=0.34, psi_b=-0.60)
)
"""Coefficients of the Mengoil soil moisture penalty function for GPP.

The values are the coefficient ($y_a$) and exponent ($y_b$) of the maximal level function and 
the coefficient ($\psi_a$) and exponent $\psi_b$) of the threshold function.
"""
```

If we go down this route, it would be nice if we could use dot notation here - maybe moving to `dotmap` or `python-box` (aka Box) similar for these constant classes would be worthwhile? We could continue to use frozen dataclasses, but then we have to define subclasses for all the nested components, and this feels more like something where we want to define a Marshmallow validator and then expose in a more lightweight way?

Alternatively, this is simpler but you need to get the order right when using the values:

```python
# Mengoli 2023 soil moisture stress
soilm_mengoli: tuple[float, ...] = (0.62, -0.45, 0.34, -0.60)
"""Coefficients of the Mengoil soil moisture penalty function for GPP.

The values are the coefficient ($y_a$) and exponent ($y_b$) of the maximal level function and 
the coefficient ($\psi_a$) and exponent $\psi_b$) of the threshold function and are given in that order.
"""
```

davidorme:
# Description

The `PModelConsts` object has a very disorganised namespace with a mix of bundled and individual parameters, and poor use of attribute docstrings to pair information with data. 

> [!TIP]
> This is a slightly sprawling PR because it restructures PModelConst - so lots of changes - but also updates the signatures of the `pmodel` functions. Most of the actual changes are in `pmodel_const.py` and `pmodel/functions.py` - everything else is just updates to match

* Reviews the `PModelConsts` contents to group closely related parameters into single attributes with better documentation.
* Adds `tk_ref` to avoid the omnipresent recalculation of this value through the code base, and renames `plant_T_ref` to `tc_ref` for consistency.
* Updates the signatures of the `pyrealm.pmodel.functions` to take more specific parameterisation, rather than simply chucking all of `PModelConst` and `CoreConst` into each one. This makes the implementation and docs clearer.
* Updates usage to match and updates tests.

Fixes #391

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/users/pmodel/pmodel_details/soil_moisture.md
+++ b/docs/source/users/pmodel/pmodel_details/soil_moisture.md
@@ -106,13 +106,14 @@ from pyrealm.pmodel.pmodel_environment import PModelEnvironment
 from pyrealm.constants import PModelConst
 
 # change default theta0 parameter
-const = PModelConst(soilmstress_theta0=0.1)
+coef = PModelConst().soilmstress_stocker
+coef["theta0"] = 0.1
 
 # Calculate q
 mean_alpha_seq = np.linspace(0, 1, 101)
 
-q = (1 - (const.soilmstress_a + const.soilmstress_b * mean_alpha_seq)) / (
-    const.soilmstress_thetastar - const.soilmstress_theta0
+q = (1 - (coef["a"] + coef["b"] * mean_alpha_seq)) / (
+    coef["thetastar"] - coef["theta0"]
 ) ** 2
 
 # Create a 1x2 plot
@@ -129,16 +130,16 @@ soilm = np.linspace(0, 0.7, 101)
 for mean_alpha in [0.9, 0.5, 0.3, 0.1, 0.0]:
 
     soilmstress = pmodel.calc_soilmstress_stocker(
-        soilm=soilm, meanalpha=mean_alpha, pmodel_const=const
+        soilm=soilm, meanalpha=mean_alpha, coef=coef
     )
     ax2.plot(soilm, soilmstress, label=r"$\bar{{\alpha}}$ = {}".format(mean_alpha))
 
-ax2.axvline(x=const.soilmstress_thetastar, linestyle="--", color="black")
-ax2.axvline(x=const.soilmstress_theta0, linestyle="--", color="black")
+ax2.axvline(x=coef["thetastar"], linestyle="--", color="black")
+ax2.axvline(x=coef["theta0"], linestyle="--", color="black")
 
 secax = ax2.secondary_xaxis("top")
 secax.set_xticks(
-    ticks=[const.soilmstress_thetastar, const.soilmstress_theta0],
+    ticks=[coef["thetastar"], coef["theta0"]],
     labels=[r"$\theta^\ast$", r"$\theta_0$"],
 )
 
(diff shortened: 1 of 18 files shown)


docs/source/users/pmodel/pmodel_details/soil_moisture.md
```diff
@@ -28,37 +28,86 @@ very open area and a number of different methods are implemented in ``pyrealm``.
...
+### Effects on the [quantum yield of photosynthesis](./quantum_yield)
+
+* The `sandoval` method ({class}`~pyrealm.pmodel.quantum_yield.QuantumYieldSandoval`)
+ implements an experimental calculation that modulates $\phi_0$ as a function of the
+ temperature and a local aridity index.
+
+### Post-hoc penalties on gross primary productivity
+
+These approaches both use empirical
+functions of soil moisture and aridity  data that have been parameterised to align
```
MarionBWeinzierl:
```suggestion
functions of soil moisture and aridity data that have been parameterised to align
```

docs/source/users/pmodel/pmodel_details/soil_moisture.md
```diff
@@ -28,37 +28,86 @@ very open area and a number of different methods are implemented in ``pyrealm``.
...
+raw predictions from a P Model with field observations of GPP. Two penalty function
+are available that calculate the fraction of potential GPP that is realised given the
+effects of soil moisture stress:
+
+* The {func}`~pyrealm.pmodel.functions.calc_soilmstress_stocker` function implements the
+  {cite:t}`Stocker:2020dh` GPP penalty for use with the standard P Model.
+* {func}`~pyrealm.pmodel.functions.calc_soilmstress_mengoli` function implements the
+  {cite:t}`mengoli:2023a` GPP penalty for use with the subdaily P Model..
```
MarionBWeinzierl:
```suggestion
  {cite:t}`mengoli:2023a` GPP penalty for use with the subdaily P Model.
```

docs/source/users/pmodel/pmodel_details/soil_moisture.md
```diff
@@ -324,23 +321,74 @@ $$
     \end{cases}
 $$
 
+The function for the maximum attainable level $y$ reaches the maximum value of 1 at:
+
+$$
+\textrm{AI} = e^{-log(a)/b}
+$$
+
+Given the default values, this critical level as around 0.3457: plants can only realise
```
MarionBWeinzierl:
```suggestion
Given the default values, this critical level is around 0.3457: plants can only realise
```

MarionBWeinzierl (approved):
Looks fine to me - a little bit worried that some change might have been missed, but I cannot find anything, so it should be good.

davidorme:
> a little bit worried that some change might have been missed,

My bad @MarionBWeinzierl. I let the size of the PR get away from me.

PR #443, opened 2025-03-03
Regression tests for FaparLimitation class

### LINKED ISSUE #436: Create regression test for FaparLimitation class
In #403 , we implemented the FaparLimitation class to calculate LAI fAPAR_max values. We added some unit tests and an example code, but not yet integration/regression tests. This is issue is for implementing those regressions tests. We are doing this by

- [x] writing out the required annual, already "wrangled" data from the example from #403 into a netcdf file
- [x] writing a method to read in the data
- [x] writing a regression test using the class constructor
- [x] writing out the data for the `from_pmodel` class function
- [x] writing a method to read in the data for `from_pmodel`
- [x] writing a regression test using the `from_pmodel` class function

MarionBWeinzierl:
# Description

This PR adds regression tests for the FaparLimitation class. It adds the functions to create the test data as well as the data, and tests for the FaparLimitation constructor and it's `from_pmodel` class function.

Fixes #436 

## Type of change

## Key checklist

## Further checks


MarionBWeinzierl:
Currently, the regression test for `from_pmodel` fails, as the `fapar_max` and `lai_max` values are off by orders of magnitude. 

I tried putting in the `gpp_scale` from the example , but that makes things only worse. 

Another thing that is missing the the soil moisture penalty -- can that have such a large impact? Or am I missing something else in the Pmodel or PmodelEnvironment when constructing the test case? 

The gpp was currently my best guess for the source of failure, but I don't get anywhere with it. When I used the same function with the Pmodel in the example it gave the same results as the other method, so it must be in the Pmodel.

What am I missing?

davidorme (approved):
LGTM - just concentrating on the new test files.

PR #445, opened 2025-03-05
Update two leaf irradiance to V2.0.0 

davidorme:
# Description

@j-emberton 

This branch contains your branch merged with `develop` to bring in the version 2 API and then updates your code to use the new API.

It's probably going to be easier to see what I've done if your branch is also merged with `develop` :smile: 

ETA: This PR was initially against #245, with the intention that we merge it down to bring #245 into alignment with `develop`. But, in the end, we worked on this branch and PR directly, so @davidorme rebased this against `develop`, so merging this brings everything from #245 into `develop`


j-emberton:
Fair enough. I'll take a look just now :)

j-emberton (approved):
Nice work!

PR #452, opened 2025-03-13
Revise P Model attributes

### LINKED ISSUE #386: Align the attributes of the PModel and SubdailyPModel implementations

### LINKED ISSUE #451: Revise P Model attributes
I had a discussion with Colin about aligning some of the additional predictions across the PModel and SubdailyPModel.

* `PModel.lue`: This does not have a meaningful value for the SubdailyPModel, so is either an additional attribute of PModel, of we implement it as a property on the ABC and explicitly say `SubdailyPModel.lue` is not a thing.
* `PModel.iwue` should be estimable for the SubdailyPModel as $A/g_s$ - so we can add `g_s` and then `iwue` to the Subdaily model.
* The dark respiration calculation `PModel.rd` is sufficiently doubtful that it shouldn't be provided for _either_ model, so we should ditch it.

davidorme:
# Description

This PR:

* Removes estimation of dark respiration (`rd`) from the PModel and then from the `summarize` method and tests.
* Adds an `lue` property to `SubdailyPModel` that has the sole job of saying "LUE isn't calculated for this model". This does involve overriding an attribute of the base class, which is kinda ugly, but seemed the easiest way to do this that didn't involve a complex mess of getters and setters.
* Adds stomatal conductance (gs) and intrinsic water use efficiency (iwue) to the subdaily P Model

Fixes #451 

And also, I think this polishes off all of the final issues and fixes #386 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/users/pmodel/pmodel_details/envt_variation_outputs.md
+++ b/docs/source/users/pmodel/pmodel_details/envt_variation_outputs.md
@@ -230,7 +230,6 @@ resulting absorbed irradiance ($I_{abs}$).
 The productivity variables and their units are:
 
 * Gross primary productivity (``gpp``, $\mu\text{gC}\,\mathrm{m}^{-2}\,\text{s}^{-1}$)
-* Dark respiration (``rd``, $\mu\text{mol}\,\mathrm{m}^{-2}\,\text{s}^{-1}$)
 * Maximum rate of carboxylation
     (``vcmax``, $\mu\text{mol}\,\mathrm{m}^{-2}\,\text{s}^{-1}$)
 * Maximum rate of carboxylation at standard temperature
@@ -245,14 +244,6 @@ The productivity variables and their units are:
 :tags: [hide-input]
 
 plot_fun("gpp", r"GPP   ($\mu\mathrm{g\,C}\,\mathrm{m}^{-2}\,\mathrm{s}^{-1}$)")
-```
-
-### Dark respiration (``rd``)
-
-```{code-cell} ipython3
-:tags: [hide-input]
-
-plot_fun("rd", r"$r_d$   ($\mu\mathrm{mol}\,\mathrm{m}^{-2}\,\mathrm{s}^{-1}$)")
 ```
 
 ### Maximum rate of carboxylation (``vcmax``)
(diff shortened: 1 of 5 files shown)


MarionBWeinzierl (approved):
Looks all sensible to me

PR #454, opened 2025-03-14
Interim experimental markup implementation

### LINKED ISSUE #453: Interim experimental feature annotation
Ideally, we'd like to use a simple decorator to indicate experimental functionality (see #429) but that turns out to be hard to do while maintaining the `sphinx` autodoc correctly.

In the meantime, we do need to systematically indicate experimental features, so the issue should centralise some slightly cruder tools to achieve the same end.

davidorme:
# Description

This PR introduces a consistent setup for marking experimental parts of the code base. A decorator would be much cleaner (see #429 ) but turns out to be hard. This PR:

* Adds a `pyrealm.experimental.warn_experimental()` as a simple centralised warning generator.
* Adds a custom `sphinx` extension that adds an admonition below the title of experimental features. This relies on objects having an attribute `__experimental__` set to `True`. This is a bit fiddly , but it allows the text to be set in one location rather than pasted into every docstring. It turns out you can't simply append a global string onto a docstring...

I haven't applied this to the experimental parts yet - is the basic implementation sane?

Fixes #453

## Type of change

## Key checklist

## Further checks


pyrealm/core/experimental.py
```diff
@@ -0,0 +1,91 @@
...
+def warn_experimental(qualname: str) -> None:
+    """A simple wrapper function to generate an experimental warning."""
+
+    warn(qualname, ExperimentalFeatureWarning)
+
+
+def create_experimental_docstring(qualname: str) -> str:
+    """Generate an RST admonition to indicate experimental usage.
+
+    This cannot actually be used in practice because you cannot create a docstring
```
MarionBWeinzierl, line 78: I don't understand this comment - does this mean this is something that does not work? What is the reason that it is here, then?

MarionBWeinzierl (commented):
Still in the process of trying to understand this implementation.

pyrealm/core/experimental.py
davidorme: It's mostly as a note to myself that this approach doesn't work. I agree it's weird to leave it as a function. And I was also using it to save the format of the RST.

MarionBWeinzierl (approved):
I think this looks ok, although I do not understand the implementation fully.

Issue #410, opened 2025-02-03
Review variable names across code base


davidorme:
What plan do you have for renaming these vars?

_Originally posted by @j-emberton in https://github.com/ImperialCollegeLondon/pyrealm/pull/409#discussion_r1939380162_

We currently have a bit of variation in variable names and some very terse variable names (e.g. `pn` and `rh` for `precipitation` and `relative_humidity`). It would be sensible to make these more friendly and align them all.

PR #456, opened 2025-03-17
Rework of `pyrealm.core.solar` and downstream changes.

davidorme:
# Description

This branch was opened to add some user documentation to the core solar module, to run through what can be done with the module. In starting to do that, I noted a few things that needed tidying up first. And then it all got away from me a bit and sprawled...

* As noted in #410 , we have some really obscure and terse names for variables. This is particularly noticeable in `solar` and in `splash` because the reference implementation used terse names.
* There are some issues with the docstrings and equations - mostly just needing consistent formatting and some clearer separation of maths notation for variables and their name (which basically is the first point in equation shape).

So this PR is sprawling but is largely cosmetic:

* Revision of the core constant naming for the solar module to more readable names and grouping related coefficients.
* Functions in `core.solar` have been renamed (`calc_` to `calculate_`) and have new and clearer variable names.
* Arguments in `core.solar` now all use the specifically required constants, not just a single `const` instance definition. This makes the low level functions easier to follow, and the constants now all default to the specific attributes values from within `CoreConst`.
* The docstrings in `core.solar` have been revised to make the equation easier to follow by linking symbols to variable names and simplyifying the latex where possible.
* The `splash.DailySolarFluxes` has been revised to use these functions and - especially - now uses the private `_calculate` methods that avoid repeated recalculation of the intermediate `ru`, `rv` and `rw`. Those private functions were intended for just this use case, but just hadn't been rolled out into `splash`.
* Tests updated to pass the new changes.

There are probably some more renamings from #410 to put in, but this is pretty sprawling already.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


pyrealm/constants/core_const.py
```diff
@@ -92,29 +92,54 @@ class CoreConst(ConstantsClass):
...
+    solar_constant: float = 1360.8
+    """The solar constant (:math:`G_{sc}`, W/m^2, Kopp & Lean, 2011) - the long term
+    mean total solar irradiance."""
+
+    day_seconds: float = 86400
+    """The number of seconds in one solar day."""
+
+    equation_of_time_coef: tuple[float, ...] = (
+        0.000075,
+        +0.001868,
```
MarionBWeinzierl: It does not hurt to have `+`, but is this right or a typo? And where do these tuple values come from -- I cannot match them on the left

pyrealm/core/solar.py
```diff
@@ -1,924 +1,1066 @@
...
+               \left(\frac{1.0 - e^2}
+                          {1.0 + e \cos\left(\nu) \right)}
                \right)
              \right)^2
 
     Args:
-        nu: Heliocentric true anomaly (degrees)
-        k_e: Solar eccentricity
+        nu: Heliocentric true anomaly (:math:`\nu`, degrees)
+        solar_eccentricity: Solar eccentricity (:math:`e`), defaulting to
```
MarionBWeinzierl, line 39: When I read `e` in the equation above I automatically assumed Euler's number, so maybe a less misleading name would be better.

pyrealm/core/solar.py
```diff
@@ -1,924 +1,1066 @@
...
-    nu = (lambda_ - core_const.k_omega) % 360
+    nu = (lambda_ - solar_perihelion) % 360
 
-    return (nu, lambda_)
+    return nu, lambda_
 
 
-def calc_solar_elevation(site_obs_data: LocationDateTime) -> NDArray:
-    r"""Calculate the solar elevation angle for a specific location and times.
+# TODO - @davidorme commented this out - issues with typing of usage, which is aimed at
```
MarionBWeinzierl, line 830: Create an issue for this?

tests/regression/splash/conftest.py
```diff
@@ -2,10 +2,38 @@
...
+    "psy": "psy",
+    "econ": "econ",
+    "rx": "rx",
+    "hi": "hi",
+    "cond": "cond",
+    "eet_d": "eet_d",
+    "pet_d": "pet_d",
+    "aet_d": "aet_d",
+    "wn": "wn",
+    "ro": "ro",
```
MarionBWeinzierl, line 34: Should these also be changed to something more understandable?

pyrealm/core/solar.py
davidorme: Unfortunately that is the standard notation: https://en.wikipedia.org/wiki/Orbital_eccentricity

pyrealm/core/solar.py
davidorme: Yup agreed. Now active as #459

tests/regression/splash/conftest.py
davidorme: Yes - I think so, but that then also has sweeping changes, so I stopped here for now.

pyrealm/constants/core_const.py
davidorme: It's correct but inconsistent. The values were hard-coded into `calculate_equation_of_time` in `core.solar`.

MarionBWeinzierl (approved):
Great, thanks for the clarifications!

PR #448, opened 2025-03-12
Updated golden dataset for phenology module

davidorme:
# Description

This PR updates the `pyrealm_build_data/phenology` module with:

* A new set of input files
* A pure python workflow for all of calculations described in @zhouboya's paper
* Output files containing data from that workflow at half hourly (`hh`), daily and annual scales.

This new dataset and code should provide a much more robust set of regression tests for use in turning the implementation in the python workflow into part of the `pyrealm` codebase.

Fixes #367 (issue)

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

## Key checklist

- [x] Make sure you've run the `pre-commit` checks: `$ pre-commit run -a`
- [x] All tests pass: `$ poetry run pytest`

## Further checks

- [x] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

PR #460, opened 2025-03-19
Adopt changed data and scripts from #448

MarionBWeinzierl:
This PR is to adopt changes from #448 and merge them into the code for #443.

The problems with the `from_pmodel` results are not solved by this, and there is some discrepancy between the computed and the stored values also when using the constructor directly. Therefore, the regression tests still fail.


davidorme:
@MarionBWeinzierl I've pushed a commit that gets the `__init__` test working. There was a missing squared term, but also the golden dataset implementation had an error in the water limited FAPAR term. As it happens, all the years are energy limited (which isn't ideal) so it was a silent error on this dataset. The `FaparLimitation`  equation was correct for this.

I also updated the data handling in the test file - that was mostly just to make it a little more direct to rule out any accidental sort order issues with the inputs 😄.

That hasn't got the `from_pmodel` fixed. Will look at that next.

PR #461, opened 2025-03-20
Adding plant reproductive tissue respiration and turnover

### LINKED ISSUE #458: Add plant reproductive tissue & soil exudates to GPP allocation
Add a simple implementation of plant reproductive tissuse and soil exudates to the GPP allocation. Plant reproductive tissue will be defined as a set fraction of total foliage (thus it will scale accordingly) and the exudates should be a fixed portion of GPP topsliced. These rates should be assigned in the `PFT`.

sallymatson:
# Description

The purpose of this pull request is to add reproductive tissue and root exudates into the `StemAllometry` and `StemAllocation` models, so that these are traits which can be modifiable based on the plant PFT. However, the default state of this change should not change how the rest of the T model runs, as the T model does not include plant reproductive tissue. Key changes:

- Add necessary characteristics to the `PlantFunctionalType` models
- Add functions to calculate plant reproductive tissue mass, reproductive tissue respiration, and reproductive tissue turnover. 
- Calculate root exudates. 
- Account for GPP allocated to both reproductive tissue and root exudates in the NPP calculation. 
- Add relevant tests.

Fixes #458 

## Type of change

## Key checklist

## Further checks


sallymatson:
I'm sure this will need tweaks... Not sure if big or small. I tried to generally follow the same logic used for foliage / sapwood respiration and turnover. Let me know what you think when you have time @davidorme !

pyrealm/demography/flora.py
```diff
@@ -158,6 +162,11 @@ class PlantFunctionalTypeStrict:
     z_max_prop: float = field(init=False)
     """Proportion of stem height at which maximum crown radius is found."""
 
+    p_foliage_for_reproductive_tissue: float
+    """Proportion of foliage used to calculate reproductive tissue."""
+    p_gpp_for_root_exudation: float
+    """Proportion of GPP used to calculate root exudation."""
+
```
davidorme: Minor tweak to the description and some draft math notation for use in the docs.

```suggestion
    p_foliage_for_reproductive_tissue: float
    """Mass of reproductive tissue as a proportion of foliage mass (:math:`p_{rt}`, -)."""
    p_gpp_for_root_exudation: float
    """Allocation to root exudates as a proportion of GPP (:math:`p_{ex}`, -)"""
```

pyrealm/demography/flora.py
```diff
@@ -290,6 +303,8 @@ class Flora(PandasExporter):
     r"""Foliage turnover time (:math:`\tau_f`,years)"""
     tau_r: NDArray[np.float64] = field(init=False)
     r"""Fine-root turnover time (:math:`\tau_r`,  years)"""
+    tau_rt: NDArray[np.float64] = field(init=False)
+    r"""Reproductive tissue turnover time (:math:`\tau_rt`,years)"""
```
davidorme, line 308: I'd love to find a way not to duplicate these docstrings but I can't think of an easy way that doesn't break the autodoc.

pyrealm/demography/flora.py
```diff
@@ -303,6 +318,8 @@ class Flora(PandasExporter):
     r"""Sapwood-specific respiration rate (:math:`r_s`,  year-1)"""
     resp_f: NDArray[np.float64] = field(init=False)
     r"""Foliage maintenance respiration fraction (:math:`r_f`,  -)"""
+    resp_rt: NDArray[np.float64] = field(init=False)
+    r"""Reproductive tissue respiration rate (:math:`r_rt`,  -)"""
```
davidorme:
```suggestion
    r"""Reproductive tissue respiration rate (:math:`r_{rt}`,  -)"""
```

pyrealm/demography/flora.py
```diff
@@ -502,6 +526,8 @@ class StemTraits(PandasExporter, CohortMethods):
     r"""Sapwood-specific respiration rate (:math:`r_s`,  year-1)"""
     resp_f: NDArray[np.float64]
     r"""Foliage maintenance respiration fraction (:math:`r_f`,  -)"""
+    resp_rt: NDArray[np.float64]
+    r"""Reproductive tissue respiration rate (:math:`r_rt`,  -)"""
```
davidorme:
```suggestion
    r"""Reproductive tissue respiration rate (:math:`r_{rt}`,  -)"""
```

pyrealm/demography/tmodel.py
```diff
@@ -692,7 +814,9 @@ def calculate_growth_increments(
     )
 
     # Increment of diameter at breast height
-    delta_d = _enforce_2D((npp - turnover) / (dWsdt + dWfdt))
+    delta_d = _enforce_2D(
+        (npp - turnover - reproductive_tissue_turnover) / (dWsdt + dWfdt)
+    )
```
davidorme, line 860: Damn, I'm thinking that the denominator needs to account for the change in mass in reproductive tissue too.

pyrealm/demography/tmodel.py
```diff
@@ -653,6 +773,7 @@ def calculate_growth_increments(
         zeta: The ratio of fine root mass to foliage area of the PFT
         npp: Net primary productivity of individuals
         turnover: Fine root and foliage turnover cost of individuals
```
davidorme, line 813: I think we should separate out the mass and turnover for foliage and fine roots within the T model attributes. We want to use those masses in the Virtual Ecosystem and I think it is generally useful to have those attributes recorded.

pyrealm/demography/tmodel.py
davidorme: OK - so I think we have the term `dWfdt`, which is the rate of change in foliage and fine root mass:

```python
dWfdt = (
    lai
    * ((np.pi * ca_ratio) / (4 * a_hd))
    * (a_hd * dbh * (1 - stem_height / h_max) + stem_height)
    * (1 / sla + zeta)
)
```

I think we can partition this to give:

```python
rate_of_change_in_foliage_area = (
    lai
    * ((np.pi * ca_ratio) / (4 * a_hd))
    * (a_hd * dbh * (1 - stem_height / h_max) + stem_height)
)
dWfoliage_dt = rate_of_change_in_foliage_area * (1/sla)
dWfineroot_dt = rate_of_change_in_foliage_area * zeta
dWreproductive_dt = dWfoliage_dt * p_foliage_for_reproductive_tissue
```

And then the divisor becomes `(dWsdt + dWfoliage_dt + dWfineroot_dt + dWreproductive_dt)`. I think that leaves the classic T model identical when `p_foliage_for_reproductive_tissue = 0`.

Actually, now that I've typed it out, I don't think there is any mileage in exposing those different `dW` values, so we might as well just have :

```python
dWfdt = (
    lai
    * ((np.pi * ca_ratio) / (4 * a_hd))
    * (a_hd * dbh * (1 - stem_height / h_max) + stem_height)
    * ((1 + p_foliage_for_reproductive_tissue) / sla + zeta)
)
```

davidorme (changes requested):
This looks good - there are a few minor naming tweaks and then the additional stuff we discussed in Teams: adding the optional topslice for active nutrient pathways and splitting up the root and foliar terms.

I also think we need to update the delta D calculation, because otherwise diameter increments won't change as the allocation to reproduction increases. Does my suggestion make sense to you?

pyrealm/demography/tmodel.py
sallymatson: Yes this makes sense!!

pyrealm/demography/tmodel.py
sallymatson: I now separated these out for the calculations and as separate class attributes. I also left the `turnover` attribute, in case it is useful for others? Can remove it if you think better.

pyrealm/demography/tmodel.py
```diff
@@ -445,6 +445,67 @@ def calculate_foliar_respiration(
...
+) -> NDArray[np.float64]:
+    r"""Calculate gpp topslice.
+
+    Calculates a fixed proportion of the total GPP for the crown that is removed before
+    further GPP allocation. This is intended as a helper variable for T Model users to
+    simulate processes not included in the T Model such as root exudation or active
+    nutrient servicing for mycorriza fungi.
+
+    NOTE: This is a naieve calculation method that is not part of the T model. If values
+    for GPP topslice are zero it will have no impact on the T Model calculations.
```
davidorme: Here and below, we can use this RST syntax to style the note as an admonition, which just helps draw attention to it. See e.g. https://pyrealm.readthedocs.io/en/v2.0.0-rc.1/api/pmodel_api.html#pyrealm.pmodel.optimal_chi.OptimalChiLavergne20C4


```suggestion
    .. NOTE:: 
        
        This is a naive calculation method that is not part of the T model. If values
        for GPP topslice are zero it will have no impact on the T Model calculations.
```

davidorme (approved):
LGTM - one minor formatting docs thing, but otherwise looks good to merge.

pyrealm/demography/tmodel.py
arne-exe: Hi @davidorme and @sallymatson, I was reading through this PR to better understand how the code addresses allocation under the T model. Could you please help me understand why the function for dWfdt does not multiply by dD/dt at the end (similar to the middle term in equation 16 of Li et al.)?

arne-exe: > Damn, I'm thinking that the denominator needs to account for the change in mass in reproductive tissue too.

Hi @davidorme, looking at this equation, would it be correct to assume that the unit for delta_d is some form of mass of carbon)? Also, can you please explain why you chose to divide by (dWsdt + dWfdt) instead of subtracting (dWfdt) from npp using the rest to represent dWsdt (reflecting the mass of carbon available for radial increment)? I would also like to understand how that equation is then able to calculate the actual dbh increment (m) and if/how this is linked to the relationship between diameter increment and stem increment (equation 7 in Li et al.).

davidorme: @arne-exe The unit for delta_D _is_ the actual DBH increment in metres. 


As I think about it, the equation is basically:

$$
\Delta D  \textrm{  [m]} = \frac{\text{NPP [kg C]}}{\text{Growth costs [kg C/m]}}
$$

The two terms ``dWsdt + dWfdt`` from the code above are those growth costs for the stem, foliage and fine roots. The equations to calculate them are all driven by the diameter at breast height $D$ and the PFT parameterisation. They basically capture - given the current size - how much carbon is required to increase the diameter: they are a point estimate of a slope of the changing carbon costs of diameter change with size. The $\Delta D$ is then the available carbon mass divided through by the combined costs per metre increase in diameter.

Does that make sense?

arne-exe: Hi @davidorme, thank you for the clarification, although I still have some questions:
- It seems like the unit of growth costs would be kg C over time, or if you use a delta approach this would just be kg C. Since the delta_D uses a delta approach, would the growth costs also need to be written using delta, instead of the derivative over time? Also, I don't understand where the metre in the unit of growth costs comes from. My guess is that the metre in this equation should come from multiplying this entire equation by the derivative of change in stem diameter (dD/dt) or delta_D, although currently this is not done (see my other comment in this PR).
- Lastly, I vaguely understand the reasoning of this equation and my earlier comment still remains, in that I don't understand why the formula should reflect the diameter increment. The main thing I struggle with is that the diameter increment should be directly related to the mass of carbon available (i.e., the radial increment should come from available carbon for stem growth accounting for wood density and stem volume), but at the moment, it seems like the foliage carbon is also included here. Would you agree that the current formula doesn't completely make sense?

When trying to calculate the diameter increment, have you considered the following alternative:
- Use equation 15 and rewrite it for dWs/dt (which then becomes: dWs/dt = Pnet - T - (1+zeta*sigma)dWf/dt)
- This should give us the amount of carbon available for diameter increment.
- Then use this available carbon to implement it in equation 7 and solve the equation for dD/dt or delta_D.

Finally, I'd like to mention that these comments are all with good intentions, and I promise I'm not doing it just to criticise your work :smiley:.

davidorme: > Finally, I'd like to mention that these comments are all with good intentions, and I promise I'm not doing it just to criticise your work 😃.

This hadn't even occurred to me. All of this needs to be accessible to developers and end users.

davidorme: > * but at the moment, it seems like the foliage carbon is also included here.

And indeed the fine root and now reproductive tissue carbon. 

I think the thing here is that the allocation model is deliberately incorporating all of the potential carbon costs at this point. It doesn't grow the stem and then subsequently add foliage and fine roots at a later stage - the allometry equations scale up the whole tree: if we increase the stem diameter by 1cm, how much stem mass do we add but also how much extra foliage will the larger tree require, how much will the fine root mass increase. So that denominator term includes all of those cost terms for increasing $\Delta D$

arne-exe: > @arne-exe The unit for delta_D _is_ the actual DBH increment in metres.
> 
> As I think about it, the equation is basically:
> 
> Δ D  [m] = NPP [kg C] Growth costs [kg C/m]
> 

If the numerator represents NPP minus costs for turnover, and the denominator represents the total costs for growth for all tissues, then both will cancel each other out and delta_D will always be 1 (this follows from equation 15).

davidorme: But the denominator isn't a mass, it's a rate: the combined cost per unit increase in $\Delta D$. So if you have 12 kg Carbon and the growth costs are 1200 kg per m increase then $\Delta D = 12 \text{ kg} /1200\textrm{ kg m}^{-1}  = 0.01 \textrm{ m}$

arne-exe: Aaaaaaah now I see. So that's why you don't multiply by dD/dt in the equation for dWs/dt. You divide by dD/dt which makes it a rate. Oof, thought I was going crazy for a second :satisfied:

davidorme: I'm glad that's resolved it! Was beginning to doubt myself - really useful to work through how it works!

PR #462, opened 2025-03-20
459 review two leaf model

### LINKED ISSUE #459: Review implementation of Two Leaf model
This issue is to go through the new `TwoLeaf` model and:

* work through the user experience of fitting the models to think about how it will be used and to make it as clean as possible.
* revise docstrings.

davidorme:
# Description

This is a heavy revision of the `two_leaf` module. The tests are failing but it would be good to get an eye on the new version before I commit to revamping the tests.

Much of the functionality is the same but the structure has changed enough that `git` has decided the module is a new file. There might be a way to get it to show the old and new as side by side?

---

Big picture changes:

* The constants have moved (`two_leaf_canopy.py`-> `two_leaf.py`), added several new constants that were hard-coded into methods, and I've renamed some of the constants and added detailed docstrings. There is also a new derived constant that gets set in `__post_init__`.

* The two main classes have been heavily revised:

   * The main docstrings have been revamped to add a more descriptive flow of the science around the previous sequence of function calls.
   * The attributes all now have docstrings - some are renamed for clarity.
   * The internal calculations now happen automatically by calling the private calculation method from within `__init__` rather than requiring users to do that manually by creating the instance and then calling the calculation method.
   * The functionality within those private methods was structured as a long set of calls to subsidiary function methods. That remains identical except for a handful of extremely simple calculations, where I think the model and docs are actually clearer if they don't dive off into a subroutine.
   * I've updated those subsidiary functions with clarified docstrings and added clear defaults to arguments representing constants. Most of the docstring changes are focussed on defining the mathematical symbols for the arguments and then updating the equations.
   * A couple of minor bugs.

* @j-emberton also implemented a wrapper function for solar positions (`calculate_solar_elevation`) that used a dataclass `LocationDateTime` as an input, which provided solar information for a single site. I think the main use case here is going to be for users to get solar elevation for the TwoLeafModel, where we want to support array calculations for many sites, so I've merged these two bits of code into a single class `SolarPositions` that accepts arrays and then calculates solar elevations automatically. This can then be used as a direct input for array inputs to `TwoLeafIrradiance`.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


tests/regression/two_leaf_canopy/test_two_leaf_model.py
```diff
@@ -52,58 +55,70 @@ def assimilation_single_day(solar_irradience, get_data):
...
-    )
-
-    solar_irradience.calc_absorbed_irradience()
+    # Standard P Model, simple Arrhenius scaling, phi_0 = 1/8
+    standard_pmod = PModel(pmod_env)
 
     assim = TwoLeafAssimilation(
-        standard_pmod, solar_irradience, leaf_area_index=data["LAI"].to_numpy()
+        pmodel=standard_pmod,
+        irrad=solar_irradiance,
```
MarionBWeinzierl: Minor thing, but maybe it's worth investing into a couple of extra letters for clarity, and make this "irradiance"

MarionBWeinzierl (commented):
This is a bit difficult to process. As far as I can track the changes around the restructuring makes sense.

Do I understand it correctly that the changes in `core` are manly refactoring and moving code around, rather than functional changes?

Also, you marked this as a bug fix - was the issue just a lack of clarity, or some functional bug?

davidorme:
> This is a bit difficult to process. As far as I can track the changes around the restructuring makes sense.
> 

Yes - apologies. 

> Do I understand it correctly that the changes in `core` are manly refactoring and moving code around, rather than functional changes?
> 

That's correct. The biggest functional "changes" are:
* the conversion of the `DateTimeLocation` and `calculate_solar_positions` in the `SolarPositions`, and
* the removal of a handful of functions that I thought were sufficiently simple to be cleaner as internal calculations within classes rather than functions.

> Also, you marked this as a bug fix - was the issue just a lack of clarity, or some functional bug?

It was a functional bug: 

https://github.com/ImperialCollegeLondon/pyrealm/blob/e2ac2997f5be5f57c985dbf59726427142186272/pyrealm/pmodel/two_leaf.py#L391-L395

That `diffuse_extinction_coef` term was incorrectly a repeat of the `scattered_beam_extinction_coef`.

tests/regression/two_leaf_canopy/test_two_leaf_model.py
davidorme: Yup - I think thats cleaner.

pyrealm/pmodel/two_leaf.py
```diff
@@ -0,0 +1,907 @@
...
+        irradiance: TwoLeafIrradience,
+    ):
+        """Initialize the TwoLeafAssimilation class."""
+
+        self.pmodel = pmodel
+        """A PModel or SubdailyPModel instance."""
+        self.irradiance = irradiance
+        """A TwoLeafIrradiance instance """
+
+        # TODO - both the pmodel and irrad instances have their own independent
```
MarionBWeinzierl, line 605: new issue?

pyrealm/pmodel/two_leaf.py
```diff
@@ -0,0 +1,907 @@
...
+
+def calculate_sun_vcmax25(
+    leaf_area_index: NDArray[np.float64],
+    vcmax25: NDArray[np.float64],
+    canopy_extinction_coef: NDArray[np.float64],
+    beam_extinction_coef: NDArray[np.float64],
+) -> NDArray[np.float64]:
+    r"""Calculate standardised carboxylation rate of sunlit leaves.
+
+    Calcultes the maximum carboxylation rate for sunlit leaves at the standard
```
MarionBWeinzierl:
```suggestion
    Calculates the maximum carboxylation rate for sunlit leaves at the standard
```

MarionBWeinzierl (approved):
I can understand now why git treats pmodel/two_leaf and pmodel/two_leaf_irradiance as two different files. :D

pyrealm/pmodel/two_leaf.py
davidorme: Yup - agreed and now exists as #469

PR #468, opened 2025-03-26
Implement `from_subdailypmodel` for FaparLimitation

### LINKED ISSUE #464: Create from_subdailypmodel function for FaparLimitation
Discussions with @davidorme and other Lemontree members have revealed that the current version of #403 is insufficient.

The python example for FaparLimitation uses a SubdailyPModel, while the implementation in #403 so far only has `from_pmodel`. A now `from_subdailypmodel` function should be developed. This also needs the soil moisture stress correction, i.e., a `gpp_penalty` argument.

MarionBWeinzierl:
# Description

As explained in #464, the current version of #403 is insufficient.

This PR implements a new `from_subdailypmodel` function, also adding the soil moisture stress correction, i.e., a `gpp_penalty` argument.

Fixes #464 

## Type of change

## Key checklist

## Further checks


pyrealm/phenology/fapar_limitation.py
```diff
@@ -190,6 +190,28 @@ def get_annual(
...
+        datetimes: Subdaily datetimes as np.datetime64 array.
+    """
+
+    subdaily_x = np.zeros(len(datetimes))
+
+    n_days = len(x)
+    obs_per_day = int(len(datetimes) / n_days)
+
+    for i in range(n_days):
+        subdaily_x[i * obs_per_day : i * obs_per_day + obs_per_day] = x[i]
```
davidorme: I think `np.repeat` is likely more efficient here:

```python
In [7]: np.repeat([1,2,3], 5)
Out[7]: array([1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3])
```

But, having said that I think we should insist on users doing this.

davidorme (changes requested):
This looks good. Two minor things:

1. I think we should generalise `soil_moisture_stress` to `gpp_penalty_factor`? We only currently have soil moisture stress based penalties, but it would apply to any general post-hoc penalty. 
2. I think we should insist that the subdaily soil moisture stress is at subdaily frequency. That does add a burden onto users - which we could support with a function - but it does mean that we can use that mechanism to apply penalties with subdaily resolution.

MarionBWeinzierl:
OK, I am merging this now, as I implemented the requested changes, and will continue to work and fix things from the `348` branch

PR #472, opened 2025-03-28
Update documentation T model

### LINKED ISSUE #471: Update documentation T model on readthedocs
This issue provides feedback on the documentation of the T model in Pyrealm on readthedocs.
I'm adding it here as it may be useful for the documentation of the VE.

**calculate_crown_fractions:** I believe that “stem_height” refers to tree height. I think the documentation could be more exact when referring to things like this (i.e., there should only be 1 interpretation possible). At the moment, stem height could be interpreted as total height or height below first main branch. The text also needs a space after “(fc)”.
**calculate_fine_root_respiration:** the text says “calculate foliar respiration”
**calculate_foliage_and_fine_root_turnover:** second “tau_f” needs to be replaced in the equation with “tau_r”.
**calculate_foliage_masses:** make sure to clarify the link with the crown module (i.e., how it calculates foliage area for different height layers).
**calculate_net_primary_productivity:** final “P” needs to be replaced by “Wf” (foliage mass).
**calculate_stem_masses:** functionally speaking, how does the stem mass differ from the sapwood mass in the model (i.e., what do both exactly refer to in the model and how are they different based on the equations)? Note that the equation for stem mass assumes no heartwood and a paraboloid stem shape.
**calculate_sapwood_masses:** In the text the symbol for crown area ratio should be changed from “Ac” to “c”

It would be easier if each parameter had its unit between brackets. It’s rather difficult to keep track of all the units and it’s not convenient having to scroll up to the summary table every time you’re trying to understand a certain equation. Also, equations have symbols but the legends for the parameters have text instead of their symbols, so it defeats the purpose a having a legend, perhaps have both text and symbol (or at least symbol) and unit in the legend? Also note the importance of distinguishing units for mass as “kg” or “kg Carbon”.

I think what would make the API easier to understand is to have a flow chart of all the processes in the plant model. At least for the T model it would be very good to start from total GPP and then be able to visually see how that total pool is split up into different components. Currently, I am getting a bit lost in the list of API terms, equations and symbols.
In general, the order of this section could be rearranged so that different calculations build on top of each other (e.g., the explanation for calculating “Wf” should come before other equations that use “Wf”).

arne-exe:
# Description

This pull request makes changes to the tmodel.py script, particularly the docstrings so that the documentation for the T model can be updated on readthedocs. The changes consist of fixing spelling errors and clarifying some of the descriptions and units.

Partially fixes #471 - additional PR needed to add units

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/users/demography/t_model.md
+++ b/docs/source/users/demography/t_model.md
@@ -68,9 +68,9 @@ the predictions of the T Model for:
 * Crown radius scaling factor (`crown_r0`, -), and
 * Height of maximum crown radius (`crown_z_max`, m).
 
-Note that :mod:`~pyrealm.demography.stem_height` denotes the total tree height,
-as used interchangeable in Li et al. (2014), rather than just the height of the trunk
-below the canopy.
+Note that {attr}`~pyrealm.demography.tmodel.StemAllometry.stem_height` denotes the total
+tree height, as used interchangeable in {cite:t}`Li:2014bc`, rather than just the height
+of the trunk below the canopy.
 
 The DBH input can be a scalar array or a one dimensional array providing a single value
 for each PFT. This then calculates a single estimate at the given size for each stem.
--- a/docs/source/users/demography/t_model.md
+++ b/docs/source/users/demography/t_model.md
@@ -68,6 +68,10 @@ the predictions of the T Model for:
 * Crown radius scaling factor (`crown_r0`, -), and
 * Height of maximum crown radius (`crown_z_max`, m).
 
+Note that :mod:`~pyrealm.demography.stem_height` denotes the total tree height,
+as used interchangeable in Li et al. (2014), rather than just the height of the trunk 
+below the canopy.
+
 The DBH input can be a scalar array or a one dimensional array providing a single value
 for each PFT. This then calculates a single estimate at the given size for each stem.
 
--- a/docs/source/users/demography/t_model.md
+++ b/docs/source/users/demography/t_model.md
@@ -69,7 +69,7 @@ the predictions of the T Model for:
 * Height of maximum crown radius (`crown_z_max`, m).
 
 Note that :mod:`~pyrealm.demography.stem_height` denotes the total tree height,
-as used interchangeable in Li et al. (2014), rather than just the height of the trunk 
+as used interchangeable in Li et al. (2014), rather than just the height of the trunk
 below the canopy.
 
 The DBH input can be a scalar array or a one dimensional array providing a single value
--- a/pyrealm/demography/tmodel.py
+++ b/pyrealm/demography/tmodel.py
@@ -149,14 +149,14 @@ def calculate_crown_fractions(
 ) -> NDArray[np.float64]:
     r"""Calculate tree crown fraction under the T Model.
 
-    The crown fraction (:math:`f_{c}`)is calculated from individual diameters at breast
-    height and stem height (:math:`D`), along with the initial slope of the height /
+    The crown fraction (:math:`f_{c}`) is calculated from individual diameters at breast
+    height (:math:`D`) and stem height (:math:`H`), along with the initial slope of the height /
     diameter relationship (:math:`a`) of the plant functional type
     :cite:p:`{Equation 11, }Li:2014bc`:
 
     .. math::
 
-        \frac{H}{a D}
+        f_{c} =\frac{H}{a D}
 
     Args:
         a_hd: Initial slope of the height/diameter relationship of the PFT
--- a/pyrealm/demography/tmodel.py
+++ b/pyrealm/demography/tmodel.py
@@ -1,5 +1,5 @@
 """The ``t_model`` module provides the basic scaling relationships of the T Model
-:cite:`Li:2014bc`:. This provides scaling relationships using the plant functional type
+:cite:`Li:2014bc`. This provides scaling relationships using the plant functional type
 traits defined in the :mod:`~pyrealm.demography.flora` module and the diameter at breast
 height of individual stems to define the stem geometry, masses, respiration and hence
 calculate stem growth given net primary productivity.
--- a/pyrealm/demography/tmodel.py
+++ b/pyrealm/demography/tmodel.py
@@ -909,7 +909,7 @@ class StemAllocation(PandasExporter):
     turnover: NDArray[np.float64] = field(init=False)
     """Allocation to leaf and fine root turnover (g C)"""
     delta_dbh: NDArray[np.float64] = field(init=False)
-    """Predicted increase in stem diameter from growth allocation (g C)"""
+    """Predicted increase in stem diameter from growth allocation (m)"""
     delta_stem_mass: NDArray[np.float64] = field(init=False)
     """Predicted increase in stem mass from growth allocation (g C)"""
     delta_foliage_mass: NDArray[np.float64] = field(init=False)
--- a/pyrealm/demography/tmodel.py
+++ b/pyrealm/demography/tmodel.py
@@ -2,9 +2,10 @@
 :cite:`Li:2014bc`. This provides scaling relationships using the plant functional type
 traits defined in the :mod:`~pyrealm.demography.flora` module and the diameter at breast
 height of individual stems to define the stem geometry, masses, respiration and hence
-calculate stem growth given net primary productivity. Note that :mod:`~pyrealm.demography.stem_height` denotes the
-total tree height, as used interchangeable in Li et al. (2014), rather than just the height
-of the trunk below the canopy.
+calculate stem growth given net primary productivity. Note that
+:attr:`~pyrealm.demography.tmodel.StemAllometry.stem_height` denotes the total tree
+height, as used interchangeable in :cite:`Li:2014bc`, rather than just the height of the
+trunk below the canopy.
 """  # noqa: D205
 
 from dataclasses import InitVar, dataclass, field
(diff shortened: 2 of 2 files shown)


arne-exe:
I made one change so far and created the pull request (haven't assigned anyone for review yet because I will make more changes).
I like GitHub for tracking the overview of different commits. In general, is it ok to make a pull request before having finished all the changes, or should I only create the pull request once I've made all the changes and it's ready for review?

davidorme:
> I like GitHub for tracking the overview of different commits. In general, is it ok to make a pull request before having finished all the changes, or should I only create the pull request once I've made all the changes and it's ready for review?

That is absolutely fine and standard practice. If you look at the top right of this page, you will see a link to "convert to draft". You can actually create a PR as draft in the first place too (There is a dropdown menu on the create PR button).

davidorme:
@arne-exe I'm just moving your comment on the issue over here.

In general, we'd use the issue to describe the thing to change and once we've agreed it _is_ a thing to change (which sometimes goes without saying) then the discussion on the actual changes happens on the PR conversation here.

> General notes:
-You can hover over the function in VSC to get a popup window that shows you how the “rendered” docstring looks like, this may be sufficient for now to see what the text looks like under RST format.
> Things we could adjust later:
> -calculate_crown_fractions: I believe that “stem_height” refers to tree height (following Li et al., 2014). However, I didn’t change it to tree height because one of the arguments is called stem_height.

I think Li et al. 2014  use stem and tree interchangeably. For example, they say "the form of stem is assumed to be paraboloid" and the equation that follows uses tree height $H$. I get that stem could mean something different but I don't think Li et al 2014 mean it to? I'm not inclined to change all of the arguments, but we could definitely add something to the doc introduction that makes it clear that we mean total height of the tree stem.

> -The unit of delta_dbh is currently in g C, it seems like this should be in mm/cm/m?

Indeed. And it is metres.

Also, I spent a few seconds trying to figure out what units of mm per cm per m where before I realised it was just uncertainty 😆 

> -Whether or not to include units under each function can be decided later.

I'm not 100% sure what you mean here but I think what you're saying is that it would be clearer if the docstrings systematically laid out the notation and units in the function arguments. So for example:

```rst
Calculate turnover costs.

This function calculates the costs associated with the turnover of fine roots and
foliage, given the foliage mass and plant functional traits, following Equation 15
of :cite:t:`Li:2014bc`.

.. math::

    T = W_f \left( \frac{1}{\tau_f} + \frac{\sigma \zeta}{\tau_r} \right)

Args:
    sla: The specific leaf area (:math:`\sigma`, ???)
    zeta: The ratio of fine root mass to foliage area (:math:`\zeta`, ???)
    tau_f: The turnover time of foliage (:math:`\tau_f`, years -1)
    tau_r: The turnover time of fine roots (:math:`\tau_r`, years -1)
    foliage_mass: The foliage mass (:math:`W_f`, kg Carbon)
    validate: Boolean flag to suppress argument validation
```

If that's what you mean then I completely agree - its much more compact and I've just been making very similar changes to the docstrings of the two leaf, two stream model. I don't know if you want to do these changes here or on another PR. I think I would suggest another PR as @sallymatson has an open PR on this file too (#461).

> -Whether or not to rearrange the order of functions in readthedocs can be decided later.

That is a setting we'd need to change in another file but I'm completely happy to do so. What is a sensible order?

arne-exe:
> @arne-exe I'm just moving your comment on the issue over here.
> 
> In general, we'd use the issue to describe the thing to change and once we've agreed it _is_ a thing to change (which sometimes goes without saying) then the discussion on the actual changes happens on the PR conversation here.
> 

Alright, that's good to know! Seems like my issue broke nearly every possible rule haha. Thanks for being patient with me while I learn the correct way of doing things.

> > General notes:
> > -You can hover over the function in VSC to get a popup window that shows you how the “rendered” docstring looks like, this may be sufficient for now to see what the text looks like under RST format.
> > Things we could adjust later:
> > -calculate_crown_fractions: I believe that “stem_height” refers to tree height (following Li et al., 2014). However, I didn’t change it to tree height because one of the arguments is called stem_height.
> 
> I think Li et al. 2014 use stem and tree interchangeably. For example, they say "the form of stem is assumed to be paraboloid" and the equation that follows uses tree height H . I get that stem could mean something different but I don't think Li et al 2014 mean it to? I'm not inclined to change all of the arguments, but we could definitely add something to the doc introduction that makes it clear that we mean total height of the tree stem.
> 

I agree that Li et al. use it interchangeable, and I also understand that changing the arguments is a pain given that it is not a strictly essential change. Your suggestion of adding something in the intro that clarifies it should work well, with the main purpose of avoiding confusion for people that use Pyrealm (without having read the Li et al. paper).

> > -The unit of delta_dbh is currently in g C, it seems like this should be in mm/cm/m?
> 
> Indeed. And it is metres.
> 
> Also, I spent a few seconds trying to figure out what units of mm per cm per m where before I realised it was just uncertainty 😆
> 

My bad! I'll change the unit accordingly.

> > -Whether or not to include units under each function can be decided later.
> 
> I'm not 100% sure what you mean here but I think what you're saying is that it would be clearer if the docstrings systematically laid out the notation and units in the function arguments. So for example:
> 
> ```
> Calculate turnover costs.
> 
> This function calculates the costs associated with the turnover of fine roots and
> foliage, given the foliage mass and plant functional traits, following Equation 15
> of :cite:t:`Li:2014bc`.
> 
> .. math::
> 
>     T = W_f \left( \frac{1}{\tau_f} + \frac{\sigma \zeta}{\tau_r} \right)
> 
> Args:
>     sla: The specific leaf area (:math:`\sigma`, ???)
>     zeta: The ratio of fine root mass to foliage area (:math:`\zeta`, ???)
>     tau_f: The turnover time of foliage (:math:`\tau_f`, years -1)
>     tau_r: The turnover time of fine roots (:math:`\tau_r`, years -1)
>     foliage_mass: The foliage mass (:math:`W_f`, kg Carbon)
>     validate: Boolean flag to suppress argument validation
> ```
> 
> If that's what you mean then I completely agree - its much more compact and I've just been making very similar changes to the docstrings of the two leaf, two stream model. I don't know if you want to do these changes here or on another PR. I think I would suggest another PR as @sallymatson has an open PR on this file too (#461).
> 

Yes, that's exactly what I meant. The reason for this was that it get's quite hard to keep track of all the units when trying to understand an equation, especially when you have to scroll back up to the main table that has the units. So I think adding it between parentheses like your example would help a lot. 
I could add the units in a new PR, assuming that most of them have the same units as in Li et al. (2014)?

> > -Whether or not to rearrange the order of functions in readthedocs can be decided later.
> 
> That is a setting we'd need to change in another file but I'm completely happy to do so. What is a sensible order?

Again, this is slightly nitpicking. Overall, all the info is there and the menu on the left side allows you to jump to specific functions easily, which is great. The reason why I brought this up was that some of the functions use variables that have their formula discussed after they are first mentioned (I think this is the case for foliage mass). I don't think this needs a lot of change, and as I go through the units I'll pay attention to which things may need to be moved up or down. As you said, it should just be a matter of changing the order of indexing (not requiring changes in the actual python script).

davidorme:
@arne-exe Is there more to come on this PR? I think the restructuring of the notation and units into the function arguments is worth breaking out into a new PR, but did you want to add something on this:

> I agree that Li et al. use it interchangeable, and I also understand that changing the arguments is a pain given that it is not a strictly essential change. Your suggestion of adding something in the intro that clarifies it should work well, with the main purpose of avoiding confusion for people that use Pyrealm (without having read the Li et al. paper).

arne-exe:
> @arne-exe Is there more to come on this PR? I think the restructuring of the notation and units into the function arguments is worth breaking out into a new PR, but did you want to add something on this:
> 
> > I agree that Li et al. use it interchangeable, and I also understand that changing the arguments is a pain given that it is not a strictly essential change. Your suggestion of adding something in the intro that clarifies it should work well, with the main purpose of avoiding confusion for people that use Pyrealm (without having read the Li et al. paper).

Hi @davidorme, I've had a quick scan where would be the best place for adding that info, and I think it would fit best either:
- https://pyrealm.readthedocs.io/en/latest/api/demography_api.html#module-pyrealm.demography.tmodel (at the end of that first paragraph) or
- https://pyrealm.readthedocs.io/en/latest/users/demography/t_model.html#stem-allometry (as a separate single line just below the arguments)

I was thinking of adding something like: "Note that 'stem_height' denotes the total tree height, following Li et al. (2014), where 'stem' and 'tree' are used interchangeably.".

Which of the two options do you prefer?

davidorme:
I think both places make sense - different audiences, so some duplication is needed sometimes. I'd also add what it might have meant (so, not the height of the trunk below the canopy?)

arne-exe:
> I think both places make sense - different audiences, so some duplication is needed sometimes. I'd also add what it might have meant (so, not the height of the trunk below the canopy?)

Great, both places sounds great. I've added it to the tmodel script and the script is ready for merging.
As for the other place (in the user section), where exactly can I find the document that needs changing for the added line to be included?

Edit: think I found it, in the source tab. Give me one sec.

arne-exe:
Okay @davidorme, I think things are ready for approval

davidorme (approved):
Thanks @arne-exe. Looks good to me.

You'll probably have seen that code checks showed some failures. If you look at the output here:

https://results.pre-commit.ci/run/github/317618141/1743437038.DyS7tgowSQO-khwsTVo-_Q

You'll see it is complaining about lines that are too long (more than 88 characters). With the full developer set up, this kind of thing gets flagged and is easy to fix 😄 There are also a couple of minor linking issues (the path to the the `stem_height` attribute within the code, and using the citation code for Li et al). I've pushed a commit to fix up those minor things, because they are fiddly. Hopefully that should get the PR through all the checks and you can merge it!

PR #480, opened 2025-04-10
Moving potential_gpp calculation outside StemAllocation

### LINKED ISSUE #479: StemAllocation is pinned to full canopy model
The StemAllocation class unavoidably uses the calculation of actual GPP from Equation 12 of Li et al 2014, $P_0 A_c (1 − e^{−kL})$. That is assuming a big leaf approximation of the assimilation with a single fraction of absorbed light through the whole canopy. 

That isn't compatible with the `Canopy` implementation, which has layers of leaves following the PPA, each of which has its own fAPAR. The `StemAllocation` should probably take an _actual_ GPP and the demography module should provide tools to implement different GPP calculation options.

To begin with, we can fix this by adding an "actual GPP" init argument, that bypasses the current calculation. We might then want to remove the current calculation and move it to a new standalone function that can be used to calculate an "actual GPP" input.

davidorme:
# Description

This PR takes the calculation of GPP outside of the `StemAllocation` class so that users can provide GPP calculated by different approaches, not just the whole crown approach from Equation 12 of Li et al 2014.

The user docs are updated to show how it works and the tests are updated to match

Fixes #479

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/users/demography/t_model.md
+++ b/docs/source/users/demography/t_model.md
@@ -35,7 +35,11 @@ import numpy as np
 import pandas as pd
 
 from pyrealm.demography.flora import PlantFunctionalType, Flora
-from pyrealm.demography.tmodel import StemAllocation, StemAllometry
+from pyrealm.demography.tmodel import (
+    StemAllocation,
+    StemAllometry,
+    calculate_whole_crown_gpp,
+)
 ```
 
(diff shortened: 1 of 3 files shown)


sallymatson (approved):
LGTM

PR #485, opened 2025-04-15
Add subdaily check

### LINKED ISSUE #416: Add failure for scaler if not subdaily
We might want to use the `SubdailyScaler` functions from the outside, for example to take daily or annual means of data. The class should check for the data to be actually subdaily, and not daily or monthly, for example, and fail otherwise.

MarionBWeinzierl:
# Description

The underlying issue #416 wanted to add a check for `SubdailyScaler` to make sure that the data is indeed subdaily. The scaler has in the meantime been replaced by the `AcclimationModel`, which is paired with the SubdailyPModel. 

Fixes #416 

## Type of change

## Key checklist

## Further checks


davidorme:
> Or does all data in AcclimationModel need to be subdaily -- I suppose not?

The SubdailyPModel and AcclimationModel are basically paired - I don't think we should allow the AcclimationModel to create inputs that SubdailyPModel will reject, so it _is_ probably better placed in AcclimationModel.

It's also marginally more convenient since there is already one check on the `spacing` attribute:

https://github.com/ImperialCollegeLondon/pyrealm/blob/483976bcbca4a7b4666286605c2902098fab2c77/pyrealm/pmodel/acclimation.py#L222-L223

pyrealm/pmodel/acclimation.py
```diff
@@ -222,6 +222,12 @@ def _validate_and_set_datetimes(self, datetimes: NDArray[np.datetime64]) -> None
         if spacing < 0:
             raise ValueError("Datetime sequence must be increasing")
 
+        # Validate that we are dealing indeed with subdaily data
+        if spacing > np.timedelta64(1, "D"):
+            raise ValueError(
+                "The AcclimationModel data for SubdailyPModel is not subdaily."
+            )
+
```
davidorme: Technically, we allow daily data so I think this is:

```suggestion
        # Validate that we are dealing indeed with subdaily data
        if spacing > np.timedelta64(1, "D"):
            raise ValueError(
                "The AcclimationModel requires datetimes sampled at daily frequency or faster."
            )

```

davidorme (approved):
LGTM - can we add a test?

MarionBWeinzierl:
> LGTM - can we add a test?

I added a test now

PR #496, opened 2025-05-15
Broadcastable array inputs

### LINKED ISSUE #494: Allow broadcastable array inputs
Currently array inputs are required by `check_input_shapes` to fully match shapes. It would be better if broadcastable arrays were accepted instead (satisfy broadcast shapes)

Ideally all internal operations will work with broadcastable arrays. Any manual loops over axes (eg. time) will need to be modified.

A possible simpler approach / stepping stone could be to broadcast the input arrays and store those. `check_array_shapes` cannot (currently) do this, but a similar approach will be needed for the xarray inputs (#43).

sjavis:
# Description

I have modified the check_input_shapes function to allow broadcastable shapes (although more of the function could possibly be replaced by `np.broadcast_arrays`). Tests are still needed on the functions that use this to ensure they are indeed compatible with broadcastable arrays.

Fixes #494

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/core/time_series.py
+++ b/pyrealm/core/time_series.py
@@ -16,7 +16,7 @@ class AnnualValueCalculator:
 
     This class is used to calculate annual means and totals from time series data. An
     instance is created by providing a set of timings for the times series data, either
-    as an an array of datetimes or as an AcclimationModel instance from a
+    as a one-dimensional array of datetimes or as an AcclimationModel instance from a
     SubdailyPModel, which provides validated datetimes at subdaily temporal resolutions.
 
     The calculation process accounts for observations that span year boundaries, such as
@@ -100,11 +100,12 @@ class AnnualValueCalculator:
             or (
                 isinstance(timing, np.ndarray)
                 and np.issubdtype(timing.dtype, np.datetime64)
+                and timing.ndim == 1
             )
         ):
             raise ValueError(
                 "The timings argument must be an AcclimationModel "
-                "or an array of datetime64 values"
+                "or a one-dimensional array of datetime64 values"
             )
 
         if isinstance(timing, AcclimationModel):
@@ -316,6 +317,12 @@ class AnnualValueCalculator:
         else:
             weights = self.duration_weights
 
+        # Make weights broadcastable in case vals is multidimensional
+        if values.ndim > 1:
+            for i, wghts in enumerate(weights):
+                shape = (wghts.shape[0],) + (1,) * (values.ndim - 1)
+                weights[i] = wghts.reshape(shape)
+
         # Calculate the weighted mean in a np.nan friendly way: the product of np.nan
         # and a weight is np.nan and the isnan term omits the weights of nan
         # observations from the weighted average.
@@ -376,6 +383,12 @@ class AnnualValueCalculator:
             ]
         else:
             weights = self.fractional_weights
+
+        # Make weights broadcastable in case vals is multidimensional
+        if values.ndim > 1:
+            for i, wghts in enumerate(weights):
+                shape = (wghts.shape[0],) + (1,) * (values.ndim - 1)
+                weights[i] = wghts.reshape(shape)
 
         return np.array(
             [np.nansum(vals * wghts) for vals, wghts in zip(values_by_year, weights)]
--- a/pyrealm/core/time_series.py
+++ b/pyrealm/core/time_series.py
@@ -33,7 +33,7 @@ class AnnualValueCalculator:
     the ``year_completeness`` attribute records what fraction of a year has been sampled
     to give a particular value.
 
-    .. Note:
+    .. Note::
 
         The class handles a wide range of different possible sampling frequencies and
         calculates weights for observations using the duration of observations with
@@ -289,6 +289,10 @@ class AnnualValueCalculator:
         the observations marked as the growing season are set to zero. The calculation
         handles missing values.
 
+        The annual means can be computed for a range of different sites by using n-D
+        array for ``values``. In this case time is assumed along axis 0 and so an n-D
+        array will be returned with annual values along axis 0.
+
         Example:
             >>> # Three years of monthly data
             >>> datetimes = np.arange(
@@ -330,6 +334,7 @@ class AnnualValueCalculator:
         # Calculate the weighted mean in a np.nan friendly way: the product of np.nan
         # and a weight is np.nan and the isnan term omits the weights of nan
         # observations from the weighted average.
+        # The mean is computed along just the time (0) axis.
         return np.array(
             [
                 np.nansum(vals * wghts, axis=0)
@@ -351,6 +356,10 @@ class AnnualValueCalculator:
         ``True``, the weights for observations not identified as growing season values
         are set to zero. The method handles missing data (`np.nan`) but obviously the
         resulting annual total will be reduced.
+
+        The annual totals can be computed for a range of different sites by using an n-D
+        array for ``values``. In this case time is assumed along axis 0 and so an n-D
+        array will be returned with annual values along axis 0.
 
         Example:
             >>> # Three years of monthly data with incomplete years at start and end
@@ -395,6 +404,7 @@ class AnnualValueCalculator:
                 shape = (wghts.shape[0],) + (1,) * (values.ndim - 1)
                 weights[i] = wghts.reshape(shape)
 
+        # The total is computed along just the time (0) axis.
         return np.array(
             [
                 np.nansum(vals * wghts, axis=0)
--- a/pyrealm/core/time_series.py
+++ b/pyrealm/core/time_series.py
@@ -1,6 +1,7 @@
 """This module provides general tools for working with time series data. It currently
 provides the :class:`AnnualValueCalculator`, which is used to calculate annual means and
-totals of time series data.
+totals of time series data, and :func:`broadcast_time`, which is used to broadcast
+arrays over the time axis.
 """  # noqa : D205
 
 from itertools import pairwise
@@ -411,3 +412,32 @@ class AnnualValueCalculator:
                 for vals, wghts in zip(values_by_year, weights)
             ]
         )
+
+
+def broadcast_time(values: NDArray, shape: tuple[int, ...]) -> NDArray:
+    """Broadcast an array along the time (zeroth) axis.
+
+    The ``values`` array must be broadcastable to the full shape, however it does not
+    need the full set of dimensions as defined by ``shape``. The returned array will
+    have the full set of dimensions, and be broadcast along just the zeroth axis.
+
+    Example:
+        >>> broadcast_time(np.ones((1,3)), (2,3))
+        array([[1., 1., 1.],
+               [1., 1., 1.]])
+        >>> broadcast_time(np.ones(3), (2,2,3)).shape
+        (2, 1, 3)
+
+    Args:
+        values: The array to broadcast.
+        shape: The full n-dimensional shape, where the first value is the length of the
+            time axis to broadcast over.
+    """
+    if values.ndim > len(shape):
+        raise ValueError("The input array has more dimensions than the broadcast shape")
+    # Get any missing axes
+    full_shape = (1,) * (len(shape) - len(values.shape)) + values.shape
+    # Define the shape to broadcast to
+    bcast_shape = (shape[0], *full_shape[1:])
+    # Return the broadcasted array
+    return np.broadcast_to(values, bcast_shape)
--- a/pyrealm/core/time_series.py
+++ b/pyrealm/core/time_series.py
@@ -264,7 +264,11 @@ class AnnualValueCalculator:
             values: An array of values.
         """
 
-        if values.shape[0] != self.n_obs:
+        if values.shape[0] == 1:
+            # Broadcast to match the number of observations if constant
+            values = np.broadcast_to(values, (self.n_obs, *values.shape[1:]))
+
+        elif values.shape[0] != self.n_obs:
             raise ValueError(
                 "First axis of values shape does not match number of observations."
             )
@@ -328,7 +332,8 @@ class AnnualValueCalculator:
         # observations from the weighted average.
         return np.array(
             [
-                np.nansum(vals * wghts) / np.nansum(~np.isnan(vals) * wghts)
+                np.nansum(vals * wghts, axis=0)
+                / np.nansum(~np.isnan(vals) * wghts, axis=0)
                 for vals, wghts in zip(values_by_year, weights)
             ]
         )
@@ -391,5 +396,8 @@ class AnnualValueCalculator:
                 weights[i] = wghts.reshape(shape)
 
         return np.array(
-            [np.nansum(vals * wghts) for vals, wghts in zip(values_by_year, weights)]
+            [
+                np.nansum(vals * wghts, axis=0)
+                for vals, wghts in zip(values_by_year, weights)
+            ]
         )
--- a/pyrealm/core/utilities.py
+++ b/pyrealm/core/utilities.py
@@ -17,9 +17,10 @@ from numpy.typing import NDArray
 def check_input_shapes(*args: float | int | np.generic | np.ndarray | None) -> tuple:
     """Check sets of input variables have congruent shapes.
 
-    This helper function validates inputs to check that they are either scalars or
-    arrays and then that any arrays of the same shape. It returns a tuple of the common
-    shape of the arguments, which is (1,) if all the arguments are scalar.
+    This helper function validates inputs to check that they are either scalars
+    or arrays and that any arrays are of (or broadcastable to) the same
+    shape. It returns a tuple of the common shape of the arguments, which is
+    (1,) if all the arguments are scalar.
 
     Parameters:
         *args: A set of numpy arrays or scalar values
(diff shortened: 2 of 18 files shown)


sjavis:
Force pushed to pick up changes on develop (twice because I forgot to rebase again before pushing).

sjavis:
This now seems to be working fully. The main changes are:
- Change `check_input_shapes` to not require the same shapes, but instead allow broadcastable inputs with the same number of dimensions. I also added a `shape` argument to more clearly show when a specific shape is expected, rather than using a previously defined variable as such variables may no longer have the full shape.
- The `test_broadcasting` test and associated functions defined in `utils.py`. This iterates over all functions and methods in the package using a range of different input shapes for array arguments. This then compares the function results and class attributes for methods to ensure they are the same (aside from array shapes) as when using fully broadcasted inputs.
- These tests are added to the git workflow but kept out of the coverage check. Are we happy running these every time along with the unit tests?

There are also a number of fixes where the code required the full shapes (mostly due to indexing, especially along time). I will annotate the code to explain each of these. These are:
- Splash model - `SplashModel`, `DailySolarFluxes`, `DailyEvapFluxes`
- `AnnualValueCalculator`
- `SubdailyPModel`
- `C3C4Competition`

Most of the code does not rely upon inputs containing all of the dimensions. So, as a possible extension, there could be a second check inputs function that just ensures the inputs are broadcastable. Then each function could use the relevant validation. However, I have not done this here.

pyrealm/core/time_series.py
```diff
@@ -263,7 +264,11 @@ def _split_values_by_year(
             values: An array of values.
         """
 
-        if values.shape[0] != self.n_obs:
+        if values.shape[0] == 1:
+            # Broadcast to match the number of observations if constant
+            values = np.broadcast_to(values, (self.n_obs, *values.shape[1:]))
+
+        elif values.shape[0] != self.n_obs:
```
sjavis, line 272: Broadcast over time needed due to the indexing on line 277.

pyrealm/core/time_series.py
```diff
@@ -377,6 +389,15 @@ def get_annual_totals(
...
+            for i, wghts in enumerate(weights):
+                shape = (wghts.shape[0],) + (1,) * (values.ndim - 1)
+                weights[i] = wghts.reshape(shape)
+
         return np.array(
-            [np.nansum(vals * wghts) for vals, wghts in zip(values_by_year, weights)]
+            [
+                np.nansum(vals * wghts, axis=0)
+                for vals, wghts in zip(values_by_year, weights)
+            ]
```
sjavis, line 413: `get_annual_means` and `get_annual_totals` both implicitly assumed that `values` was 1D. I have made the weights take the same number of dimensions as `values` and only performed the reduction on the time axis. This should allow multiple sites.

pyrealm/pmodel/competition.py
```diff
@@ -226,9 +226,11 @@ def __init__(
         # Step 4: remove areas below minimum temperature
         # mypy - this is a short term fix awaiting better resolution of mixed scalar and
         #        array inputs.
+        below_t_min = np.broadcast_to(below_t_min, self.shape)
         frac_c4[below_t_min] = 0  # type: ignore
 
         # Step 5: remove cropland areas
+        cropland = np.broadcast_to(cropland, self.shape)
         frac_c4[cropland] = np.nan  # type: ignore
```
sjavis, line 234: `below_t_min` and `cropland` need to be the full shape for the indexing.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -127,7 +127,7 @@ def __init__(
         self._check_shapes()
 
         # Make sure the aridity index is not zero
-        if aridity_index <= 0:
+        if np.any(aridity_index <= 0):
```
sjavis, line 130: Necessary if aridity_index is an array.

pyrealm/pmodel/pmodel.py
```diff
@@ -705,10 +707,11 @@ def _fit_model(
...
         self.acclim_model = acclim_model
 
-        if self.acclim_model.datetimes.shape[0] != self.env.tc.shape[0]:
+        datetimes = self.acclim_model.datetimes
+        if self.shape[0] not in (datetimes.shape[0], 1):
             raise ValueError(
-                "The PModelEnvironment data and AcclimationModel datetimes "
-                "are of different lengths."
+                "The first dimension of the PModelEnvironment data is a different "
+                "length to the number of AcclimationModel datetimes."
```
sjavis, line 714: This is to allow inputs that are constant in time.

pyrealm/pmodel/pmodel.py
```diff
@@ -779,8 +782,15 @@ def _fit_model(
...
+                # Broadcast the first dimension (if constant) since the acclimation
+                # model requires this dimension to be equal to the number of times
+                env_var_shape = (1,) * (len(self.shape) - env_var.ndim) + env_var.shape
+                env_var_bcast = np.broadcast_to(
+                    env_var,
+                    [datetimes.shape[0], *env_var_shape[1:]],
+                )
                 daily_environment[env_var_name] = self.acclim_model.get_daily_means(
-                    values=env_var,
+                    values=env_var_bcast,
```
sjavis, line 793: `get_daily_means` needs each of the variables to be broadcast along the time dimension.

pyrealm/splash/evap.py
```diff
@@ -79,6 +79,18 @@ def __post_init__(self, pa: NDArray[np.float64], tc: NDArray[np.float64]) -> Non
...
+
+        # Broadcast along the time axis (necessary for the indexing in estimate_aet)
+        def bcast_time(var: NDArray) -> NDArray:
+            full_shape = (1,) * (len(self.shape) - len(var.shape)) + var.shape
+            bcast_shape = (self.shape[0], *full_shape[1:])
+            return np.broadcast_to(var, bcast_shape)
+
+        pa = bcast_time(pa)
+        tc = bcast_time(tc)
+
```
sjavis, line 96: The variables need to be broadcast over time for the daily indexing in `estimate_aet`. So I do it now rather than later. This is a little more general than necessary because it assumes that the variables may not have the full number of dimensions, which is being enforced by `check_input_shapes'. I have left it in case we want to relax this in future.

pyrealm/splash/solar.py
```diff
@@ -92,10 +92,17 @@ def __post_init__(
         """Populates key fluxes from input variables."""
 
         # Validate the inputs
-        shapes = check_input_shapes(latitude, elevation, sunshine_fraction, temperature)
-        if len(self.dates) != shapes[0]:
+        self.shape: tuple = check_input_shapes(
+            latitude, elevation, sunshine_fraction, temperature
+        )
+        """The array shape of the input variables"""
```
sjavis, line 98: I make this an attribute so it can be used in `check_input_shapes` by `DailyEvapFluxes`.

pyrealm/splash/splash.py
```diff
@@ -69,14 +69,31 @@ def __init__(
...
+        def bcast_time(var: NDArray) -> NDArray:
+            full_shape = (1,) * (len(self.shape) - len(var.shape)) + var.shape
+            bcast_shape = (self.shape[0], *full_shape[1:])
+            return np.broadcast_to(var, bcast_shape)
+
+        elv = bcast_time(elv)
+        lat = bcast_time(lat)
+        sf = bcast_time(sf)
+        tc = bcast_time(tc)
+        pn = bcast_time(pn)
```
sjavis: As with `DailyEvapFluxes` I broadcast the inputs over time since the time axis is indexed in other methods.

pyrealm/core/time_series.py
MarionBWeinzierl: Maybe it is worthwhile to add this explanation as a comment to the code.

pyrealm/demography/canopy.py
```diff
@@ -305,8 +305,8 @@ class CommunityCanopyData(PandasExporter):
     ground below the canopy is stored as the `transmission_to_ground` attribute.
 
     Args:
-        layer_absorption: The expected light absorption for cohorts within each layer.
-        layer_leaf_area_index: The leaf area index of cohorts within layers.
+        absorption: The expected light absorption for cohorts within each layer.
```
MarionBWeinzierl, line 308: Why are you removing "layer"?

pyrealm/demography/flora.py
```diff
@@ -275,7 +275,7 @@ class Flora(PandasExporter):
     """
 
     # The only init argument.
-    pfts: InitVar[Sequence[type[PlantFunctionalTypeStrict]]]
+    pfts: InitVar[Sequence[PlantFunctionalTypeStrict]]
```
MarionBWeinzierl, line 278: Is type no longer needed?

pyrealm/demography/canopy.py
sjavis: I think this was an error when these variables were added in 01058b1. I changed the names in the docstring to match the variable names.

pyrealm/demography/flora.py
sjavis: I could be wrong about this, but I think the type hint wasn't correct and pfts is supposed to be a sequence of instances rather than a sequence of types. It was causing some issues when test_broadcasting was attempting to define it.

davidorme: I think that's right - I cannot recall why I did this.

pyrealm/core/time_series.py
davidorme: I think we should make this part of the docstring: time is assumed along axis 0 and so n-D arrays will come back nD with annual values along axis 0.

pyrealm/core/utilities.py
```diff
@@ -14,35 +14,46 @@
...
+        >>> check_input_shapes(np.ones((1,2,3)), np.ones((2,3)))
+        Traceback (most recent call last):
+        ...
+        ValueError: Inputs contain arrays of different dimensions.
     """
 
     # Collect the shapes of the inputs
     shapes = set()
+    if shape:
+        shapes.add(shape)
```
davidorme, line 56: I'm not convinced by the new `shape` argument here. I think the implementation is only used in testing and it isn't clear to me how we'd use it in real code. There are definitely places where we could need to test if shapes are the same, but I think an explicit test within calling code is going to give a clearer error than adding that expected shape into.

pyrealm/pmodel/competition.py
davidorme: Ah - ok. The `check_input_shapes` call above checks they _are_ broadcastable, but if they aren't actually broadcast then the indexing doesn't work?

pyrealm/splash/evap.py
```diff
@@ -135,10 +147,10 @@ def estimate_aet(
         # subset the calculations to particular request days or use the entire array of
         # soil moisture. The slice here is used to programatically select `array[:]`.
         if day_idx is None:
-            check_input_shapes(wn, self.sat)
-            didx: int | slice = slice(self.sat.shape[0])
+            check_input_shapes(wn, shape=self.shape)
+            didx: int | slice = slice(self.shape[0])
```
davidorme: Ah. OK. This is the use case of the shape argument.  Hmmm...

I think the previous code was a bit of a lazy hack on my part - it's using `check_input_shapes` purely to raise an error. I think my comment on the suggested changes to `check_input_shapes` probably holds. I think it is cleaner here to do something like:

```suggestion
            if not wn.shape == self.shape:
                raise ValueError("The shape of wn does not match the existing SPLASH model data")
            didx: int | slice = slice(self.shape[0])
```

And then similar below. I think trying to dual purpose the function feels like it isn't keeping concerns separated?

pyrealm/splash/splash.py
```diff
@@ -69,14 +69,31 @@ def __init__(
...
+                "The first dimension of inputs must either match the number of dates or"
+                " have a length of one."
+            )
+
+        # Broadcast all the inputs over time to simplify the daily indexing if any
+        # inputs are constant over time
+        def bcast_time(var: NDArray) -> NDArray:
+            full_shape = (1,) * (len(self.shape) - len(var.shape)) + var.shape
+            bcast_shape = (self.shape[0], *full_shape[1:])
+            return np.broadcast_to(var, bcast_shape)
```
davidorme: This is used as an internal static method in two modules. Can we move this into `pyrealm.core.time_series` (or possibly `utilities`) and add the `shape` argument and then import it in both locations.

pyrealm/splash/splash.py
```diff
@@ -266,10 +283,10 @@ def estimate_daily_water_balance(
         # Check day_idx inputs to map either the single time index given in day_idx or
         # the whole dataset.
         if day_idx is None:
-            check_input_shapes(previous_wn, self.pn)
+            check_input_shapes(previous_wn, shape=self.shape)
             didx: int | slice = slice(self.shape[0])
```
davidorme: Same as above, I think - this would be much better as an explicit test and raise, rather my original subversion of the `check_input_shapes` process.

tests/broadcasting/test_broadcasting.py
```diff
@@ -0,0 +1,94 @@
...
+
+shape_full: list[tuple[int, ...]]
+shape_full = [(3, 2, 2)]
+shapes_list = [
+    [(3, 2, 2), (1, 2, 2), (3, 1, 1), (1, 1, 1)],
+    [(1, 2, 2), (3, 2, 2)],
+    [(3, 1, 1), (1, 2, 2), (3, 2, 2)],
+    [(1, 1, 1)],
+]
+method_list = get_method_list()
```
davidorme: Global constants are typically uppercased (https://peps.python.org/pep-0008/#constants) - I think that is a useful convention for following where values come from.

```suggestion
SHAPE_FULL: list[tuple[int, ...]] = [(3, 2, 2)]
SHAPES_LIST: list[list[tuple[int, ...]]] = [
    [(3, 2, 2), (1, 2, 2), (3, 1, 1), (1, 1, 1)],
    [(1, 2, 2), (3, 2, 2)],
    [(3, 1, 1), (1, 2, 2), (3, 2, 2)],
    [(1, 1, 1)],
]
METHOD_LIST = get_method_list()
```

tests/broadcasting/utils.py
```diff
@@ -0,0 +1,613 @@
...
+from pyrealm.demography.flora import (
+    PlantFunctionalType,
+    PlantFunctionalTypeStrict,
+    StemTraits,
+)
+
+## Lists / functions to manually define arguments, methods or outputs to ignore, etc.
+
+# These methods are not relevant or are incompatible without additional work
+skip_methods = [
```
davidorme: Could you use uppercase for these module constants too?

tests/broadcasting/test_broadcasting.py
```diff
@@ -0,0 +1,94 @@
...
+    # Fail if function outputs not equal
+    if not is_equal(result, result_full):
+        result_comparison = comparison_string(result, result_full)
+        raise ValueError(f"Results do not match in {name} ({result_comparison})")
+
+
+if __name__ == "__main__":
+    for method_info in method_list:
+        for shapes in shapes_list:
+            test_array_input_broadcasting(method_info, shapes)
```
davidorme: Do we need this? We can invoke the individual test as `pytest tests/broadcasting/test_broadcasting.py`.

tests/broadcasting/utils.py
davidorme, line 1: There's a lot in here and the functions are nested - could you doc the workflow and the main function in the docstring here?

davidorme (changes requested):
@sjavis This looks great - nice working getting that package introspection set up. I've got a few things that stick out, but nothing major. The `tests.broadcasting.utils` went slightly over my head, so I might have another look after this stuff is resolved, but the shape of it looks good.

We should probably add the remaining `skip_methods` as an issue.

pyrealm/core/utilities.py
davidorme: OK - I've noted below where this is used for real. I think what I had been doing with the function in those cases was a hack.

pyrealm/core/time_series.py
sjavis: I have added this in 2223f1a0.

pyrealm/pmodel/competition.py
sjavis: Yes, that's right.

pyrealm/splash/splash.py
sjavis: I have added the `broadcast_time` function to `time_series.py` in 906a1ae3.

tests/broadcasting/test_broadcasting.py
sjavis: I was using this for debugging purposes but it is unnecessary so I have removed it.

tests/broadcasting/test_broadcasting.py
sjavis: Agreed. I have updated this.

tests/broadcasting/utils.py
sjavis: Done in ced0ac3c.

tests/broadcasting/utils.py
sjavis: Done!

pyrealm/splash/evap.py
sjavis: I agree about the separation of concerns regarding `check_input_shapes`, but I would be inclined to leave this from a practical point of view:

We can't compare `wn.shape` (or equivalents where this is used elsewhere) to `self.shape` directly because `wn` may not be fully broadcast, and we can't use `np.broadcast_shapes` because we require the same number of dimensions. So it would probably be best to define a new function to compare the shape.

But this function would necessarily be very similar to `check_input_shapes`, at which point I think it is simpler to instead use that function directly. What are your thoughts on this?

davidorme: @sjavis I see what you mean. And there is the additional rationale here that switching from `self.sat` to `self.shape` here is that `self.shape` isn't a single variable - it's the maximal size across all dimensions. And that is more robust - but it's a tuple of sizes, not an array so has to be passed in a different way.

OK - yup. Sold. Maybe in these use cases we wrap the `check_input_shapes` in a try block to provide more informative feedback in failure cases.

sjavis: Agreed, I will add some try blocks.

sjavis:
Rebased to pick up changes in develop

pyrealm/splash/evap.py
sjavis: I have added these in 7bdcdd0

sjavis:
I believe all of the comments have been addressed and I have added an issue to look at reducing the `skip_methods` list (#537).

davidorme (approved):
LGTM - thanks for the updates.

PR #497, opened 2025-05-15
xarray inputs in pmodel

### LINKED ISSUE #630: `xarray` inputs to `pmodel` module
The array inputs should be allowed to be passed as `xarray.DataArray`s in the `pmodel` module functions. With the dimension names being used to add missing dimensions.

sjavis:
# Description

Added a function to convert xarray input arguments to numpy arrays. Still need to add this to functions and check it is working properly.

Fixes #630

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.github/workflows/pyrealm_ci.yaml
+++ b/.github/workflows/pyrealm_ci.yaml
@@ -44,10 +44,10 @@ jobs:
       run: poetry install
 
     - name: Run tests
-      run: poetry run pytest -m "not profiling and not broadcasting" --cov-report xml
+      run: poetry run pytest -m "not profiling and not array_inputs" --cov-report xml
 
-    - name: Run broadcasting tests
-      run: poetry run pytest -m "broadcasting"
+    - name: Run array input tests
+      run: poetry run pytest -m "array_inputs"
 
     - name: Upload coverage reports to Codecov
       id: codecov
--- a/docs/source/api/core_api.md
+++ b/docs/source/api/core_api.md
@@ -101,3 +101,11 @@ language_info:
     :autosummary:
     :members:
 ```
+
+## The {mod}`~pyrealm.core.xarray` submodule
+
+```{eval-rst}
+.. automodule:: pyrealm.core.xarray
+    :autosummary:
+    :members:
+```
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -175,6 +175,10 @@ nitpick_ignore = [
     ),
     ("py:class", "pandas.core.frame.DataFrame"),
 ]
+# Ignore TypeVars (anything named 'T')
+nitpick_ignore_regex = [
+    ("py:class", r".*\.T"),
+]
 
 intersphinx_mapping = {
     "pytest": ("https://docs.pytest.org/en/stable/", None),
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -121,6 +121,7 @@ nitpick_ignore = [
     ("py:class", "numpy._typing._generic_alias.ScalarType"),
     ("py:class", "numpy.float32"),
     ("py:class", "numpy.float64"),
+    ("py:class", "numpy.floating"),
     ("py:class", "numpy.int64"),
     ("py:class", "numpy.timedelta64"),
     ("py:class", "numpy.bool_"),
@@ -175,9 +176,12 @@ nitpick_ignore = [
     ),
     ("py:class", "pandas.core.frame.DataFrame"),
 ]
-# Ignore TypeVars (anything named 'T')
 nitpick_ignore_regex = [
+    # Ignore TypeVars (anything named 'T')
+    ("py:class", "T"),
     ("py:class", r".*\.T"),
+    # Ignore non-expanded 'np' (in ArrayType parameters)
+    ("py:class", r"np\..*"),
 ]
 
 intersphinx_mapping = {
@@ -210,6 +214,11 @@ mathjax3_config = {
 # Turn off ugly rendering of class attributes
 napoleon_use_ivar = True
 napoleon_custom_sections = [("PModel Parameters", "params_style")]
+
+# Ensure type aliases aren't expanded
+autodoc_type_aliases = {
+    "ArrayType": "pyrealm.core.xarray.ArrayType",
+}
 
 # Autodoc configuration:
 # - Suppress signature expansion of arguments
@@ -277,6 +286,9 @@ def setup(app):  # type: ignore
 
     app.connect("build-finished", strip_jupytext)
 
+    # Resolve cross references for type aliases
+    app.connect("missing-reference", resolve_alias_fallback)
+
 
 def strip_jupytext(app, exception):  # type: ignore
     """Remove jupytext metadata from download notebooks.
@@ -335,3 +347,31 @@ def strip_jupytext(app, exception):  # type: ignore
 
     print(f"{bold_start_text} {converted} of {len(ipynb_downloads)} .ipynb files.")
     return
+
+
+def resolve_alias_fallback(app, env, node, contnode):  # type: ignore
+    """Resolve failing type alias cross-references by using py:data instead of py:class.
+
+    This is necessary for type aliases which autodoc creates with py:data but it looks
+    for with py:class.
+
+    This should be resolved in sphinx 9.0.
+    Relevant issue: https://github.com/sphinx-doc/sphinx/issues/10785
+    """
+
+    # If py:class
+    if node["refdomain"] == "py" and node["reftype"] == "class":
+        target = node["reftarget"]
+        # Attempt to resolve the cross reference
+        return env.domains["py"].resolve_xref(
+            env,
+            node["refdoc"],
+            app.builder,
+            "data",  # Changed from 'class'
+            target,
+            node,
+            contnode,
+        )
+
+    else:
+        return None
--- a/docs/source/users/array_inputs.md
+++ b/docs/source/users/array_inputs.md
@@ -121,3 +121,26 @@ print("patm shape:", patm.shape)
 # Call PModelEnvironment with all the inputs
 env = PModelEnvironment(tc=temp, co2=co2, patm=patm, vpd=vpd, fapar=fapar, ppfd=ppfd)
 ```
+
+## Xarray inputs
+
+```{note}
+This is currently limited to `pmodel` module only.
+```
+
+Array inputs to functions will also accept `xarray.DataArray` inputs. These are mostly
+equivalent to `numpy` arrays but with the addition of named dimensions. `pyrealm` will
+internally convert these into `numpy` arrays. So all outputs will also be `numpy`
+arrays.
+
+The `pyrealm` function will use the dimension names to add any missing dimensions and to
+reorder them if inputs have different orders. The final order is decided by iterating
+through each input and appending any missing dimensions.
+
+For example, if there are three `xarray.DataArray` inputs with the following dimension
+names: `("time")`, `("lat", "lon")`, `("lon", "lat")`. Then they will all be converted
+to the shape `("time", "lat", "lon")`. With missing dimensions given a size of 1.
+
+If there is a combination of `xarray` and `numpy` inputs. Then the `numpy` arrays must
+match the full set of dimensions given by the `xarray` inputs with the same rules [given
+above](#array-shapes).
--- a/docs/source/users/pmodel/pmodel_details/worked_examples.md
+++ b/docs/source/users/pmodel/pmodel_details/worked_examples.md
@@ -35,10 +35,10 @@ the P Model.
 ```
 
 The first example uses a single point but the second shows how the package can be used
-with array data. The `pyrealm` package uses the `numpy` package and expects arrays of
-data to be be passed to all inputs. Input arrays can be a single scalar value, but all
-non-scalar inputs must be **arrays with the same shape**: the `pyrealm` packages does
-not attempt to resolve the broadcasting of array dimensions.
+with array data. The `pyrealm` package uses the `numpy` package and expects inputs to
+be passed either as single scalar values, or as arrays with shapes that are consistent
+under broadcasting rules. See the page on [array inputs](../../array_inputs.md) for
+details.
 
 ```{code-cell} ipython3
 from importlib import resources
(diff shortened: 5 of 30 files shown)


sjavis:
I have modified the `xarray_inputs` decorator to also expand out any dimensions according to the dimension names for xarray input variables. Eg. if `array1`, `array2`, and `array3` are passed with dimensions `('x', 'y')`, `('t')`, and `('x', 't')`. Then they will have the missing dimensions added to become: `('x', 'y', 1)`, `(1, 1, 't')`, `('x', 1, 't')`.

Unfortunately there is a problem with the type hinting for the decorator approach. The way it is currently implemented is that the decorator takes any `xarray.DataArray` inputs and then passes these as numpy arrays to the underlying function. The decorator therefore needs to change the function signature so that numpy array arguments can also be passed as DataArrays. However, there is seemingly no way to iterate over the parameters to change them dynamically in this way.

<details>
The basic code for a decorator that keeps the signature of the underlying function is:

```python
P = ParamSpec('P')
T = TypeVar('T')

def decorator(fn: Callable[P, T]) -> Callable[P, T]:
    @functools.wraps(fn)
    def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
        # Strip xarray inputs
        return fn(*args, **kwargs)
    return wrapper
```

For `xarray_inputs` the function return type would need a different `ParamSpec` based on `P` but with modified parameter types (Something like `def decorator(fn: Callable[P, T]) -> Callable[Modify[P], T]:`). Parameters can be added or removed with `Concatenate` and positional arguments with a known position can be modified. But it doesn't seem possible to iterate over the parameters to modify them — unless you modify the signature at runtime, but this isn't that helpful.
</details>

Therefore the options are to:

(a) Make the argument types of the wrapped function `Any`. In which case it will not be possible to do type checking with any functions that accept xarray inputs. However the return types will still be accurate and the types in the documentation still seem to be correct.

(b) Change the type hints of the various functions to explicitly accept `xarray.DataArray` (or `npt.ArrayLike`). To avoid mypy errors this would require explicitly converting to numpy inside the functions, in which case it is probably easiest not to use a decorator at all.

davidorme:
> I have modified the `xarray_inputs` decorator to also expand out any dimensions according to the dimension names for xarray input variables. Eg. if `array1`, `array2`, and `array3` are passed with dimensions `('x', 'y')`, `('t')`, and `('x', 't')`. Then they will have the missing dimensions added to become: `('x', 'y', 1)`, `(1, 1, 't')`, `('x', 1, 't')`.

Therefore making them broadcastable across the dimensions, right? Good call!

davidorme:
> Unfortunately there is a problem with the type hinting for the decorator approach. The way it is currently implemented is that the decorator takes any `xarray.DataArray` inputs and then passes these as numpy arrays to the underlying function. The decorator therefore needs to change the function signature so that numpy array arguments can also be passed as DataArrays. However, there is seemingly no way to iterate over the parameters to change them dynamically in this way.
> 
> Therefore the options are to:
> 
> (a) Make the argument types of the wrapped function `Any`. In which case it will not be possible to do type checking with any functions that accept xarray inputs. However the return types will still be accurate and the types in the documentation still seem to be correct.
> 
> (b) Change the type hints of the various functions to explicitly accept `xarray.DataArray` (or `npt.ArrayLike`). To avoid mypy errors this would require explicitly converting to numpy inside the functions, in which case it is probably easiest not to use a decorator at all.

Well, bugger. I don't like this outcome much:

> it will not be possible to do type checking with any functions that accept xarray inputs

For all of the irritation of using mypy to type check, I think alerting us to issues with mismatched types on data inputs is pretty critical. My instinct is that this would seriously impact the maintainability of the code. Thoughts?

So, if I'm ruling out (a) then:

* We need to type the array inputs differently in all functions accepting arrays. It looks like `numpy.typing.ArrayLike` is fine (because `xarray.DataArray` implements the `__array__` protocol). @j-emberton Went for a more complex solution in #272 by building up a composite type using `array_api_compat`: we might want to do that at some point, but for now I think the `ArrayLike` is simpler...

*  **EXCEPT** that it looks like `ArrayLike` does not support declaration of `dtype` as `NDArray[np.floating]` does. I think that might be a huge problem for typing because  e.g. `ArrayType + ArrayType` has a whole bunch of possible types that don't support `+`.  

* I agree that a decorator no longer makes sense. So we then need an internal function that converts the array inputs. I can't currently figure out a way to update values of arguments in place rather than having to reassign. It would be nice to avoid having to do something like:

```python
array_one, array_two = convert(array_one, array_two)
```

and just use

```python
convert(array_one, array_two)
```

*  It might be useful to have explicit metadata on array arguments? Something like:

```python

def set_array_arguments(array_arguments: tuple[str, ...]):
    def _callable(func):
        func.array_arguments = array_args
        return func

    return _callable

@set_array_arguments(("a", "b"))
def testing(a: ArrayLike, b: ArrayLike, add: bool = True) -> ArrayLike:
    
    convert(testing.array_arguments)

    if add:
        return a + b

    return a - b
```

sjavis:
> For all of the irritation of using mypy to type check, I think alerting us to issues with mismatched types on data inputs is pretty critical. My instinct is that this would seriously impact the maintainability of the code. Thoughts?

I think it wouldn't necessarily be too bad if we only allow xarray inputs on select entry points to the code. Then most of the internal type checking would remain. On the other hand, if we limit the number of places where it is used then there isn't much reason for using a decorator anyway.

> We need to type the array inputs differently in all functions accepting arrays. It looks like numpy.typing.ArrayLike is fine (because xarray.DataArray implements the __array__ protocol). @j-emberton Went for a more complex solution in
>
> - Implement array interoperability across numpy, dask and cupy #272 by building up a composite type using array_api_compat: we might want to do that at some point, but for now I think the ArrayLike is simpler...
>
> - EXCEPT that it looks like ArrayLike does not support declaration of dtype as NDArray[np.floating] does. I think that might be a huge problem for typing because e.g. ArrayType + ArrayType has a whole bunch of possible types that don't support +.

I think this is OK if we just use ArrayLike for the inputs arguments and explicitly convert them to numpy arrays immediately. This would allow us to still use the array API for the internals down the line.

> - I agree that a decorator no longer makes sense. So we then need an internal function that converts the array inputs. I can't currently figure out a way to update values of arguments in place rather than having to reassign. It would be nice to avoid having to do something like:
> ```python
> array_one, array_two = convert(array_one, array_two)
> ```
> and just use
> ```python
> convert(array_one, array_two)
> ```

Unfortunately we would need to use the former approach. Even though we can change the array contents inside the function, python doesn't allow the variable itself to be changed. I think the simplest solution would be to replace the decorator with this conversion function.


> - It might be useful to have explicit metadata on array arguments? Something like:
> ```python
> def set_array_arguments(array_arguments: tuple[str, ...]):
>     def _callable(func):
>         func.array_arguments = array_args
>         return func
>
>     return _callable
>
> @set_array_arguments(("a", "b"))
> def testing(a: ArrayLike, b: ArrayLike, add: bool = True) -> ArrayLike:
>    
>     convert(testing.array_arguments)
>
>     if add:
>         return a + b
>
>    return a - b
> ```

Unless I misunderstand this, I think using metadata like this would only make it more complicated. Since the metadata only stores the names of the array arguments it would be necessary to use `inspect.signature` and iterate over the parameters.

davidorme:
> > * EXCEPT that it looks like ArrayLike does not support declaration of dtype as NDArray[np.floating] does. I think that might be a huge problem for typing because e.g. ArrayType + ArrayType has a whole bunch of possible types that don't support +.
> 
> I think this is OK if we just use ArrayLike for the inputs arguments and explicitly convert them to numpy arrays immediately. This would allow us to still use the array API for the internals down the line.
> 

I think the thing we'd lose is that it would no longer be a `mypy` error if one function calls another passing an array with different dtype to  to a `NDArray[np.floating]` argument. For example, this code is `mypy` error free:

```python
def array_function(a: ArrayLike) -> NDArray[np.floating]:
    """Testing typing."""
    array_np: NDArray[np.floating] = np.array(a)
    return array_np

def using_function() -> NDArray:
    """Testing calling."""
    x: NDArray[np.str_] = np.array(["a", "b", "c"])
    x_np = array_function(x)
    return x_np * x_np
```

but raises:

```python
UFuncTypeError: ufunc 'multiply' did not contain a loop with signature matching types (dtype('<U1'), dtype('<U1')) -> None
```

I don't think we can (easily) do anything about this and use `xarray` though. 

> Unfortunately we would need to use the former approach. Even though we can change the array contents inside the function, python doesn't allow the variable itself to be changed. I think the simplest solution would be to replace the decorator with this conversion function.

That was my conclusion too. The explicit reassignment is also somewhat clearer within the function. I think we're going to have to update the docstrings - at the moment I'm using the typing to show the expected dtypes of array arguments, but we'll lose that with `ArrayLike`. We also have a slightly fussy problem if we want to retain the subtyping of NDArray arguments within the function - I think that's useful information, but `mypy` complains about variable redefinition in the example below. We'd have to rename input arrays internally (or turn off the `mypy` `redef` rule locally, which is also clunky). 

```python

def user_old(a:  NDArray[np.floating]) -> None:
    print(a)

def user_new(a: ArrayLike) -> None:
    a: NDArray[np.floating] = converter(a)
    print(a)
```

sjavis:
The best bet might be to use `NDArray[dtype] | xr.DataArray` instead of `ArrayLike`. This allows us to keep the indication of the datatype and lets mypy catch the first problem you gave. It won't catch it for xarray inputs but I don't think there is anything we can do about that.

As for the redefinition problem, I think it works if we leave off the typehint (for whatever reason `x : int | str = 1; x = str(x)` is deemed acceptable but explicitly writing `x : int | str = 1; x : str = str(x)` is not). Instead it would be possible to explicitly pass the dtype to the converter function to make it clear what it is doing and prevent having to make duplicate functions for different dtypes.

So I think the best option probably looks like this:

```python
T = TypeVar("T", bound=np.generic)

def converter(*arrays: NDArray | xr.DataArray, dtype: type[T]) -> list[NDArray[T]]:
    result = []
    for a in arrays:
        if isinstance(a, xr.DataArray):
            result.append(a.asnumpy(dtype=dtype))
        else:
            result.append(a.astype(dtype))
    return result

def array_function(a: NDArray[np.float64] | xr.DataArray) -> NDArray[np.float64]:
    a, = converter(a, dtype=np.float64)
    return a
```

sjavis:
@davidorme It would be good if you could give this a preliminary review to see if this approach looks reasonable, currently I have updated all of the PModel module to allow `xarray.DataArray` inputs, 

The main additions are:

In `core/xarray.py`:
- An `ArrayType` type alias for both numpy arrays and DataArrays.
- A `xarray_inputs` function to convert `ArrayType`s to numpy arrays (and `xarray_inputs_kw` when it also needs to convert a dictionary of inputs). This also expands / reorders dimensions of any DataArrays so that they all have consistent dimensions.
- An `is_arraytype` function which is needed for checks when a variable can also be other types such as scalars. Currently only used in `PModelABC` and `SubdailyPModel`.

In the `pmodel` folder:
- Changed `NDArray` argument types to `ArrayType` and added `xarray_inputs` calls to convert these to numpy arrays.

A couple of minor things to note:
- Two calls to `xarray_inputs` are needed in `C3C4Competition.__init__` because it cannot handle different array dtypes.
- `xarray_inputs_kw` is needed in `PModelEnvironment.__init__` because the kwargs are be `ArrayType`s.
- I have left datetime and bool array inputs in `acclimation.py` as they are.

If it looks reasonable I will continue with the rest of the package. I also plan to use the broadcasting test functionality to ensure that any functions taking `ArrayType`s are consistent whether xarray or numpy inputs are provided, rather than manually adding tests for everything.

davidorme:
> In `core/xarray.py`:
> 
> * An `ArrayType` type alias for both numpy arrays and DataArrays.

Yup - I think this is the cleanest way

> * A `xarray_inputs` function to convert `ArrayType`s to numpy arrays (and `xarray_inputs_kw` when it also needs to convert a dictionary of inputs). This also expands / reorders dimensions of any DataArrays so that they all have consistent dimensions.

I'd be tempted to call this something like `normalise_array_args_to_numpy()` or something similar? It's a little clearer and it would also mean if we did - for example - extend this to allow `pd.Series` then we could just change the internals and not have to change the function name.

Having said that `pd.Series` is a very specific use case where _all_ of the inputs would have to be 1D arrays, so it might be more difficult. 

> * An `is_arraytype` function which is needed for checks when a variable can also be other types such as scalars. Currently only used in `PModelABC` and `SubdailyPModel`.

✅ 

> In the `pmodel` folder:
> 
> * Changed `NDArray` argument types to `ArrayType` and added `xarray_inputs` calls to convert these to numpy arrays.

✅ 
 
> A couple of minor things to note:
> 
> * Two calls to `xarray_inputs` are needed in `C3C4Competition.__init__` because it cannot handle different array dtypes.

I can't immediately see why it doesn't? What's the issue?

> * `xarray_inputs_kw` is needed in `PModelEnvironment.__init__` because the kwargs are be `ArrayType`s.
> * I have left datetime and bool array inputs in `acclimation.py` as they are.
> 

✅ 

> If it looks reasonable I will continue with the rest of the package. I also plan to use the broadcasting test functionality to ensure that any functions taking `ArrayType`s are consistent whether xarray or numpy inputs are provided, rather than manually adding tests for everything.

That sounds like the only sane way of approaching it!

davidorme (approved):
I think we go for it - sorry for all the indecision on this!

sjavis:
I have added the testing of the xarray inputs. By extending the broadcasting testing (now renamed to array_inputs).

Currently the tests are broken because:
- Different input types need to be called multiple times - causing issues with determining the dimensions.
- Some methods need input dimensions consistent with class attributes, not just other arguments.

sjavis:
This should now be ready for a final review. These are the additional changes made since the previous review:

- I have made sure the xarray inputs work properly for "dependent functions", i.e. when the shape must match a previously defined shape. To do this:
  - Store `dims` as an attribute in relevant classes (`PModelEnvironment` and `C3C4Competition`).
  - Added a `get_common_dims` function to get this set of dimension names given the list of inputs.
  - Added a `dims` argument to the `xarray_inputs` conversion function to use these already found dimensions.
  - Made sure the dependent functions were using the `shape` argument of `check_array_inputs` rather than using a class attribute as these may not be broadcasted (e.g. using `shape=self.env.shape` rather than `self.env.tc`)
  
- Updated `xarray_inputs` to also accept keyword arguments, removing the need for a separate `xarray_inputs_kw` function.

- I have added the tests for the xarray inputs to check that it behaves the same as providing numpy arrays. This is in the `array_inputs/test_xarray.py` file. There are also many changes to `array_inputs/utils.py` but these probably do not need as thorough a review.
  - Changed how the shapes of the array arguments are generated. Rather than looping through a list of shapes, this now uses a generating function. For xarray, this will randomly delete and reorder the dimensions but under the constraint that the full set of array inputs will have all of the dimensions in the right order (this constraint is relaxed for dependent functions).
  - Reorganised the logic on how to define each argument (manual, default, automatic) to enable the above.
  - Made the arguments in `REQUIRES` not use default values and added arguments of `estimate_chi` to this list to ensure these are properly tested.
  - Made various fixes to resolve failing tests due to the changes to shapes.

- Updated the documentation to not expand the `ArrayType` alias because it was resulting in messy signatures. This also needed a workaround in the sphinx `conf.py` for these cross references to work.
  - A related question I had here is if it would be good to use `autodoc_typehints="description"` in `conf.py`? I find it can be hard to read the argument types in the signatures if there are many arguments and this might be clearer.

- Updated the worked examples to remove the now unnecessary conversions to numpy.

- Added a section about using xarray inputs to the array inputs page of the documentation. This would be good to review to make sure it is clear.

If the changes seem good then I will rebase onto main to do another regression test to make sure nothing has caused significant changes to the performance. 

There is related work that I haven't done here and will do in other PRs to keep this PR from getting too overwhelming:
- Add xarray inputs to the other modules.
- Reorganise the array inputs tests to make it a bit easier to work with. Mainly, separate the manual overrides, etc into a different module from the utility functions. And also to rework the messy `defined_method_args` function.

davidorme (approved):
LGTM @sjavis - thanks!

I do think it would be better to have the module as `pyrealm.core.array_inputs` not `xarray`, because I think expanding this framework to support `pandas` would be useful. But none of this functionality is user-facing and so I think we can claim that as a non-breaking change if and when  that extension happens.

sjavis:
Yes, I agree with that, especially if we want to add other inputs such as pandas or to combine with the `check_input_shapes`. But as you say this is all internal so I will leave it for now and we can change it later.

sjavis:
This PR doesn't seem to have significantly affected the performance.

```
/home/sa2329/pyrealm/prof/combined-old.prof
Old time: 67.616
/home/sa2329/pyrealm/prof/combined-new.prof
New time: 70.59899999999999
Times haven't changed.                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                        
No significant performance regression detected.
```

PR #503, opened 2025-06-06
Fixing errors in canopy calculation

### LINKED ISSUE #502: Errors in canopy calculation
There are some issues in the `demography` module inside the canopy calculation.

Two obvious ones:

* Negative crown radius: this is actually expected from the form of the crown shape equations when the crown radius is calculated above the stem height or below the crown stem fraction, but it probably should be clipped out rather than assuming downstream code handles the negative values.

* The calculation of LAI within layers is bogus. I think the light transmission factor $e^{-kL}$ should be constant across layers: this model implies that the big leaf disc with "thickness" $L$ is simply being cut up into chunks that occur in different layers. The current implementation does something odd that does not conserve the LAI value from the T Model.

davidorme:
# Description

This PR provides a reworking of the light capture model in the `demography.canopy` module.

I'm pretty sure the previous version is wrong - it certainly gives bad predictions and I can't honestly follow the logic what I previously implemented. The replacement version is cleaner and feels much more intuitive - and it also gives results that agree with a simple simulation of the process of light capture. There's a little bit of circularity in the framing of the simulation and the new approach, but they give a consistent structure.

The PR:

* Updates the crown model to explicitly constrain projected crown area predictions outside of the stem height (<0 or >H) to be zero. Numerically, the equation gives values outside that range but they don't make any sense and were being passed into results.
* Updates the canopy model to use a new calculation of light absorption.
* Updates the tests to test the clipping of crown area and provide more grounded tests of the light capture model.
* Splits the existing `docs/source/users/demography/canopy.md` out into two new files:
    * The updated `canopy.md` just presents the canopy structure calculations, with some new simpler examples that build from simple to more complex communities and parameterisations. 
    * All of the light capture material from the previous `canopy.md` - which I think was mostly wrong - has been revised and put into `light_capture.md`. This presents the theory and then both the simulation and array based calculations of the values, followed by simple demonstrations of the implemented structures in `pyrealm`.

I've built the PR on RTD to make the docs easier to review. The pages are:

https://pyrealm.readthedocs.io/en/502-errors-in-canopy-calculation/users/demography/canopy.html
https://pyrealm.readthedocs.io/en/502-errors-in-canopy-calculation/users/demography/light_capture.html

Fixes #502

## Type of change

## Key checklist

## Further checks


docs/source/users/demography/light_capture.md
```diff
@@ -0,0 +1,520 @@
...
+$$
+\begin{align}
+A_{fl} &= \tilde{A}_{cp}(z_l), \textrm{if} \; l=1 \\
+       &= \tilde{A}_{cp}(z_l) - \tilde{A}_{cp}(z_{l-1}), \textrm{otherwise}
+\end{align}
+$$
+
+```{attention}
+The calculation of $A_{fl}$ uses the **projected leaf area** and not the projected crown
+area: the projected crown area dicates how the crown of a given stem occupies space in
```
MarionBWeinzierl:
```suggestion
area: the projected crown area dictates how the crown of a given stem occupies space in
```

docs/source/users/demography/light_capture.md
```diff
@@ -0,0 +1,520 @@
...
+the canopy actual light capture uses the leaf area, which can be vertically displaced in
```
MarionBWeinzierl: I don't understand this sentence: "the projected crown area dictates how the crown of a given stem occupies space in the canopy actual light capture uses the leaf area" -- should it be "the projected crown area dictates how the crown of a given stem occupies space in the canopy. Actual light capture uses the leaf area " ?

docs/source/users/demography/light_capture.md
```diff
@@ -0,0 +1,520 @@
...
+$$
+
+```{attention}
+Note that the calculation of average transmission requires the intermediate calculation
+of average absorbance in order to account for unoccupied area in the layer. This can
+arise in incompletely closed layers at the bottom of the canopy or where a canopy or
+crown gap fraction is set. Because the absorbance of empty space is zero, the
+calculation of average absorbance implicitly accounts for unfilled space.
+
+The same is not true for transmission, wher unfilled area would need to be included
```
MarionBWeinzierl:
```suggestion
The same is not true for transmission, where unfilled area would need to be included
```

docs/source/users/demography/light_capture.md
```diff
@@ -0,0 +1,520 @@
...
+
+And again check that we get the same solution:
+
+```{code-cell} ipython3
+if np.allclose(simulated_per_stem_f_abs, per_stem_f_abs):
+    print("\U00002705 Per stem fraction absorbed matches")
+```
+
+We can now calculate the total absorption in each layer for each stem. This is the key
+result for use in most cases: **how much light flux is by each stem in the leaf area of
```
MarionBWeinzierl: Not sure about this one, but should it be

```suggestion
result for use in most cases: **how much light flux is absorbed by each stem in the leaf area of
```

?

pyrealm/demography/canopy.py
```diff
@@ -312,61 +293,81 @@ class CommunityCanopyData(PandasExporter):
...
 
     # Calculated variables
-    f_trans: NDArray[np.float64] = field(init=False)
-    """The fraction of light transmitted by the whole community within a layer."""
-    f_abs: NDArray[np.float64] = field(init=False)
-    """The fraction of light absorbed by the whole community within a layer."""
+    average_layer_absorption: NDArray[np.float64] = field(init=False)
+    """The average absorption within layers across the community."""
+    average_layer_lai: NDArray[np.float64] = field(init=False)
+    """The average leaf area index of the the community within layers."""
```
MarionBWeinzierl:
```suggestion
    """The average leaf area index of the community within layers."""
```

pyrealm/demography/crown.py
```diff
@@ -63,8 +69,13 @@ def calculate_relative_crown_radius_at_z(
         )
 
     z_over_height = z / stem_height
+    q_z = m * n * z_over_height ** (n - 1) * (1 - z_over_height**n) ** (m - 1)
+
+    # Set predictions to zero where z is below zero or above the stem height.
+    if clip:
+        q_z = np.where(np.logical_and(z >= 0, z <= stem_height), q_z, 0)
```
MarionBWeinzierl, line 76: Does one of the `<=` need to be `<` ?

tests/unit/demography/test_canopy.py
```diff
@@ -6,117 +6,123 @@
...
-    - LAI = (2 leaf area * 2 individuals * 2 LAI) / 8 area = 1
-    - trans = e ^ {-k L}, and since L = 1, trans = e^{-k}
+    Since the creation of the community canopy data is built into the cohort canopy
+    data creation, that dataclass is implicitly tested by checking the CohortCanopyData
+    class.
 
-    Simple community tests
-    - Three identical cohorts so community trans = (e{-k})^3 for each layer
-    - Transmission profile (e{-k})^3, e{-k})^6, e{-k})^9)
+    The simple four layer test three identical cohorts to give some easily defined
```
MarionBWeinzierl:
```suggestion
    The simple four layer test uses three identical cohorts to give some easily defined
```

MarionBWeinzierl (commented):
I've been through all the changes once and made some comments. Overall it looks good to me,  but @sjavis , could you please also review this PR, as it is quite substantial and I might have missed stuff.

docs/source/users/demography/light_capture.md
```diff
@@ -0,0 +1,520 @@
...
+```
+
+The average transmission through the layer can then be calculated as $t_l = 1 - a_l$ and
+the cumulative transmission $T_l$ can be calculated as the cumulative product of layer
+transmission values after setting $t_1 = 1$:
+
+$$
+\begin{align}
+T_{l} &= 1 \; l=1 \\
+       &= T_{l-1} \cdot t_l, \textrm{otherwise}
```
sjavis:
```suggestion
T_{l} &= 1, \textrm{if} \; l=1 \\
       &= T_{l-1} \cdot t_l, \textrm{otherwise}
```

pyrealm/demography/crown.py
davidorme: The predictions do make sense within $\[0, H\]$ and can be non-zero at those interval endpoints, so it is only values outside the interval that get set to zero.

docs/source/users/demography/light_capture.md
davidorme: That's correct - I've fixed this, and also unpacked it a bit more.

sjavis (approved):
It looks good to me too. The light model makes a lot more sense to me now.

docs/source/users/demography/canopy.md
```diff
@@ -45,188 +45,180 @@ from pyrealm.demography.canopy import Canopy
...
-  leaf surface for light capture. Values of $L$ are always expressed as the area of leaf
-  surface per square meter.
-* The extinction coefficient $k$ for a PFT sets how much light is absorbed when passing
-  through the leaf surface of that PFT.
+## Vertically structured canopies
 
-For a single stem, the fraction of light absorbed through the entire crown is described
-by the Beer-Lambert law:
+The simplest "big leaf" model of a canopy, represents the entire area of the crown
+($A_c$) as a single flat disc at the top of stem. However, the canopy model uses the
```
arne-exe, line 69:
```suggestion
Under the simplest "big leaf" model of a canopy, the area of an individual crown ($A_c$)
is represented as a single flat disc at the top of the stem. However, the canopy model uses the
```

arne-exe:
```suggestion
($A_c$) as a single flat disc at the top of the stem. However, the canopy model uses the
```

docs/source/users/demography/canopy.md
```diff
@@ -45,188 +45,180 @@ from pyrealm.demography.canopy import Canopy
...
+crown shape described in the [crown model](./crown.md) to model how leaf area
+accumulates vertically through the cohorts of a community.
+
+The canopy model can profile community crown area at given heights - useful to generate
+detailed plots of vertical structure - but also implements a specific model that
+partitions cohort crown area vertically into a series of closed canopy layers. The
+partitioning follows the perfect plasticity approximation (PPA) model
+{cite}`purves:2008a`, which assumes that plants are always able to plastically arrange
+their crown within the broader canopy of the community to maximise their crown size and
+fill the available space. The available space $A$ is simply the plot size that the
```
arne-exe: These sentences imply that PPA only works horizontally, but a crown will grow both horizontally (increase the projected crown area) and vertically (increase the crown length). 

Edit: After going through the rest of the documentation below, would it be correct to say that the model first maximises horizontal expansion, and if there's extra crown area that did not fit inside the "total area" then this extra gets added to the layer above (representing vertical expansion)? 
Related to this, I wonder if ecologically speaking this "crown-formation process" would take place bottom-up (i.e., upward crown growth is stimulated) or top-down (i.e., growth is suppressed at the bottom of the crown due to lack of space/competition, in other words the extra does not get added but gets removed instead). Not suggesting any changes here, just interesting to think about.

docs/source/users/demography/light_capture.md
davidorme: I agree it looks rubbish. I've updated this to use `\begin{cases}`, which adds a nice curly bracket and aligns the conditions. Like this:

image

docs/source/users/demography/canopy.md
```diff
@@ -45,188 +45,180 @@ from pyrealm.demography.canopy import Canopy
...
+community is growing in - all crown area is constrained to occur within the plot.
+
+If the total canopy area of all individuals within cohorts is less than $A$ then the
```
arne-exe: Should it not be "across cohorts" instead of "within cohorts"? If this calculation is done for each cohort separately (so within cohorts) then wouldn't this result in excessively large crown areas?

Edit: I think I understand why you used "within cohorts", it seems I misunderstood this part and thought you were describing here how the total projected crown area within a cohort is calculated (which is actually derived from the crown module). To avoid others making this same mistake, perhaps you can reiterate that this total crown area is derived from the crown module to smoothen this transition.

docs/source/users/demography/canopy.md
```diff
@@ -45,188 +45,180 @@ from pyrealm.demography.canopy import Canopy
...
+canopy model is complete - in this case, the predications are effectively the same as
+the big-leaf model. However if the total crown area across the stems is greater than $A$
```
arne-exe: Okay, after realising my edit in the comment above this second part also makes a lot more sense now. I still have a few questions here:
- So if this calculation happens within cohorts, it seems unlikely to me that the projected crown area of a particular cohort at a given height exceeds the total area of the grid. Am I missing something here? Can it be that this step works at community level and that the reference to cohorts is confusing?
- If this calculation/designation of total crown area happens within cohorts, and if this step is then repeated for each cohort, wouldn't that result in the overlapping of a certain available spacy by multiple tree crowns at once (which wouldn't be very realistic)? My guess is that by working with the total grid area you can avoid this (i.e., a certain space within a layer can only be filled once) but I am then curious as to how you decide which cohort gets priority to fill up that space first?

docs/source/users/demography/canopy.md
```diff
@@ -45,188 +45,180 @@ from pyrealm.demography.canopy import Canopy
...
+then the model finds heights at which successive layers fill up the available area,
+giving closed canopy layers.
+
+We can do this using the projected crown area ${A}_{p}(z)$ from the crown model, which
```
arne-exe: You nicely explain it here, perhaps worth moving this paragraph up a bit as it would help to interpret earlier paragraphs.

docs/source/users/demography/canopy.md
```diff
@@ -45,188 +45,180 @@ from pyrealm.demography.canopy import Canopy
...
+describes how crown area accumulates with height $z$ from the top of the stem to the
+ground, given the crown shape of the tree. If we sum the projected crown area across all
+stems in a community, we generate a prediction of the total community crown area at a
```
arne-exe: Here it is clear that you work with the summed projected crown area across all stems in the community. If you would move this paragraph up then the previous paragraph becomes a lot easier to interpret (i.e., it is clear that you are working with community projected crown area and not within cohorts).

docs/source/users/demography/canopy.md
```diff
@@ -45,188 +45,180 @@ from pyrealm.demography.canopy import Canopy
...
-A_{l[a,b]} = \tilde{A}_{cp}(z_a) - \tilde{A}_{cp}(z_b)
+\sum_1^{N_s}{ A_p(z^*_l)} = l A
 $$
 
-Given that information, the calculation of $f_{abs}$ becomes:
+An extra detail is that the canopy model includes a community-level canopy gap
+fraction ($f_G$) that captures the overall proportion of the canopy area that is left
+unfilled by canopy. This gap fraction, capturing processes such as crown shyness,
+describes the proportion of open sky visible from the forest floor. This term reduces
+the amount of $A$ that can be occupied by crown area before a layer closes:
```
arne-exe: Another thing to consider is that the total available area would also need to be corrected for the area occupied by stems, as leaves cannot grow inside a stem, although this area is likely to be relatively small at plot level.

docs/source/users/demography/canopy.md
```diff
@@ -45,188 +45,180 @@ from pyrealm.demography.canopy import Canopy
...
+partitioning follows the perfect plasticity approximation (PPA) model
+{cite}`purves:2008a`, which assumes that plants are always able to plastically arrange
+their crown within the broader canopy of the community to maximise their crown size and
+fill the available space. The available space $A$ is simply the plot size that the
+community is growing in - all crown area is constrained to occur within the plot.
+
+If the total canopy area of all individuals within cohorts is less than $A$ then the
+canopy model is complete - in this case, the predications are effectively the same as
+the big-leaf model. However if the total crown area across the stems is greater than $A$
+then the model finds heights at which successive layers fill up the available area,
```
arne-exe: Out of curiosity, what is the depth/thickness of a layer?

docs/source/users/demography/canopy.md
```diff
@@ -45,188 +45,180 @@ from pyrealm.demography.canopy import Canopy
...
+\sum_1^{N_s}{ A_p(z)}
 $$
 
-However, to calculate a vertical profile of light extinction through a crown with total
-area $A_c$ and maximum stem height $H$, that equation needs to be expanded to calculate
-the fraction of $L$ that falls between pairs of vertical heights $z_a > z_b$. The actual
-area amount of leaf area $A_l$ for an individual stem falling between those two heights
-is simply the diffence in projected leaf area between the two heights:
+We then need to find the heights $z^*_l$ at which the total community crown area
+occupies multiples of the available area $A$, giving the heights at which canopy
```
arne-exe: The purpose of this step is not entirely clear. Is the goal here to define the heights where the total community crown area is more than 1 time the available area? If so, does this mean that when it is more than 1 time the available area the additional crown area gets added to the next layer? Looking at the equations that follow it doesn't seem like the total community crown area for a given height is "updated" based on any extra from previous layers. It could be that I'm misunderstanding things here.

arne-exe (commented):
Hi @davidorme, it's nice to see this and I'm starting to better understand the canopy and crown module. I've left some comments on the canopy section, I'll have another look at the rest with a fresh mind.

docs/source/users/demography/canopy.md
arne-exe: When saying "total crown area across the stems is greater than A" does this refer to the total crown area at a specific height (I'm thinking of it like a slice of the crown at a specific height, done for all individuals in that cohort) or does this refer to the cumulative crown area of all the "slices" in a crown at all the different heights that should contain crown, summed across all individuals in that cohort?

docs/source/users/demography/light_capture.md
```diff
@@ -0,0 +1,526 @@
...
+$$
+
+The first two terms are the PPFD at the top of the canopy and a term $T_l$ that captures
+the fraction of the canopy top flux that is transmitted down to layer $l$.
+
+The last two terms then describe the ability of the individual to capture that
+downwelling light flux. The Beer-Lambert term is assumed not to change with height - the
+fraction of light absorbed per metre squared of leaf area does not vary through the
+canopy. The other term $A_{fl}$ is the area of foliage of that individual in the layer
+and is the difference between the projected leaf area at the top and bottom of the
```
arne-exe, line 130: I don't really understand this step, but that's probably because I don't fully understand the canopy model yet. The parts prior to this step are clear.

docs/source/users/demography/canopy.md
davidorme: Good point - will extend the description.

That's an interesting point about the growth of canopies - the stem has a fixed vertical shape for the crown area but as the stem grows of course the amount of crown at different points changes. The model _only_ follows that shape - trees do not grow into available space unless they have crown area within a particular layer. I guess the end result is that the shape represents ecological constraints on where a particular growth form can place leaves. That growth form is less plastic than reality.

docs/source/users/demography/crown.md
arne-exe, line 1: The paragraph starting on line 139 about the crown gap fraction states: "This trait
models how leaf gaps higher in the crown are filled by leaf area at lower heights:
it does not change the profile of the crown but allows leaf area in the crown to be
displaced downwards". 

I find the reasoning here a bit strange. I'm assuming that the model's gap fraction remains constant within the crown right? If so, wouldn't a tree choose to create more leaves in the gaps of the upper crown, where light availability is greatest? I'm not following why leaf area in the crown would be displaced downwards.

arne-exe: Line 160: provides --> provide

arne-exe: Line 193: calculting --> calculating

arne-exe: Line 290: add "how" in "the solid line shows **how** the actual crown radius $r(z)$ varies with height"

arne-exe: Line 421: seems like this sentence was abrupted.

arne-exe: The graphs and visualisation in the crown section are really nicely done!

docs/source/users/demography/light_capture.md
arne-exe: Does each layer consist of multiple horizontal "slices" or is each "slice" one layer? If it is the latter, then "top and bottom of the layer" would refer to the projected leaf area of the layer right? Perhaps you could add a bit more detail in the documentation on what is meant by a layer (also see the comment on the thickness of a layer in the canopy section).

docs/source/users/demography/light_capture.md
```diff
@@ -0,0 +1,526 @@
...
+ground.
+
+```{code-cell} ipython3
+simulated_transmission
+```
+
+The per stem fraction absorbed gives the fraction of the canopy top incident light
+absorbed by stem leaf area for each cohort in each layer.
+
+Critically these values **do not** assume any steam leaf area is actually present in the
```
arne-exe, line 297:
```suggestion
Critically these values **do not** assume any stem leaf area is actually present in the
```

arne-exe:
Hi @davidorme, I've gone through the other sections and provided feedback where I could. Some parts are a bit difficult for me to give feedback on as I don't fully understand what's going on. I hope the comments are helpful.

docs/source/users/demography/canopy.md
davidorme: The terminology is horrible in this section 😄 

I think avoiding referring to cohorts is worth doing, except to note that:

* a community is made up of a set of stems and the canopy model works across all stems simultaneously.
* those stems may be grouped into cohorts with identical DBH and PFT, but actually that's kind of irrelevant to the canopy model.

There is no priority system beyond the crown area at height $z$ function. The sum of crown area is across all stems simultaneously, so that allocation to layers purely follows that total accumulating area.

docs/source/users/demography/canopy.md
davidorme: I think this is effectively absorbed into the leaf area index? 

The projected crown area is not the total amount of leaf area in the layer - it's the area occupied by the crown in the layer. The same (confusingly) is true of the projected leaf area. The actual absorption is a function of the total leaf surface within the layer which is governed by the 2D projected area and the LAI in the Beer-Lambert law. So any single layer is really a volume containing stems and leaves and the total light absorption (which would be modified by stems) is captured by the LAI. A particularly stemmy PFT might have a lower LAI?

Thoughts?

docs/source/users/demography/canopy.md
davidorme: It is variable. The first layer starts at the top of the highest stem and closes at the first multiple of plot area. Then each lower layer is between successive lower layers until the last layer closes at ground level.

docs/source/users/demography/canopy.md
davidorme: The projected community crown area accumulates from the top of the highest stem down to ground area. The heights $z^*_l$ are then the points at which the community area exceeds multiples of $A$ or $A(1-f_G)$.

I think starting with that curve might be a good place to go from.

docs/source/users/demography/crown.md
davidorme: > I find the reasoning here a bit strange. I'm assuming that the model's gap fraction remains constant within the crown right? If so, wouldn't a tree choose to create more leaves in the gaps of the upper crown, where light availability is greatest? I'm not following why leaf area in the crown would be displaced downwards.

I see what you mean - I think of this as a crown morphology trait - some trees occupy a large space in the canopy but have relatively open crowns that let light pass down into canopy more. Others form compact crowns that form a more closed surface. This parameter is intended to capture that growth form.

If a tree has a more open canopy then the model says that it has the same projected crown area at a given height $z$ - and that drives how the tree takes up space in the canopy - but the actual leaf area can be distributed over  wider vertical range because of the openness of the canopy.

I'll add this.

davidorme:
@arne-exe I've reworked the docs  - in particular I have restructured the canopy model to give a more "across all individuals" based explanation. I've started with a more complex answer (ditching the simple conifer example) to try and connect the graphics to the explanation more cleanly. 

Could you have another read through and see what you think - does it address your more general comments well enough?

arne-exe:
> @arne-exe I've reworked the docs - in particular I have restructured the canopy model to give a more "across all individuals" based explanation. I've started with a more complex answer (ditching the simple conifer example) to try and connect the graphics to the explanation more cleanly.
> 
> Could you have another read through and see what you think - does it address your more general comments well enough?

Hi @davidorme, thank you for clarifying and replying to my questions, it seems I was misunderstanding several things, I will have another read through the crown-canopy-light capture modules again now.

docs/source/users/demography/canopy.md
```diff
@@ -43,209 +47,74 @@ from pyrealm.demography.community import Cohorts, Community
...
+### Community projected crown area
 
-$$
-L_H = \frac{i_h L_h A_h}{A_T}
-$$
+Each individual stem in a community has an projected crown area that describes how the
+crown area of that stem accumulates with height ($z$) from the top of the tree to the
+ground. The form of that curve varies with the size and PFT of the stem. If we have a
+community with $N_s$ individuals, then each stem $i = 1, ..., N_s$ has a projected crown
+area function $A_{p}(z)_i$.
```
arne-exe, line 99:
```suggestion
Each individual stem in a community has a crown area profile that describes how the
projected crown area of that stem accumulates with height ($z$) from the top of the tree
to the ground. For a community with $N_s$ individuals, the crown area profile of each stem
$i = 1, ..., N_s$ is captured by the function $A_{p}(z)_i$, where the shape of this curve
varies with the size and PFT of the stem. Summing the projected crown area for all values
of ($z$) contained within an individual's crown yields the total crown area.
```

arne-exe: The way I interpret this is that the crown area function predicts a projected crown area at a specific height (i.e., a particular slice of the crown at that height). This function is then repeated for all heights contained within a tree's stem height, which combined gives the crown area profile (and when summed gives the total projected crown area for that individual).

docs/source/users/demography/canopy.md
```diff
@@ -43,209 +47,74 @@ from pyrealm.demography.community import Cohorts, Community
...
-* a shorter understory tree with a columnar canopy and no crown gaps
-* a taller canopy tree with a top heavy canopy and more crown gaps
-
-and then three cohorts:
-
-* 7 saplings of the short PFT
-* 3 larger stems of the short PFT
-* 2 large stems of tall PFT
+To demonstrate this, the code below creates a plant community: there are 12 individual
+stems, grouped into 3 cohorts with different stem sizes and crown shapes.
```
arne-exe, line 109:
```suggestion
To demonstrate this, the code below creates a plant community that consists of 12 individual
stems, grouped into 3 cohorts with different stem sizes and crown shapes.
```

docs/source/users/demography/canopy.md
```diff
@@ -262,356 +131,363 @@ community = Community(
...
 
 ```{code-cell} ipython3
-fig, ax = plt.subplots(ncols=1)
+total_crown_area = np.sum(
+    community.stem_allometry.crown_area * community.cohorts.n_individuals, keepdims=True
+)
+total_crown_area
+```
+
+So, the maximum projected community crown area is around 120 m2. The plot below shows
```
arne-exe, line 152:
```suggestion
So, the total community crown area is around 120 m2. The plot below shows
```

docs/source/users/demography/canopy.md
arne-exe: This is more related to the crown module, but when looking at the figure below containing the community crown area, it seems like crowns can run all the way to the ground. I would assume that each stem has a crown-free section. Is this something that is included in the model (for example by restricting the lower limit of z values depending on stem height)?

docs/source/users/demography/canopy.md
```diff
@@ -262,356 +131,363 @@ community = Community(
...
+community is growing in a smaller area, and some crown area must overlie other parts
+of the community, creating a vertical structure.
 
-:::{admonition} TODO
+The canopy module implements the **perfect plasticity approximation (PPA)** model
+{cite}`purves:2008a` to generate this structure. The PPA model assumes that all the
+individuals within the community are able to plastically arrange their crown at each height
+within the broader canopy of the community to fill available space. When the available
+space ($A$) is filled, crown area lower in the canopy forms another layer until another
+area $A$ is filled and this repeats down to the ground.
```
arne-exe, line 243: I understand this now. I was initially focusing too much on how this would work for individual trees at each specific height (z), but that's not how it works.

Instead, it defines canopy layers by operating across all individuals and across multiple height values. For each height z, the projected crown area is summed across all individuals at that height. If the total does not exceed the available horizontal space, the model proceeds to the next lower height and adds the crown area at that level to a cumulative "bucket." This continues until the cumulative projected crown area exceeds the available space, at which point the current canopy layer is considered full. A new layer (or "bucket") is then started, and the process repeats down through the canopy.

arne-exe, line 243:
```suggestion
horizontal space ($A$) is filled by the cumulative crown area across all individuals at
one or more height levels ($z$), that canopy layer is considered "full" (or "closed").
The remaining crown area below this layer then begins to accumulate into a new layer.
This process repeats downward until the entire canopy structure is formed down to the ground.
```

docs/source/users/demography/canopy.md
```diff
@@ -262,356 +131,363 @@ community = Community(
...
 canopy_ppa.heights
 ```
 
-And the final value in the canopy extinction profile still matches the expectation from
-above:
+We plot those layer heights below. The vertical dashed lines in the right hand plot show
+the values $lA$ at which the cumulative space across layers forms closed layers: that is
+$32 \times 1 = 32$, $32 \times 2 = 64$, etc. . The horizontal dashed lines then show
+those calculated vertical heights. These are the canopy heights at which the vertical
+lines intersect the cumulativecommunity projected crown area: where the total crown area
```
arne-exe, line 291:
```suggestion
lines intersect the cumulative community projected crown area: where the total crown area
```

arne-exe: This figure is very neat. Once it clicks how everything works it's a very nice visualisation.

docs/source/users/demography/canopy.md
```diff
@@ -262,356 +131,363 @@ community = Community(
...
-# LH plot - projected leaf area with height.
+plt.tight_layout()
+```
 
-# Add lines showing the canopy closure heights and closure areas.
-for val in canopy_ppa.heights:
-    ax1.axhline(val, color="red", linewidth=0.5, zorder=0)
+We can look in detail at the how much crown area is in each layer from each individual.
+The canopy model object contains a crown profile which records the projected crown
+area at each height:
```
arne-exe, line 334:
```suggestion
area at each height, summarised at cohort level:
```

arne-exe: Looking at the example array this seems to give a summary at cohort level, not individual

docs/source/users/demography/canopy.md
```diff
@@ -262,356 +131,363 @@ community = Community(
...
-ax1.plot(
-    community_leaf_area,
-    canopy.heights,
-    zorder=1,
-    linestyle="--",
-    color="black",
-    linewidth=1,
-    label="Leaf area",
-)
+Those values are the projected *individual* crown areas: the accumulating crown area
```
arne-exe, line 340:
```suggestion
Those values are the cumulative projected crown areas: the accumulating crown area
```

arne-exe: I removed *individual* as this does not seem to refer to individual trees anymore, but to cohort level instead (see comment above).

docs/source/users/demography/canopy.md
```diff
@@ -262,356 +131,363 @@ community = Community(
...
+from the top of the canopy down to the ground. We can take the differences between
+vertical layers to give the actual amount of crown area in each layer for each
+individual.
```
arne-exe, line 343:
```suggestion
cohort.
```

arne-exe: This also works at cohort level right (see comments above)?

docs/source/users/demography/canopy.md
```diff
@@ -262,356 +131,363 @@ community = Community(
...
+)
+individual_crown_in_layer
 ```
 
-## Light allocation
+To bring those values back to the community model: each column represents a cohort, so
+we can multiply those values by the number of individuals and then find the row sums
+to show the amount of crown area in each layer. As expected, the crown area in each
+layer matches the 32 m2 of available space, except for the last layer that is not
+completely filled.
```
arne-exe, line 356: Neat, nice when things add up! 😁

docs/source/users/demography/light_capture.md
arne-exe: I think one of the reasons for why I was confused is that I did not understand that this works with the *cumulative* projected leaf area. I think this is because of a nuance that is taken for granted here, being that "projected" implies "cumulative", which may not be clear to everyone. This could be solved by adding "cumulative projected leaf area".

docs/source/users/demography/light_capture.md
```diff
@@ -0,0 +1,558 @@
...
+    # Scale that back down to give the average captured flux per m2
+    average_captured_ppfd = cohort_capture.sum() / cell_area
+    simulated_total_capture += cohort_capture.sum()
+
+    # Calculate the remaining flux transmitted to the next layer
+    ppfd -= average_captured_ppfd
+    simulated_transmission[layer + 1, 0] = ppfd / initial_ppfd
+```
+
+The resulting transmission start with the fraction of light reaching the first layer
```
arne-exe, line 286:
```suggestion
The resulting transmission starts with the fraction of light reaching the first layer
```

arne-exe (commented):
Hi @davidorme, I've gone through the canopy module again and it makes a lot of sense now. I've added a few suggestions but it's looking very good!

docs/source/users/demography/canopy.md
arne-exe: Thanks for clarifying, I now also realise that the first part of my question was not relevant, so no more questions here.

docs/source/users/demography/canopy.md
arne-exe: This question is no longer relevant.

docs/source/users/demography/canopy.md
arne-exe: Thanks! This part is clear now.

docs/source/users/demography/canopy.md
arne-exe: I guess it would depend on how LAI data is collected, and whether or not they corrected for stems (e.g., in hemisperical images). 

The other thought I had on this was whether gaps in the crown/canopy are corrected for twice (as the model uses both LAI and gap fraction)? I'm sure this has crossed your mind already, but I couldn't quite figure out how one correction relates to the other, as they are used in different modules (crown and light capture) but the light capture module relies on the crown leaf area (if I'm not mistaken).

docs/source/users/demography/canopy.md
arne-exe: This makes sense now!

docs/source/users/demography/canopy.md
arne-exe: Yup, this makes sense now!

docs/source/users/demography/canopy.md
davidorme: There isn't a double counting - I'll try and clarify this somewhere. Does this text help?

* The LAI is "just" a characteristic value of the PFT setting the amount of leaf area per m2 of crown area.
* The crown gap fraction effectively tunes where vertically in the crown that LAI is deployed. The crown gaps do not push all the way down to the ground - they only allow light to penetrate more deeply into the crown.
* The canopy gap fraction is never intercepted by any crown - all light within the canopy gap fraction passes unimpeded from the top of the canopy to the ground.

docs/source/users/demography/canopy.md
davidorme: Not quite - those values are for each individual within each cohort. I've clarified to make this structure clearer.

docs/source/users/demography/canopy.md
davidorme: Nope - that is individual level. The bottom row matches the total individual crown area from just the allometry above . I've added a note.

docs/source/users/demography/canopy.md
davidorme: Good - glad it helps. I think it is a lot clearer now this is set out like this

docs/source/users/demography/canopy.md
davidorme: Exactly. The "next lowest height" is not found from a fixed set of heights, but instead a numeric method is used to wiggle the height around until the bucket pretty much exactly fills (to about 1 cm2 precision by default). But otherwise this is a great summary.

docs/source/users/demography/canopy.md
davidorme: @arne-exe Thanks a lot for going through and saying where questions have been resolved - super helpful for me to know when I don't need to take any further action. There is also the "resolve conversation" button below, but having the person who asked the question specifically say that it is resolved is clearer, so much appreciated.

docs/source/users/demography/canopy.md
davidorme: Hmmm. This is something I need to clarify in the crown model. If I've got this right, the crown is technically only the upper surface of the overall crown shape (above the maximum crown radius). So the stem below that level, which has a relative height of  1 - the crown fraction, is crown free.

BUT... The crown gap fraction pushes leaf area down below that crown surface layer, so trees where the crown gap fraction is > 0 will have some leaf surface below that crown fraction height.

I might make this a new issue, just to get this closed down!

arne-exe: > Hmmm. This is something I need to clarify in the crown model. If I've got this right, the crown is technically only the upper surface of the overall crown shape (above the maximum crown radius). So the stem below that level, which has a relative height of 1 - the crown fraction, is crown free.
> 

Ah, you're right, I didn't realise this before. Indeed, the figure right above where you introduce the PPA model shows that below the maximum crown radius the cumulative projected crown area does not increase for a particular individual/cohort.

PR #522, opened 2025-06-23
Adding UUIDs to cohorts

### LINKED ISSUE #521: Add internal cohort ID to cohorts
The identity of cohorts across Communities and other classes supporting the CohortMethods mixin relies on the order of the communities. If a user is saving data for communities (e.g. to csv) - or really anything that uses `CohortMethods` - then some kind of ID is needed to identify which rows belong to which cohort.

The simplest method here is automatically populate an internal UUID field within the `Cohorts` class. UUIDs might be overkill here but they should be unique and are reasonably cheap to generate. We could make users provide their own IDs (maybe some combination of a Community cell id and a per instance sequence generator). We could come back to this, but for the moment, using UUIDs seems quick and easy.

davidorme:
# Description

This PR adds a new instance attribute to the `Cohorts` dataclass that is used to contain unique IDs for each cohort. The main rationale (as in the issue #521) is to provide a unique cohort tag if data is exported from `pyrealm` community objects.

* The UUID values are populated via the `__post_init__` method. 
* A `@property` and setter are defined to enforce unique IDs - not because of the minute chance of UUID collisions, but because users could potentially add existing cohort instances onto themselves, which would create a duplicate.
* The tests are extended to check UUIDs are being populated with UUID values and to check the duplicate detection works. One test that would have failed (because it added a `Cohorts` instance to itself) is also fixed.
* And `jupytext` stuck its oar in and made an unrelated format of a Markdown file.

Fixes #521

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/users/demography/canopy.md
+++ b/docs/source/users/demography/canopy.md
@@ -17,7 +17,7 @@ language_info:
   name: python
   nbconvert_exporter: python
   pygments_lexer: ipython3
-  version: 3.12.3
+  version: 3.11.9
 settings:
   output_matplotlib_strings: remove
 ---
@@ -110,11 +110,9 @@ individual stems, grouped into 3 cohorts with different stem sizes and crown sha
 
 ```{code-cell} ipython3
 # Define PFTs
-short_pft = PlantFunctionalType(
-    name="short", h_max=15, m=1.5, n=1.5, f_g=0, ca_ratio=380
-)
+short_pft = PlantFunctionalType(name="short", h_max=15, m=2, n=4, f_g=0, ca_ratio=380)
 tall_pft = PlantFunctionalType(
-    name="tall", h_max=30, m=1.5, n=2, par_ext=0.6, f_g=0, ca_ratio=500
+    name="tall", h_max=30, m=2, n=3, par_ext=0.6, f_g=0, ca_ratio=500
 )
 
 # Create the flora
--- a/tests/unit/demography/test_community.py
+++ b/tests/unit/demography/test_community.py
@@ -11,12 +11,25 @@ from numpy.testing import assert_allclose
 
 
 @contextmanager
-def not_raises(exception):
-    """Simple context manager for checking a test does not raise."""
+def not_raises():
+    """Simple context manager for checking a statement does not raise."""
     try:
         yield
-    except exception:
-        raise pytest.fail(f"DID RAISE {exception}")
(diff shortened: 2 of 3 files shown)


docs/source/users/demography/canopy.md
```diff
@@ -17,7 +17,7 @@ language_info:
   name: python
   nbconvert_exporter: python
   pygments_lexer: ipython3
-  version: 3.12.3
+  version: 3.11.9
```
MarionBWeinzierl, line 20: Is the version intentionally downgraded?

tests/unit/demography/test_community.py
```diff
@@ -8,6 +10,15 @@
 from numpy.testing import assert_allclose
 
 
+@contextmanager
+def not_raises(exception):
```
MarionBWeinzierl: Seems odd that there should not be anything built-in to check this.

tests/unit/demography/test_community.py
```diff
@@ -113,26 +131,48 @@ def test_Cohorts(args, outcome, excep_message):
...
+        _ = cohorts.add_cohort_data(new_data=cohorts)
+
+
 def test_Cohorts_CohortMethods():
     """Test the inherited CohortMethods methods."""
 
     from pyrealm.demography.community import Cohorts
 
-    # Create and instance to modify using methods
+    # Create and instances to modify using methods
```
sjavis:
```suggestion
    # Create instances to modify using methods
```
I think this was an existing typo that was missed

sjavis (commented):
It looks good to me, just one small comment.

docs/source/users/demography/canopy.md
davidorme: Gah - it just depends on which `venv` I have active when I run the `pre-commit`. I have parallel ones for different python versions.

tests/unit/demography/test_community.py
davidorme: Good point - I've revamped this to make it cleaner and indeed clearer

docs/source/users/demography/canopy.md
davidorme: It's not harmful - and I don't think is consistent across all of the Myst notebooks - but it would be good to avoid doing this in the future.

MarionBWeinzierl (approved):

PR #525, opened 2025-06-24
Fixing behaviour of T Model with zero DBH values

### LINKED ISSUE #524: Behaviour of T Model allometry and allocation with zero DBH values
Currently, if the DBH ($D$) of a stem is zero, then in the `StemAllometry` class:

* the crown fraction ($f_c$) is NaN, so...
* the sapwood mass (which uses $f_c$) is NaN 

And in the `StemAllocation` class:

* Consequently, sapwood respiration is NaN and as a result NPP and all the delta variables (diameter, mass, foliage) are NaN.

davidorme:
# Description

> [!Tip]
> OK - this is ready for review now.

Currently, if the DBH ($D$) of a stem is zero, then in the `StemAllometry` class:

* the crown fraction ($f_c$) is NaN, so...
* the sapwood mass (which uses $f_c$) is NaN 

And in the `StemAllocation` class:

* Consequently, sapwood respiration is NaN and as a result NPP and all the delta variables (diameter, mass, foliage) are NaN.

That leads to all sorts of unexpected behaviour. We _could_ make more sensible predictions when $D=0$ - which is what the first iteration of this PR did - but it basically doesn't make sense to make predictions when $D=0$.  That's actually a bit more work because we currently allow zero so this PR:

* Adds a new `_enforce_positive_sizes` helper function to `pyrealm.demography.tmodel` 
* Calls that function from within _nearly_ all of the validation sections of the T model functions. We don't call it on the reproductive mass functions because that is an experimental extension that we want to be able to set to zero.
* Add tests that zero and negative values are caught correctly.
* Updates tests to pass

The PR also: 

* Makes it a ValueError to create `StemAllometry` or `Cohorts` with negative DBH - this guarantees positive sizes and allows the per function validation to be skipped.
* Makes `Cohorts.dbh_values` a property raises ValueError when set with negative DBH values.

Fixes #524 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/demography/community.py
+++ b/pyrealm/demography/community.py
@@ -129,7 +129,7 @@ from __future__ import annotations
 import json
 import sys
 import uuid
-from dataclasses import dataclass, field
+from dataclasses import InitVar, dataclass, field
 from pathlib import Path
 from typing import Any, ClassVar
 
@@ -173,15 +173,16 @@ class Cohorts(PandasExporter, CohortMethods):
     count_attr: ClassVar[str] = "n_cohorts"
 
     # Instance attributes
-    dbh_values: NDArray[np.float64]
     n_individuals: NDArray[np.int_]
     pft_names: NDArray[np.str_]
     _cohort_id: NDArray[np.str_] = field(init=False)
+    _dbh_values: NDArray[np.float64] = field(init=False)
     n_cohorts: int = field(init=False)
+    dbh_values: InitVar[NDArray[np.float64]]
 
     __experimental__ = True
 
-    def __post_init__(self) -> None:
+    def __post_init__(self, dbh_values: NDArray[np.float64]) -> None:
         """Validation of cohorts data."""
 
         # TODO - validation - maybe make this optional to reduce workload within
@@ -191,7 +192,7 @@ class Cohorts(PandasExporter, CohortMethods):
 
         # Check cohort data types
         if not (
-            isinstance(self.dbh_values, np.ndarray)
+            isinstance(dbh_values, np.ndarray)
             and isinstance(self.n_individuals, np.ndarray)
             and isinstance(self.pft_names, np.ndarray)
         ):
@@ -199,11 +200,15 @@ class Cohorts(PandasExporter, CohortMethods):
 
         # Check the cohort inputs are of equal length
         try:
-            check_input_shapes(self.dbh_values, self.n_individuals, self.dbh_values)
+            check_input_shapes(self.pft_names, self.n_individuals, dbh_values)
         except ValueError:
             raise ValueError("Cohort arrays are of unequal length")
 
-        self.n_cohorts = self.dbh_values.size
+        if np.any(dbh_values < 0):
+            raise ValueError("Negative DBH values passed to Cohorts instance.")
+
+        self._dbh_values = dbh_values
+        self.n_cohorts = dbh_values.size
         self._cohort_id = np.array([str(uuid.uuid4()) for _ in range(self.n_cohorts)])
 
     @property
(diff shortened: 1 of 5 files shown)


pyrealm/demography/community.py
```diff
@@ -173,15 +173,16 @@ class Cohorts(PandasExporter, CohortMethods):
     count_attr: ClassVar[str] = "n_cohorts"
 
     # Instance attributes
-    dbh_values: NDArray[np.float64]
     n_individuals: NDArray[np.int_]
     pft_names: NDArray[np.str_]
     _cohort_id: NDArray[np.str_] = field(init=False)
+    _dbh_values: NDArray[np.float64] = field(init=False)
```
MarionBWeinzierl, line 179: Probably just me as I am not working in the field, but for me and other field-outsiders of the code it would help to add a short comment what `pft` or and `dbh` is (ideally, the variable names would be clear, but I assume these are abbreviations which are commonly used in the field?).

pyrealm/demography/community.py
```diff
@@ -173,15 +173,16 @@ class Cohorts(PandasExporter, CohortMethods):
...
     n_cohorts: int = field(init=False)
+    dbh_values: InitVar[NDArray[np.float64]]
```
MarionBWeinzierl, line 181: I am a bit confused here why we need `dbh_values` as well as `_dbh_values`.

MarionBWeinzierl (commented):
I am now really curious what DBH means... :D

pyrealm/demography/community.py
davidorme: They are "plant functional type" and "diameter at breast height" - which are just a bit long winded to spell out so those abbreviations are commonly used.

pyrealm/demography/community.py
davidorme: It's a property thing: the user provides `dbh_values` to create the instance, which is then passed through the validation in the property setter and stored in `_dbh_values`. When a user accesses `dbh_values` the property decorator function returns the contents of the validated  `_dbh_values`.

It's a relatively common Python idiom for controlling class attributes (e.g. https://realpython.com/python-property/), but just looks a bit weird within a dataclass because some of the class attribute creation is concealed within the `@dataclass`. We could just make it a normal class for clarity, but this works as intended.

pyrealm/demography/community.py
davidorme: It's not a direct solve for reading the code, but I've added #526 to add a glossary.

pyrealm/demography/tmodel.py
```diff
@@ -500,9 +549,13 @@ def calculate_reproductive_tissue_respiration(
         validate: Boolean flag to suppress argument validation
     """
     if validate:
+        size_args = {"reproductive_tissue_mass": reproductive_tissue_mass}
         _validate_demography_array_arguments(
-            trait_args={"resp_rt": resp_rt},
-            size_args={"reproductive_tissue_mass": reproductive_tissue_mass},
+            trait_args={"resp_rt": resp_rt}, size_args=size_args
+        )
+        _enforce_positive_sizes(
```
sjavis: A later function (calculate_reproductive_tissue_turnover) allows a reproductive tissue mass of zero. Should that be allowed here?

pyrealm/demography/tmodel.py
```diff
@@ -785,6 +855,13 @@ def calculate_growth_increments(
     * the crown area ratio (:math:`c`), and
     * the ratio of fine root mass to leaf area (:math:`\zeta`).
 
+    The value of :math`\Delta D` is unstable when :math:`D = 0` and hence :math:`H = 0`
+    and the rates of change in stem and foliar mass are also zero. If :math:`P_{net} - T
+    = 0` then :math`\Delta D` is undefined, otherwise :math`\Delta D = \pm \inf`
+    depending on whether then turnover costs exceed the available NPP. Under these
+    conditions, this function explicitly sets :math`\Delta D = 0`: **stems with zero
```
```suggestion
    The value of :math:`\Delta D` is unstable when :math:`D = 0` and hence :math:`H = 0`
    and the rates of change in stem and foliar mass are also zero. If :math:`P_{net} - T
    = 0` then :math:`\Delta D` is undefined, otherwise :math:`\Delta D = \pm \inf`
    depending on whether then turnover costs exceed the available NPP. Under these
    conditions, this function explicitly sets :math:`\Delta D = 0`: **stems with zero
```

sjavis (approved):
Overall this looks good to me aside from a couple of comments.

PR #527, opened 2025-06-27
Add a glossary

### LINKED ISSUE #526: Add a glossary
Probably just me as I am not working in the field, but for me and other field-outsiders of the code it would help to add a short comment what `pft` or and `dbh` is (ideally, the variable names would be clear, but I assume these are abbreviations which are commonly used in the field?).

_Originally posted by @MarionBWeinzierl in https://github.com/ImperialCollegeLondon/pyrealm/pull/525#discussion_r2168576221_

The comment above was more aimed at the variable names in code, but it would partly solve this issue _and_ make the docs easier to read and understand. There is a good `sphinx` extension.

davidorme:
# Description

This PR adds a glossary to the webpage and (experimentally) adds hoverxref to get tooltip definitions. This currently doesn't work for me locally due to some odd CORS issue in the links, but it does work on RTF. See the underlined terms here:

https://pyrealm.readthedocs.io/en/526-add-a-glossary/users/demography/canopy.html#community-definition
https://pyrealm.readthedocs.io/en/526-add-a-glossary/users/demography/light_capture.html#light-capture-in-a-simple-model

More broadly we need to think about:

* Where and which terms on a page get this treatment because it could get visually overwhelming. First use on a page probably, but I'm wondering if having a per page list of terms in a box might work. 
* We also need to do this in the docstrings, and that is hard to keep clean.

However, worrying about that should probably wait until _after_ we implement it and start building it out.

Fixes #526 (issue)

## Type of change

## Key checklist

## Further checks


sjavis (approved):
This looks good to me. The hover definitions don't work on my machine either, but it probably isn't very important since it works on RTD. Personally I don't view the definitions as intrusive so I don't think we would need to restrict their use to the first occurrence on a page. Although I think they could be left out from the docstrings / API.

It does not need to be included in this PR, but, for reference, here is a list of terms that I have struggled with and think would be good in the glossary:
- GPP
- PFT
- Allometry
- LAI
- FAPAR
- AET

PR #533, opened 2025-07-16
Extend `FaparLimitation` to include the $m$ parameter

### LINKED ISSUE #349: Calculate tunable parameter $m$ defined as the ratio of steady state annual LAI to GPP.
This is the next step in calculating the LAI phenology - the description is taken from the documented attached to the meta  issue #347:

The equation to calculate parameter m is shown in equation 4: 

(Eqn 4) $m =(\sigma  \cdot GSL \cdot LAI_{max})/ (A_{0sum} \cdot f_{APAR_{max}})$

This shouldn't be too difficult. Apart from $\sigma$, the values are all taken directly from the outputs and inputs to #348

* $GSL$ is the length of the continuous above 0℃ period longer than five days, which is a required input for 348.
* LAImax is annual peak LAI, calculated from Equation 2 in 348.
* A0sum is the annual total potential GPP, from the P Model input to 348.
* fAPARmax is annual peak fAPAR, calculated from Equation 1 in 348.

The parameter $\sigma$ then represents the extent to which seasonal LAI dynamics depart from a “square wave” whereby maximum LAI would be maintained over the whole growing season. IIUC, this is used to effectively round off the shoulders of that square expectation to account for leaf growth and budburst at the start of the season and then also nutrient resorption and leaf senescence processes at the end.

davidorme:
# Description

This PR:

* Extends the `FaparLimitation` class to  calculate the annual steady state ratio of LAI to GPP ($m$), as requested in #349. The $m$ values are annual varying and so fit naturally as part of `FaparLimitation`
* Adds the required `sigma` parameter to `PhenologyConsts`
* Updates the regression dataset implementation to add $m$ to fortnightly outputs and suppress some noisy near-zero values in the output files.
* Some use of glossary terms crept in by accident - committed to the wrong branch..

Fixes #349

The next stage is to take daily GPP values and convert them to the LAI time series. That will be a new structure.

## Type of change

## Key checklist

## Further checks


pyrealm/constants/phenology_const.py
```diff
@@ -25,12 +25,18 @@ class PhenologyConst(ConstantsClass):
...
     f0_coefficients: tuple[float, float, float] = (0.65, 0.604169, 1.9)
-    """Coefficients :math:`a,b,c` to calculate :math:`f_0` from the local aridity index
+    r"""Coefficients :math:`a,b,c` to calculate :math:`f_0` from the local aridity index
     (AI), where :math:`a` is the maximum value, :math:`b` is the slope of the
     relationship with AI and :math:`c` is the AI value at which the maximum value
     :math:`a` is reached.
     """
 
+    sigma: float = 0.771
+    r"""The :math:`\sigma` parameter is a proportion that captures the opportunity costs
```
MarionBWeinzierl: Sorry, you totally lost me there. I cannot understand this sentence.

tests/regression/phenology/test_phenology_faparlimitation.py
```diff
@@ -184,7 +194,7 @@ def test_faparlimitation_frompmodel_subdaily(
         tc=subdaily_data["tc"].to_numpy(),
         vpd=subdaily_data["vpd"].to_numpy(),
         co2=subdaily_data["co2"].to_numpy(),
-        patm=subdaily_data["patm"].to_numpy(),
+        patm=subdaily_data["patm"].to_numpy().astype(float),
```
MarionBWeinzierl, line 199: Why did this become necessary?

MarionBWeinzierl (approved):
Looks good, there is just one bit of documentation that I don't understand (maybe that sentence can be simplified?), and one more question.

tests/regression/phenology/test_phenology_faparlimitation.py
davidorme: For some reason it is now being read as an integer and that runs into an issue in `SubdailyPModel` - basically it tries to set `np.nan` in a integer array, which of course doesn't understand NaN.  That might be a wider issue (check the inputs are floats or convert them internally) but this is just a hack to get the test working.

pyrealm/constants/phenology_const.py
davidorme: Yeah. It's hideous now that I re-read it.

PR #535, opened 2025-07-17
Calculate the phenology

### LINKED ISSUE #350: Calculate the steady state LAI.

### LINKED ISSUE #351: Calculate the realised LAI given a lagged process.

davidorme:
# Description

This is a draft PR that:

* Adds a year attribute to `FaparLimitation` to record a timestamp for each annual value.
* Adds a draft of the `Phenology` class , which:
    * Takes a dated set of daily GPP values and an existing `FaparLimitation` instance
    * Maps annual maximum LAI values onto the daily observations
    * Calculates steady state LAI and hence the lagged realised LAI using an exponential weighted average.
* Updates tests and adds a simple regression test of `Phenology` 

> [!IMPORTANT]
> I'm not sure about the structure of this yet - we need some kind of `from_PModel` method, but I think that should be higher level. The API for a PModel should really do both the FaparLimitation and Phenology calculation in one pass - we need to keep the inputs aligned and having it as a two stage thing just makes life more complex for us and users.
> We also need something to get daily values from GPP estimated at different time resolutions.

Fixes #350 
Fixes #351 

## Type of change

## Key checklist

## Further checks


pyrealm/phenology/fapar_limitation.py
MarionBWeinzierl, line 1: Maybe this should be a phenology class, not a fapar_limitation one? Do we need to be able to compute fapar and lai anywhere else? Otherwise those constructors could become class methods for the phenology class.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -299,6 +306,100 @@ def from_pmodel(
...
+        # Check the array input shapes
+        check_input_shapes(daily_gpp, datetimes)
+
+        # Check the datetimes provide ordered daily resolution observations - don't
+        # insist on daily precision but check that second level representations of the
+        # days is consistent with daily observations and strictly increases.
+        datetimes = datetimes.astype("datetime64[s]")
+        datetime_spacing = np.diff(datetimes)
+
+        if not np.all(np.equal(datetime_spacing, 86400)):
```
MarionBWeinzierl, line 436: wondering whether there should be a `check_datetimes` function (if there isn't already somewhere? I thought there was) that does all kinds of checks to make sure datetimes are what they are supposed to be, as we keep checking them.

MarionBWeinzierl:
@davidorme I'd think it makes more sense to create a new branch from `develop` with all the changes. What do you think?

davidorme:
Yup - I'd agree. I think the core of the class is pretty solid but probably easier to close this and start afresh.

I'm inclined to keep this as requiring daily GPP and then have a `get_daily_gpp` methods that converts PModel and SubdailyPModel to daily estimates. I think that might be reincarnating the `get_daily_values` function you wrote!

MarionBWeinzierl:
After an attempted rebase that went horribly wrong, I have created #565 to continue working from. It is based off `develop` cherry picks @davidorme 's changes from here.

davidorme:
I still view rebasing as basically black magic. Technically very powerful, but I'm not going anywhere near it.

PR #557, opened 2025-10-03
Update NDArray typing

### LINKED ISSUE #556: Convert array typing from `float64` to `floating`
We're currently typing all float NDArray objects as `NDArray[np.float64]` - specifically `float64` which can cause typing issues when basically any float would be fine. It would be better to cast them all using the relatively recent `np.floating` generic  - which is a type alias for any numpy float dtype.

davidorme:
# Description

With newer `numpy` versions, it is probably cleaner to move to the newer typing options for typing floating point arrays. We currently use `np.float64` widely, but we really don't care about the float length particularly and these can be more generically typed as `np.floating`, which reduces chances of meaningless float typing problems if `pyrealm` is called in other packages.

So, this PR:
* Moves all `NDArray[np.float64]` to `NDArray[np.floating]`
* Adds `[np.floating]` typing to nearly all bare `NDArray` types. `pyrealm.core.time_series.broadcast_time` does not care what it broadcasts, although `np.floating` is most likely.
* Squashes a few rogue `NDArray[np.float32]` types back into line as `NDArray[np.floating]`

@sjavis - I think this is most likely to impact your `xarray` work so I've asked you to review. I think standardising our float array typing is going to be helpful, but it might causes issues!

Fixes #556 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/constants/core_const.py
+++ b/pyrealm/constants/core_const.py
@@ -142,7 +142,7 @@ class CoreConst(ConstantsClass):
     for 2000 CE :cite:t:`berger:1978a`."""
 
     # Hygro constants
-    magnus_coef: NDArray[np.float64] = field(
+    magnus_coef: NDArray[np.floating] = field(
         default_factory=lambda: np.array((611.2, 17.62, 243.12))
     )
     """Three coefficients of the Magnus equation for saturated vapour pressure,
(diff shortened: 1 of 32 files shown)


sjavis (approved):
I was also thinking this should be changed. I have had a brief glance over all the changes and it seems all good to me. I think some `float32`s got missed in crown.py but I am happy with it other than that.

I agree with leaving `broadcast_time` without a type. Although it is currently only used with floats this could change in the future.

davidorme:
@sjavis Not sure how I missed those extra `float32` usages - those are now fixed. Thanks!

PR #559, opened 2025-10-07
Build out FaparLimitation to handle ND inputs

### LINKED ISSUE #555: Build out FaparLimitation to handle ND inputs
At the moment `FaparLimitation` assumes 1D arrays, but it could support gridded data. At this point, we also need to assume a time axis to map years onto an axis. This is needed for the `Phenology` (see #350) code.

So we need to update `FaparLimitation` to take ND inputs on all but the `years` argument and then check `years` is congruent with the first dimension of the arrays.

davidorme:
# Description

This PR is to build out the implementation of `FaparLimitation` to handle ND inputs. That turned out to involve quite a lot of work!

* Work starts with the `AnnualValueCalculator` (`AVC`), which underpins the `FaparLimitation`. 
    * First things - the `growing_season` terminology is very fAPAR specific. It's just a temporal subset mask, so lets call it `subset_mask` in the general `AVC` implementation.
    * I think the basic idea is still fine but the dimensions in the inputs to `AnnualValueCalculator.__init__` can be quite variable. 

      Say you want to use `AVC` to get annual stats for 10 years of monthly data on a 5x5 spatial grid: the target data shape is `(120, 5, 5)`. The first axis is time, so the datetimes can only be `(120,)`  but then users can provide a `subset_mask` and that basically has to be able to broadcast onto the target shape. So it could be `(120, 1, 1)` (constant across sites) or `(120,5,5)` (site specific) or even something bizarre like `(120,5,1)` (constant across the X dimension). 

      The problem with the previous code was that actual intended data shape isn't definitively known until you call one of the `AVC.get_annual_...` methods. That makes for messy internals in the `__init__` method and a load of hanging broadcasting. If users have to explicitly say what shape the intended inputs are going to be, then the broadcasting can be handled much more cleanly.

      So this change now makes users **specifically set the intended use shape in  the `data_shape` argument**. The `AVC.__init__` process was always really about setting up preprocessed things to resample a specific data set - it's just now a bit more specific. The attributes of `AVC` are now set up to be correctly broadcast to the input data, which makes the internals of the `get_annual_...` methods simpler.
      
    * I've updated the `AVC.__init__` testing to match and added a whole new test set that checks that dimensional inputs work.

* The work in `FaparLimitation` is then pretty simple:
    * It requires users to provide `year` labels for the observations - we need this to map annual stats back out to days for calculating phenological time series.
    * I've added a draft `__repr__` and `summarise` method
    * Updated the `AVC` calls to add the new arguments.
    * Added some tests that use the regression test data to test different dimension inputs using `FaparLimitation.__init__` directly and using `FaparLimitation.from_pmodel`.

* Adds a first pass at a user facing doc on `FaparLimitation`

* Some other minor tweaks to syntax and docstrings snuck in.

Fixes #555 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/users/phenology/overview.md
+++ b/docs/source/users/phenology/overview.md
@@ -20,7 +20,7 @@ series of {term}`FAPAR` and {term}`LAI`. There are two main stages in this proce
    constraints on potential gross primary productivity arising from either water
    limitation or energy limitation {cite:p}`cai:2025a`. This is implemented in `pyrealm`
    as the {class}`~pyrealm.phenology.fapar_limitation.FaparLimitation` class and is
-   described in the [introduction to calculating maximum FAPAR](./fapar_limitation.py).
+   described in the [introduction to calculating maximum FAPAR](./fapar_limitation.md).
 
 2. The maximum annual FAPAR and LAI can then be combined with daily predictions of
    potential GPP to predict time series of realised FAPAR and LAI through the year,
--- a/docs/source/users/phenology/overview.md
+++ b/docs/source/users/phenology/overview.md
@@ -23,5 +23,5 @@ series of {term}`FAPAR` and {term}`LAI`. There are two main stages in this proce
    described in the [introduction to calculating maximum FAPAR](./fapar_limitation.md).
 
 2. The maximum annual FAPAR and LAI can then be combined with daily predictions of
-   potential GPP to predict time series of realised FAPAR and LAI through the year,
-   [TBD].
+   potential GPP to predict time series of realised FAPAR and LAI through the year.
+   The implementation of this functionality is currently in development.
--- a/docs/source/users/phenology/overview.md
+++ b/docs/source/users/phenology/overview.md
@@ -1,0 +1,27 @@
+---
+jupytext:
+  formats: md:myst
+  text_representation:
+    extension: .md
+    format_name: myst
+    format_version: 0.13
+kernelspec:
+  display_name: Python 3 (ipykernel)
+  language: python
+  name: python3
+---
+
+# The phenology module
+
+The {mod}`~pyrealm.phenology` module implements tools for predicting phenological time
+series of {term}`FAPAR` and {term}`LAI`. There are two main stages in this process:
+
+1. Annual maximum values of FAPAR and LAI for a location can be predicted by comparing
+   constraints on potential gross primary productivity arising from either water
+   limitation or energy limitation {cite:p}`cai:2025a`. This is implemented in `pyrealm`
+   as the {class}`~pyrealm.phenology.fapar_limitation.FaparLimitation` class and is
+   described in the [introduction to calculating maximum FAPAR](./fapar_limitation.py).
+
+2. The maximum annual FAPAR and LAI can then be combined with daily predictions of
+   potential GPP to predict time series of realised FAPAR and LAI through the year,
+   [TBD].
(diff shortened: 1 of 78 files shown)


davidorme:
@MarionBWeinzierl - this touches `AnnualValueCalculator` more than `FaparLimitation`. I think it's a sensible approach to extending what we have to handle more than 1D inputs. We did have some handling of that, but it didn't quite connect up and now - to me at least - feels like a cleaner UI and code.

Sorry it's a bit of a big PR - several moving parts.

docs/source/users/phenology/overview.md
```diff
@@ -0,0 +1,27 @@
...
+
+1. Annual maximum values of FAPAR and LAI for a location can be predicted by comparing
+   constraints on potential gross primary productivity arising from either water
+   limitation or energy limitation {cite:p}`cai:2025a`. This is implemented in `pyrealm`
+   as the {class}`~pyrealm.phenology.fapar_limitation.FaparLimitation` class and is
+   described in the [introduction to calculating maximum FAPAR](./fapar_limitation.md).
+
+2. The maximum annual FAPAR and LAI can then be combined with daily predictions of
+   potential GPP to predict time series of realised FAPAR and LAI through the year,
+   [TBD].
```
MarionBWeinzierl: Does "TBD" refer to missing documentation or missing implementation? Might need more explanation.

pyrealm/phenology/fapar_limitation.py
```diff
@@ -1,51 +1,80 @@
-"""Class to compute the fAPAR_max and annual peak Leaf Area Index (LAI)."""
-
-from typing import Self
+"""Class to compute the fAPAR_max and annual peak Leaf Area Index (LAI).
+
+The :class:`FaparLimitation` class and the :meth:`FaparLimitation.from_pmodel` are
+designed to work with inputs that can have multiple dimensions. The first axis is
+_always_ assumed to represent a time series of annual observations. If the inputs
+are one dimensional, then this is a time series for a single site; if they are three
+dimnsional then these are observations for a grid of sites. Usually all array inputs
```
MarionBWeinzierl:
```suggestion
dimensional then these are observations for a grid of sites. Usually all array inputs
```

pyrealm/phenology/fapar_limitation.py
```diff
@@ -205,8 +269,8 @@ def from_pmodel(
 
         * The annual mean values :math:`c_a, \chi` and :math:`D` should be estimated
           during the growing season, so the ``growing_season`` argument must be used to
-          provide a boolean value indicating which observations should be treated as in
-          the growing season.
+          provide an array of boolean value indicating which observations should be
```
MarionBWeinzierl:
```suggestion
          provide an array of boolean values indicating which observations should be
```

tests/unit/core/test_annual_value_calculator.py
```diff
@@ -9,10 +9,33 @@
...
+            ("not", "a", "tuple", "of", "ints"),
+            None,
+            None,
+            False,
+            None,
+            pytest.raises(ValueError),
+            "The data_shape argument must a tuple of integers",
+            None,
+            id="data_shape_not_tuple_of_ints",
+        ),
```
MarionBWeinzierl, line 36: What is the difference between these two?

MarionBWeinzierl (commented):
Is is intended in the description above that it says that this PR fixes itself?

I've been through the code and it makes sense to me.

Some questions and typos:

docs/source/users/phenology/overview.md
davidorme: Missing implementation - have clarified.

tests/unit/core/test_annual_value_calculator.py
davidorme: We're looking for `(1,2,3)` - the first is just a string and the second is a tuple, but not of ints. Different failure modes on the same Raise.

davidorme:
I did not intend for it to close itself - updated!

MarionBWeinzierl (approved):

PR #564, opened 2025-10-14
New downloadable notebooks

### LINKED ISSUE #563: New approach to downloadable notebooks
@MarionBWeinzierl Really sorry about this - I've finally joined the dots in my head on a couple of things about working with notebooks in `sphinx`

So, we use the `mystnb` package within `sphinx` to run our Myst Markdown Jupyter notebooks to dynamically generate usage examples. That executes the notebooks and - it turns out - stores those in the built site under the `_download` directory as executed `.ipynb` notebooks. They use a hash in the path name:

```sh
> ls docs/build/html/_downloads/*          
docs/build/html/_downloads/062dc71f23bb507ee568bc0032351a32:
TModel.ipynb

docs/build/html/_downloads/091d0394b63a7219a45bdd9955f223ec:
OptimalChi.ipynb
```

It turns out that `mystnb` also provides an _incredibly_ poorly documented download role for those files:

https://github.com/search?q=repo%3Aexecutablebooks%2FMyST-NB%20nb-download&type=code

So - there's an automatically built version available for download. We then need to worry about the file links. I've stolen the idea of using the `pooch` package from `scikit` to provide a proof of concept `get_pyrealm_data` function. When that is run locally (e.g. in building docs) we can set an environment variable to just point to the local file, but otherwise `pooch` downloads and caches data files.

The end result is the same notebook source can be used to build a webpage _and_ provide a download link. We don't need to write separate standalone ones or keep them updated.

davidorme:
# Description

This PR:

* Adds the new `core.datasets` module providing the `get_pyrealm_data` command and some supporting objects and commands.
* Adds an admonition to `FaparLimitation.md` that should provide links to download it.
* Replaces the `importlib` stuff to get files with the `get_pyrealm_data` command.
* Mild testing.

This is a PR off #555, since that added a new notebook without a standalone version.

I've trialled it here:

https://pyrealm.readthedocs.io/en/563-new-downloadable-notebooks/users/phenology/fapar_limitation.html

There does seem to be an issue here. If I download that file then `jupyter-lab` objects to opening it. I think this is because we do not include `.ipynb` formats in our `jupytext` settings, so `jupyter-lab` with `jupytext` objects to opening a document that shouldn't exist. I'm not sure what happens if it is opened in a `jupyter-lab` environment that does not include `jupytext`.

Anyway - thoughts? 

Fixes #563

## Type of change

## Key checklist

## Further checks


MarionBWeinzierl:
Well, this approach will have the benefit that the links to classes and methods etc will work directly. For all the notebooks I created I had to do some "cleaning up" to get rid of them, or add full links.

I guess that means basically all the notebooks in https://github.com/ImperialCollegeLondon/pyrealm/tree/develop/docs/source/_static can be deleted then?

davidorme:
Hmm - hadn't thought about links. I don't think these are being resolved, actually.

ETA - I don't think it's possible. The `myst-nb` package _must_ run the notebooks first (otherwise there are no outputs to render) and then these are passed into `sphinx` to render the resulting document as HTML and resolve links.

So - I think we'd have to hard code links in the notebooks. At least now we have a `/stable/` URL that we can use to point to API docs?

MarionBWeinzierl:
I tried downloading it and opening it in Jupyter lab (with the environment as described [here](https://pyrealm.readthedocs.io/en/563-new-downloadable-notebooks/users/getting_started.html)), and it works for me without problems.

But you are right, the links don't resolve.

sjavis (commented):
The changes look good to me. There are just some links to the notebooks in 'getting_started.md' that still need to be updated.

MarionBWeinzierl:
> The changes look good to me. There are just some links to the notebooks in 'getting_started.md' that still need to be updated.

Thank you. Yes, I am currently considering what's the best way for treating these links is. It is not only the ones in `getting started`, but also those in the notebooks themselves, as discussed above.

MarionBWeinzierl:
> I tried downloading it and opening it in Jupyter lab (with the environment as described [here](https://pyrealm.readthedocs.io/en/563-new-downloadable-notebooks/users/getting_started.html)), and it works for me without problems.

I spoke too soon... It used to work, but now I am also getting problems with the jupyter notebooks, with an error as described by @davidorme above.  

I checked whether it makes a difference if I build the docs manually on my computer or use the one downloaded, and found something strange. Sometimes, when I click on the link in the docs it will show me the notebook as json file: 

image

Other times, it will download properly (even the same link), and I can then successfully open the jupyter notebook.

I will look further into this.

davidorme:
I've had the same thing happen. I thought it was browser specific but then found that _local_ docs opened as JSON but RTD docs downloaded in the same browser. 

We can always say to right click and "download as..." But that kinda clunks....

davidorme:
Watching a thunderstorm and torrential rain in the middle of the jungle 😄

MarionBWeinzierl:
> Watching a thunderstorm and torrential rain in the middle of the jungle 😄

I envy you being able to go out and see the real-world part of your work!

davidorme:
@MarionBWeinzierl I think the problem here is that all our notebooks should only use:

```yaml
jupytext:
  formats: md:myst
```

in the markdown YAML metadata. This is because we don't want to maintain paired ipynb notebooks - we build straight from Myst and don't want the other version cluttering the file structure (we can exclude them from the repo but it is clumsy. The issue with this for the downloads is that if a user has `jupytext` in their `jupyter-lab` setup, then `jupytext` sees the downloaded `ipynb` version and then the metadata saying MD only and complains about the unexpected format type. I think a corollary if this explanation is correct is that the files should load without issue in a jupyter-lab setup without jupytext - that's likely to be most users, but we still need to fix it.

I think the easiest solution is probably to hook onto the end of the sphinx build process and delete the `jupytext` metadata from those files. I've got a test implementation which seems to work for me - are you happy if I push it here?

MarionBWeinzierl:
> @MarionBWeinzierl I think the problem here is that all our notebooks should only use:
> 
> ```yaml
> jupytext:
>   formats: md:myst
> ```
> 
> in the markdown YAML metadata. This is because we don't want to maintain paired ipynb notebooks - we build straight from Myst and don't want the other version cluttering the file structure (we can exclude them from the repo but it is clumsy. The issue with this for the downloads is that if a user has `jupytext` in their `jupyter-lab` setup, then `jupytext` sees the downloaded `ipynb` version and then the metadata saying MD only and complains about the unexpected format type. I think a corollary if this explanation is correct is that the files should load without issue in a jupyter-lab setup without jupytext - that's likely to be most users, but we still need to fix it.
> 
> I think the easiest solution is probably to hook onto the end of the sphinx build process and delete the `jupytext` metadata from those files. I've got a test implementation which seems to work for me - are you happy if I push it here?

Sure, I hijacked this branch from you anyway.

davidorme:
> Sure, I hijacked this branch from you anyway.

I think we call that "collaborated on", right 😄  I've pushed it - can you see if that now works for you in downloading and launching?

MarionBWeinzierl:
I can download the notebooks now fine (that already worked in my last configuration yesterday). But (same as yesterday again), I can only open and run some of the notebooks (e.g. the C3/C4 competition one), but not others (e.g. the splash model one). Error message: `jupytext.paired_paths.InconsistentPath: Path 'docs/jupyter_execute/users/splash.ipynb' matches none of the export formats. Please make sure that jupytext.formats covers the current file (e.g. add 'ipynb' to the export formats)`

This might be directly related to jupyter lab, as I am able to open those in Google Colab.

davidorme:
@MarionBWeinzierl That's odd - `splash.ipynb` works for me. Just to check - have you run `make clean` and then `make html` to completely rebuild it all? I don't think that will matter as the new `jupyter_strip` action targets all download files, not just recently built, but just trying to clear up possible issues.

I have also just removed a couple of rogue markdown format settings, but not in that `splash` file.

MarionBWeinzierl:
OK, removing everything seems to have worked, I can open the splash file now. It is slightly worrying that the local build affects this (I thought I just wanted to test what was built and deployed online), but hopefully any users won't have spurious local files.

This now leaves the question of whether we want to change the links. They are actually quite nice, looking, for example, at the [tree crown page](https://pyrealm.readthedocs.io/en/563-new-downloadable-notebooks/users/demography/crown.html) with the mouse-over effects, the references etc (through the `{term}`, `{class}`, `{meth}` statements). Not sure whether it's quite the same to replace all of this through direct links to the `stable` page? Or leave it like it is and accept that the downloaded notebooks have some not-so-nice unresolved links?

davidorme:
> This now leaves the question of whether we want to change the links. They are actually quite nice, looking, for example, at the [tree crown page](https://pyrealm.readthedocs.io/en/563-new-downloadable-notebooks/users/demography/crown.html) with the mouse-over effects, the references etc (through the `{term}`, `{class}`, `{meth}` statements). Not sure whether it's quite the same to replace all of this through direct links to the `stable` page? Or leave it like it is and accept that the downloaded notebooks have some not-so-nice unresolved links?

I think accepting the downloads aren't quite so clean is the right decision. It's a lot of work to replace all the role links and look up the correct URLS and then maintain that. Those pages are _primarily_ documentation and we've made a good job of allowing people to download and run the code - I think we just accept that there are limitations?

sjavis (commented):
I agree with leaving the links as they are.

I'm happy with the changes and cannot see any issues with the code. But I am still unable to open the jupyter notebooks due to jupytext whether creating them locally or downloading them. For example, when I download the notebook from [here](https://pyrealm.readthedocs.io/en/563-new-downloadable-notebooks/users/pmodel/pmodel_details/worked_examples.html) it still seems to have the jupytext metadata.

davidorme:
Huh - good catch @sjavis. If I look at the build notes on RTD, then I see:

```
Jupytext stripping: no ipynb files found.
```

So - it's not targeting the right folder on RTD.

davidorme:
@sjavis I think that's fixed it - does it work for you now?

sjavis (approved):
That works now. Locally, I was trying to run the notebooks from `build/jupyter_execute/`, but I realised they work when running from `build/_downloads`.

PR #565, opened 2025-10-21
Phenology calculation - new version

### LINKED ISSUE #350: Calculate the steady state LAI.

### LINKED ISSUE #351: Calculate the realised LAI given a lagged process.

MarionBWeinzierl:
# Description

This is a new version of #535 (still draft to work from), cherry picking the changes from there onto a new branch based from `develop`.

Taken from #535:
* Adds a draft of the `Phenology` class , which:
    * Takes a dated set of daily GPP values and an existing `FaparLimitation` instance
    * Maps annual maximum LAI values onto the daily observations
    * Calculates steady state LAI and hence the lagged realised LAI using an exponential weighted average.
* Updates tests and adds a simple regression test of `Phenology` 
*  Get daily values from GPP estimated at different time resolutions.

Fixes #350 and fixes #351

Next step will be #566 

## Type of change

## Key checklist

## Further checks


davidorme:
> * Adds a year attribute to `FaparLimitation` to record a timestamp for each annual value.

This was actually already merged in to `develop` in #559, so we can drop it here.

> * Adds a draft of the `Phenology` class , which:
>   * Takes a dated set of daily GPP values and an existing `FaparLimitation` instance.

This is easy for a SubdailyModel because the values are already at a subdaily frequency and come with datetimes built in, so all we have to do is scale the GPP at subdaily observations from µgC m-2 s-1 to the total mol C day. 

It's more of a problem for the Standard P Model, which will probably be at coarser than daily frequency, and which doesn't require datetimes to fit. However, FaparLimitation already requires that users provide datetimes for Standard P Models to get annual values, so it is already part of the workflow. The more awkward problem is going from e.g. fortnightly observations to daily values for the phenology but that is basically just interpolation so we can use `scipy`. It's almost worth basically leaving it up to the user, but interpolating with datetime64 is mildly fussy.

>   * Maps annual maximum LAI values onto the daily observations
>   * Calculates steady state LAI and hence the lagged realised LAI using an exponential weighted average.
> * Updates tests and adds a simple regression test of `Phenology`

MarionBWeinzierl:
@davidorme in #535 you mention reviving `get_daily_values`. There is still `get_daily_means` in the `AcclimationModel`, is that what you mean?

davidorme:
The `AcclimationModel.get_daily_means` method specifically gets the mean values during the acclimation window - however we _could_ use it to get mean GPP from subdaily models. We'd just need to make a new Acclimation window that makes _every_ observation part of the acclimation window. That's sort of a hack, but also isn't - the machinery is there to split data up by day and calculate a summary statistic - it would just be a sort of "off-label" usage.

I'd feel happier about doing that if it were wrapped inside a user facing `get_daily_values` method as the implementation for SubdailyPModel inputs. As noted, for standard PModels we'd need dates and some specification of the interpolation scheme to be used. It might be cleaner to have these as separate functions  - the APIs will be quite different.

davidorme:
Ooops! Sorry!

MarionBWeinzierl:
I had almost forgotten that we also have the time_series.py core file, which currently only has the `AnnualValueCalculator` and the `broadcast_time` function. 

It might be the cleanest/most general option to implement a `DailyValueCalculator` which is able to take >daily as well as subdaily data and convert it to daily data. But I am not sure how to include the conversion factor for gpp that you mentioned. 

Alternatively, the `get_daily_values` (or even `get_daily_gpp`) function could sit in the pmodelABC, with implementations for the standard and subdaily pmodel.

davidorme:
> It might be the cleanest/most general option to implement a `DailyValueCalculator` which is able to take >daily as well as subdaily data and convert it to daily data. But I am not sure how to include the conversion factor for gpp that you mentioned.

I think the difference here is that the `AVC` does basically the same thing for both standard and subdaily P Model: it aggregates values from faster time scales to annual means or totals. It would be nice to have a single interface but the `DVC` basically has to do two totally different things and wrap them into a single API. For subdaily (which already has timestamps), that is aggregating subdaily observations to a single daily total. For standard P Models, that is requiring the timestamps and then interpolating to daily (by whatever means).

> Alternatively, the `get_daily_values` (or even `get_daily_gpp`) function could sit in the pmodelABC, with implementations for the standard and subdaily pmodel.

I think I like this quite a lot. It is potentially quite a useful general tool. Again the implementations are very different, but it is defined explicitly from the model class. We might want to implement `get_daily_means` and leave space for `get_daily_total`, but I don't think we need that now. We might also want to add a GPP conversion function to allow people to swap from µg C m2 s1 (the PModel output) and various other representations - we'd need mol C day but people often want µmol C m2 s-1. 

```python
gpp_moles_day = daily_mean_pmodel_gpp 
    * (24 * 60 * 60)  # seconds to day
    *  1e-6 # micrograms to grams
    / core_const.k_c_molmass # g to moles
)
```

The awkwardness for both models again comes in where people want to apply a posthoc penalty.  We'd have to include the same `gpp_penalty` argument that we use for calculating the annual values.

I'm beginning to wonder if  the PModelABC should have a `apply_gpp_penalty` method, that populates a `penalised_gpp` attribute that can then be picked up by downstream methods.

MarionBWeinzierl:
@davidorme, I implemented some very simple unit tests for the daily gpp, and for the subdaily data, I adjusted the existing gpp regression test to use the new function instead. For the test to work, I do not apply the gpp penalty. 

How do I best construct a regression test for the PModel data? We have the fortnightly data from the farparlim tests, but where would I get the golden dataset to compare against from for the daily interpolated data? Should it be the same target dataset as for the subdaily data, or does it not work that way?

davidorme:
> How do I best construct a regression test for the PModel data? We have the fortnightly data from the farparlim tests, but where would I get the golden dataset to compare against from for the daily interpolated data? Should it be the same target dataset as for the subdaily data, or does it not work that way?

OK - so the `python_implementation.py` file in the phenology build data directory does contain a calculation of the daily phenology predictions from the subdaily data:

https://github.com/ImperialCollegeLondon/pyrealm/blob/60bb551d008d1b7849f19123c95edaa6f218df98/pyrealm_build_data/phenology/python_implementation.py#L323-L350

That isn’t true for the fortnightly example - largely because I hadn’t committed to how we’d go from fortnightly to daily. This bit of it is new research in a sense, so we don’t really have a regression test. But the original implementation did make predictions from subdaily.

MarionBWeinzierl:
> > How do I best construct a regression test for the PModel data? We have the fortnightly data from the farparlim tests, but where would I get the golden dataset to compare against from for the daily interpolated data? Should it be the same target dataset as for the subdaily data, or does it not work that way?
> 
> OK - so the `python_implementation.py` file in the phenology build data directory does contain a calculation of the daily phenology predictions from the subdaily data:
> 
> https://github.com/ImperialCollegeLondon/pyrealm/blob/60bb551d008d1b7849f19123c95edaa6f218df98/pyrealm_build_data/phenology/python_implementation.py#L323-L350
> 
> That isn’t true for the fortnightly example - largely because I hadn’t committed to how we’d go from fortnightly to daily. This bit of it is new research in a sense, so we don’t really have a regression test. But the original implementation did make predictions from subdaily.

I have adapted the existing test for the subdaily data to use the new function here: https://github.com/ImperialCollegeLondon/pyrealm/pull/565/files#diff-1320757eb608b3eded7396f321b54b95785768c8991956324e159bf3e5925e12

pyrealm/pmodel/pmodel.py
```diff
@@ -93,6 +94,14 @@ def __init__(
...
         """The CoreConst instance used to create the model environment."""
 
+        self.gpp_penalty = (
+            (
+                (24 * 60 * 60)  # seconds to day
+                * 1e-6
+            )  # micrograms to grams
+            / self.core_const.k_c_molmass
+        )
+        """Penalty to apply to gpp when moving to daily gpp"""
```
davidorme: This isn't a penalty, it's just a conversion factor and one of many possible ones.

pyrealm/pmodel/pmodel.py
```diff
@@ -294,6 +307,10 @@ def summarize(self, dp: int = 2) -> None:
 
         summarize_attrs(self, self._data_attributes, dp=dp)
 
+    def apply_gpp_penalty(self, daily_mean_pmodel_gpp: NDArray) -> NDArray:
+        """Apply the gpp penalty to the input daily gpp."""
+        return daily_mean_pmodel_gpp * self.gpp_penalty
+
```
davidorme: Not used anywhere?

pyrealm/pmodel/pmodel.py
```diff
@@ -487,6 +504,33 @@ def _fit_model(self) -> None:
...
+            raise ValueError("Datetime sequence must be increasing")
+
+        # Validate that we are not dealing with subdaily data.
+        if spacing < np.timedelta64(1, "D"):
+            raise ValueError("This function expects datetime intervals > 1 day.")
+
+        gpp_data = pd.DataFrame(data={"gpp": self.gpp}, index=pd.to_datetime(datetimes))
+
+        # Interpolate to daily values
+        daily_gpp = gpp_data.resample("D").interpolate(method="linear")
```
davidorme: I've remembered why it was that I had an issue with `pandas` here. The interpolation is _only_ 1D along a `pandas.Series`. It's a fiddle to interpolate using scipy directly with a time series but it's the only way to interpolate along the first of multiple dimensions.

davidorme (commented):
On balance - and I realise I'm changing my tune here - I think this is adding quite phenology specific additional functionality into `PModel` and `SubdailyPModel`.

I think the simplest implementation is probably to have `FaparLimitation` as is with the `__init__` and `from_pmodel` methods and then `Phenology` (or maybe `DailyPhenology`) with `__init__` and `from_pmodel`, where `from_pmodel` really just does the resampling according to model input, converts to moles and calls its own `__init__` with the result.

There _is_ overlap between those two classes, but they are conceptually separate, so I think keeping them as a two step process works ok. I think there is more general functionality that we can pluck out of this, but it feels like we're circling lots of possible solutions.

pyrealm/pmodel/pmodel.py
MarionBWeinzierl: Yeah, I had it applied to the daily gpp at the averaging step, but the results did then no longer match those in the regression test.

pyrealm/pmodel/pmodel.py
MarionBWeinzierl: Is it ok to use it here, though?

tests/regression/phenology/test_phenology_gpp.py
```diff
@@ -74,17 +75,30 @@ def test_phenology_gpp_calculation(
...
+    )
+    fortnightly_datetimes = de_gri_fortnightly_data["time"]
+    de_gri_pmodel = PModel(env=fortnightly_env)
+    pmodel_gpp_from_fortnightly = de_gri_pmodel._get_daily_gpp(
+        datetimes=fortnightly_datetimes
+    )
+
+    assert_allclose(
+        de_gri_fortnightly_daily_outputs["daily_gpp"],
+        pmodel_gpp_from_fortnightly[:-1],
```
MarionBWeinzierl: I am not sure why this is needed. The arrays seem to be off.

pyrealm/pmodel/pmodel.py
davidorme: No, sadly not. If you try and use that with e.g. a 3x3 spatial grid over time, then it throws up its hands and whines that it can only handle 1D data. The direct usage of `scipy` in the `python_implementation.py` with all the casting to and from integer _does_ handle the ND case.

tests/regression/phenology/test_phenology_gpp.py
davidorme: The assertion, or the `[:-1]`?

MarionBWeinzierl: the `[:-1]`

MarionBWeinzierl: It tells me the arrays are off by one if I don't add it, but still tells me the same if I use `[:-2]`

MarionBWeinzierl: Why is one array longer?

davidorme: Not sure - my guess would be the `np.range(from, to)` not including the final `to` endpoint?

MarionBWeinzierl: Ah, looked at the wrong `_get_daily_gpp` function. I assume it might be that it is off by one due to the interpolation, leading to an array that is at the `n-1` interpolation points. That should be resolved when we no longer use the pandas interpolation, but the method you use to create that array - is that doing some sort of extrapolation at the end points?

pyrealm/pmodel/pmodel.py
MarionBWeinzierl: And we are happy to use the legacy interp1d method? I remember that we discussed moving away from that.

davidorme: I think so yes - it is used elsewhere, so if we switch it at some point, we can do it across the board. And honestly, it's an easier API than what replaces it!

tests/regression/phenology/test_phenology_gpp.py
davidorme: I think the issue is that we're interpolating onto daily timesteps between the start and endpoints, but the end isn't included in the list of observations:

Instead of:

```python
daily_timestamps = np.arange(time[0], time[-1], np.timedelta64(1, "D"))
```

We could do:

```python
daily_timestamps = np.arange(time[0], time[-1]+np.timedelta64(1, "D"), np.timedelta64(1, "D"))
```

That should ensure that the set of days includes both ends?

davidorme (approved):
Sorry for the delay on this. I've now got the code for the alternative implementation and I think we'll need to refactor to bring that in, but I think this is a good point to bank the work we've got and then build on that.

Issue #262, opened 2024-07-17
Incorrect display of function/class argument types in documentation.


j-emberton:
**Describe the bug**
Incorrect display of function/class argument types in documentation.

**Expected behavior**
I would expect to just see a display of arguments with a simple type e.g. my function(arg1: NDArray)

**Screenshots**
<img width="750" alt="Screenshot 2024-07-17 at 15 07 25" src="https://github.com/user-attachments/assets/8932fb29-a267-4202-b8b3-b23588a50f22">

**Desktop (please complete the following information):**

- OS: MacOS 14
- Safari

**Additional context**

j-emberton:
Possible solution:

Create a python file with the following contents and add to sphinx project:

`# sphinx_initvar_filter.py
from sphinx.ext.autodoc import ClassDocumenter, ModuleLevelDocumenter
from sphinx.util.inspect import getdoc
from sphinx.util import logging

logger = logging.getLogger(__name__)

class InitVarFilterDocumenter(ClassDocumenter):
    def add_directive_header(self, sig):
        self.options.noindex = True
        super().add_directive_header(sig)
        self.options.noindex = False

    def filter_members(self, members, want_all):
        filtered_members = []
        for name, obj in members:
            if not name.startswith('__') and 'InitVar' not in str(obj):
                filtered_members.append((name, obj))
            else:
                logger.info(f'Filtering out: {name} ({obj})')
        return filtered_members

def setup(app):
    app.add_autodocumenter(InitVarFilterDocumenter)`
    

Then modify conf.py to include the custom operation:

`# conf.py
extensions = [
    'sphinx.ext.autodoc',
    'sphinx_autodoc_typehints',
    'sphinx.ext.napoleon',
    'sphinx_initvar_filter'  # Add your custom extension here
]

autodoc_typehints = "description"
autodoc_typehints_format = "short"`

j-emberton:
Hasn't been tried and tested yet

j-emberton:
also more in pressure submodule:

<img width="729" alt="Screenshot 2024-07-18 at 14 22 46" src="https://github.com/user-attachments/assets/97326566-bd14-47a2-be34-9bc483a6f6ff">

davidorme:
This looks relevant:
https://stackoverflow.com/questions/67473396/shorten-display-format-of-python-type-annotations-in-sphinx

j-emberton:
Nice find.

is there a simple way to build the docs locally to test the changes?

[...]

PR #571, opened 2025-11-06
xarray core utils functions

### LINKED ISSUE #570: `xarray` core utils functions
In #43 we want to allow xarray inputs by converting them to numpy arrays immediately when they are passed. However, there are a lot of simple functions in `core/` that take and return arrays. Rather than stripping the `xr.DataArray`s in these functions it could be useful to allow pass-through of the DataArrays.

Functions whose return type depends on the input functions. For most functions this will already happen at runtime, they will just need to have the type hints updating.

For the static typing there are two possibilities - using `TypeVar` to match the output type to the input type(s), or using function overloads. The former requires minimal changes, but overloads will allow a mix of `xr.DataArray` and `np.array` inputs as well as being required if `isinstance` checks are needed.

sjavis:
# Description

This adds xarray support for most functions in 'core' modules such that if DataArrays are provided as inputs then DataArrays are returned. For the most part the functions themselves do not need to be significantly altered, but the static typing needs to be adjusted.

I have done this in two ways: with `TypeVar` and overload decorators. TypeVar is simpler and doesn't clutter the files and documentation as much, so I have preferred this. But it doesn't support a mix of input types and doesn't work if there is different logic required for the different types.

I have used overloads where necessary which allows a mix of numpy array and DataArray inputs, and will return DataArray if any of the inputs are DataArrays. This is done in 'utilities.py' where the functions use `isinstance` checks dynamically, and in 'solar.py' since some functions take both float and int arrays (and TypeVar functions seem to have problems calling overloaded functions so I just overloaded them all).

This PR will just merge the changes back into the main xarray branch, but I think this needs a separate review for if we want this.

Fixes #570

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/core/bounds.py
+++ b/pyrealm/core/bounds.py
@@ -29,7 +29,8 @@ from typing import Any, ClassVar
 from warnings import warn
 
 import numpy as np
-from numpy.typing import NDArray
+
+from pyrealm.core.xarray import ArrayTypeVar
 
 
 @dataclass
@@ -119,9 +120,7 @@ class BoundsChecker:
 
         self._data[bounds.var_name] = bounds
 
-    def check(
-        self, var_name: str, values: NDArray[np.floating]
-    ) -> NDArray[np.floating]:
+    def check(self, var_name: str, values: ArrayTypeVar) -> ArrayTypeVar:
         r"""Check inputs fall within bounds.
 
         This method checks whether the provided values fall within the bounds specified
--- a/pyrealm/core/hygro.py
+++ b/pyrealm/core/hygro.py
@@ -6,17 +6,17 @@ and relative humidity. The implementation is drawn from and validated against th
 ``bigleaf`` R package.
 """  # noqa: D205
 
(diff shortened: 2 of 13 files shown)


sjavis:
For context, the overloads look like this in sphinx:
image
Whereas the TypeVar approach looks like this (with ArrayTypeVar defined in the link):
image

MarionBWeinzierl:
> For context, the overloads look like this in sphinx: image Whereas the TypeVar approach looks like this (with ArrayTypeVar defined in the link): image

Maybe we need to dig out issue https://github.com/ImperialCollegeLondon/pyrealm/issues/262 at some point

davidorme:
> > For context, the overloads look like this in sphinx: 
> Maybe we need to dig out issue #262 at some point

Yeah - I've been running into this a lot recently. I really dislike how sphinx configures things like this - it is incredibly opaque.

sjavis:
Closing as we have decided not to go ahead with this, as described in #570.

Issue #566, opened 2025-10-22
Add higher-level from_pmodel function to fapar_lim and phenology


MarionBWeinzierl:
Taking @davidorme 's note from #535:

> I'm not sure about the structure of this yet - we need some kind of from_PModel method, but I think that should be higher level. The API for a PModel should really do both the FaparLimitation and Phenology calculation in one pass - we need to keep the inputs aligned and having it as a two stage thing just makes life more complex for us and users.

MarionBWeinzierl:
Hm, sorry if that is a stupid question, but what is exactly the system for entities to end up in PModel/SubdailyPModel, FaparLimitation or Phenology? I seems like some of those (like gpp) could potentially "live" in any of those classes? 

And could FaparLimitation, instead of being its own class, be a function in phenology? That seems to make sense, logically. Or would that blow up phenology too much?

davidorme:
> Hm, sorry if that is a stupid question, but what is exactly the system for entities to end up in PModel/SubdailyPModel, FaparLimitation or Phenology? 

I don't think we have a system - the growth of the package has been fairly research lead and organic so bits just get added in as they fit with the existing structure. That's not quite fair - we extensively remodelled the core PModel and SubdailyPModel for 2.0.0 - but there's nothing formal. Is that something that exists in commercial software development?

> I seems like some of those (like gpp) could potentially "live" in any of those classes?

I think the class structure here does have a clear hierarchy. The `PModel` and `SubdailyPModel` classes predict GPP (and other things) for a set of observations, so GPP is a clear member of those classes. We then want to do things with those predictions - in the case of the `phenology` module, that is working out core annual stats from GPP and precipitation and then predicting a daily time series of leaf area.

> And could FaparLimitation, instead of being its own class, be a function in phenology? That seems to make sense, logically. Or would that blow up phenology too much?

I think `FaparLimitation` makes sense as a class rather than a function - it has quite a lot of attributes, so it would be quite a complex return argument. I also think it is possible that people would want to calculate those limitation terms independently of a Phenology time series, but that is a probably rare use case. So having Phenology calculate those limitation terms internally, rather than users having to pre-calculate them, would be a better user experience, I think. The Phenology class could then store a `FaparLimitation` object as part of its own attributes. But I think we get everything working first?

MarionBWeinzierl:
> But I think we get everything working first?

[...]

PR #575, opened 2025-11-21
Alternative implementations of water density and viscosity

### LINKED ISSUE #573: Simpler implementations of water density and viscosity.
We're using some ultra precise formulations for water density and viscosity that are computationally intensive. This issue is to add the Kell and Vogel formulations for those quantities and to assess the impact on predicted GPP.

davidorme:
# Description

The `pyrealm` package inherited the `fisher` and `huber` methods for calculating density and viscosity from `rpmodel` and they are wildly more precise than anyone is likely to need. We did have alternatives but with a rather inconsistent setup:

* Density had the `CoreConst.water_density_method` attribute that allowed users to switch between `fisher` and `chen`,  and the `calc_density_h2o` function detected that and used it.
* Viscosity just had `huber` that had an alternative internal `simple` implementation that was controlled by the `CoreConst.simple_viscosity` attribute.

This PR:

*  Adds a few additional approaches for each function, mostly from a raid on a compendium of viscosity and density calculations (https://library-search.imperial.ac.uk/discovery/fulldisplay?docid=alma9956076101801591&context=L&vid=44IMP_INST:ICL_VU1&lang=en&search_scope=MyInst_and_CI&adaptor=Local%20Search%20Engine&tab=Everything&query=any,contains,Viscosity%20Liquids%20Theory%20Estimation%20Experiment&offset=0)
* Wraps those into registries - using a decorator to add decorated functions to the registry after checking their signature.
* Add/updates wrapper functions (`calculate_viscosity_h20` and `calculate_density_h20`) to look at the `CoreConst.water_density_method` and new `CoreConst.water_viscosity_method` attributes. The string in those attributes is used to recover the appropriate function from the signature and calculate the values.
* Adds a new doc page showing the differences.
* Updates the tests to check the registration process and some simple value checks.

Notes:

1. One oddity here is that all the functions require `patm`, even though not all of the algorithms use it. This is purely to keep a single simple API / signature for using the functions.

2. I think we should switch over to using the low complexity implementations as the default but I do want to discuss it with the user community first. We'd need to update the unit test values (or configure the tests to use non-default methods). We can do that in a later PR.

Fixes #573

## Type of change

## Key checklist

## Further checks


pyrealm/pmodel/__init__.py
```diff
@@ -42,7 +42,6 @@
     calc_ns_star,
     calc_soilmstress_mengoli,
     calc_soilmstress_stocker,
-    calc_viscosity_h2o,
```
MarionBWeinzierl, line 45: Do we not need `calculate_viscosity_h2o` here instead? Why is this not imported here, but in `functions`?

pyrealm/pmodel/__init__.py
```diff
@@ -42,7 +42,6 @@
     calc_ns_star,
```
MarionBWeinzierl, line 42: `calc_viscosity_h2o` was changed to spell out `calculate` -- should we do the same with all the other functions, for consistency?

MarionBWeinzierl (approved):

pyrealm/core/water.py
```diff
@@ -1,79 +1,243 @@
...
+)
+"""The expected signature for registered water density functions, as tuples of parameter
+name and type annotation."""
+
+
+def register_density_method(method_name: str) -> Callable:
+    """Registration decorator for water density functions.
+
+    Functions decorated with ``register_density_method`` are automatically added to the
+    ``DENSITY_METHODS`` registry when imported, using the the provide name as a key.
```
sjavis:
```suggestion
    ``DENSITY_METHODS`` registry when imported, using the provided name as a key.
```

pyrealm/core/water.py
```diff
@@ -155,68 +308,286 @@ def calc_density_h2o(
...
+"""The expected signature for registered water density functions, as tuples of parameter
+name and type annotation."""
+
+
+def register_viscosity_method(method_name: str) -> Callable:
+    """Registration decorator for water viscosity functions.
+
+    Functions decorated with ``register_viscosity_method`` are automatically added to
+    the ``VISCOSITY_METHODS`` registry when imported, using the the provide name as a
+    key.
```
sjavis:
```suggestion
    the ``VISCOSITY_METHODS`` registry when imported, using the provided name as a key.
```

pyrealm/core/water.py
```diff
@@ -155,68 +308,286 @@ def calc_density_h2o(
...
+expected to have the same signature."""
 
-def calc_viscosity_h2o(
-    tc: NDArray[np.floating],
+VISCOSITY_FUNCTION_SIGNATURE: tuple[tuple[str, str], ...] = (
+    ("tk", "NDArray[np.floating]"),
+    ("patm", "NDArray[np.floating]"),
+    ("core_const", "CoreConst"),
+)
+"""The expected signature for registered water density functions, as tuples of parameter
```
sjavis:
```suggestion
"""The expected signature for registered water viscosity functions, as tuples of parameter
```

pyrealm/core/water.py
```diff
@@ -155,68 +308,286 @@ def calc_density_h2o(
...
+) -> NDArray[np.floating]:
+    r"""Calculate viscosity of water using the Daubert Danner form.
+
+    This function calculates water viscosity (:math:`\mu`, Pa s) as a function of
+    temperature (:math:`T`, K) using the Daubert and Danner form from Equation 4.41 of
+    :cite:`viswanath:2007a`.
+
+    .. math::
+
+        \mu =  \exp \left( a + \frac{b}{T} + c \ln T  + dT^10 \right
```
sjavis:
```suggestion
        \mu =  \exp \left( a + \frac{b}{T} + c \ln T  + dT^10 \right)
```

sjavis (approved):
I have a few suggestions for issues in docstrings but it looks good to me otherwise.

Also, I realise that there is already a list of water and density methods in `water.md` despite what I had mentioned in our meeting earlier.

pyrealm/pmodel/__init__.py
davidorme: Honestly not sure. The original names come from `rpmodel` but I dislike abbreviated function names. So my inclination is yes, but it will be a breaking change - that could be forgiven as a `2.0.1` update, I think?  The alternative would be to move the new code back to `calc_` but I don't like it as much.

Thoughts?

MarionBWeinzierl: Hm, I have no strong feelings of abbreviating this or not (it seems relatively clear even in abbreviated form). I just don't like to have both versions around. Do you think this would be a separate issue/PR to fix later?

sjavis: I also slightly favour the `calculate_` form. I think if it is to be changed it should be now while v2 is new.

davidorme: We have quite a few abbreviated forms:

```python
16 results - 4 files

pyrealm/core/hygro.py:
   16  
   17: def calc_vp_sat(
   18      ta: NDArray[np.floating], core_const: CoreConst = CoreConst()

  202  
  203: def calc_saturation_vapour_pressure_slope(
  204      tc: NDArray[np.floating],

  226  
  227: def calc_enthalpy_vaporisation(tc: NDArray[np.floating]) -> NDArray[np.floating]:
  228      """Calculate the enthalpy of vaporization.

  243  
  244: def calc_specific_heat(tc: NDArray[np.floating]) -> NDArray[np.floating]:
  245      """Calculate the specific heat of air.

  275  
  276: def calc_psychrometric_constant(
  277      tc: NDArray[np.floating],

pyrealm/core/pressure.py:
  10  
  11: def calc_patm(
  12      elv: NDArray[np.floating], core_const: CoreConst = CoreConst()

pyrealm/pmodel/functions.py:
  158  
  159: def calc_ftemp_inst_rd(
  160      tc: NDArray[np.floating],

  198  
  199: def calc_gammastar(
  200      tk: NDArray[np.floating],

  255  
  256: def calc_ns_star(
  257      tk: NDArray[np.floating],

  298  
  299: def calc_kmm(
  300      tk: NDArray[np.floating],

  368  
  369: def calc_soilmstress_stocker(
  370      soilm: NDArray[np.floating],

  480  
  481: def calc_soilmstress_mengoli(
  482      soilm: NDArray[np.floating] = np.array(1.0),

  579  
  580: def calc_co2_to_ca(
  581      co2: NDArray[np.floating], patm: NDArray[np.floating]

pyrealm/pmodel/isotopes.py:
  102  
  103:     def calc_c4_discrimination(self, pmodel: PModel) -> None:
  104          r"""Calculate C4 isotopic discrimination.

  139  
  140:     def calc_c4_discrimination_vonC(self, pmodel: PModel) -> None:
  141          r"""Calculate C4 isotopic discrimination.

  193  
  194:     def calc_c3_discrimination(self, pmodel: PModel) -> None:
  195          r"""Calculate C3 isotopic discrimination.
```

I think we should change them, but perhaps not in this PR?

pyrealm/pmodel/__init__.py
davidorme: Honestly, I don't think this `__init__` is at all well thought through. When the module was small and didn't have too many submodules, it made sense to import things in `__init__` to shorten the namespace, but now we have an idiosyncratic selection. If we import all the public objects here, it would be very crowded, but it isn't clear what the "correct" selection would be!

MarionBWeinzierl: See my comment [here](https://github.com/ImperialCollegeLondon/pyrealm/issues/566#issuecomment-3627366488) - it almost feels like, now that this part of the software is growing, we need to put some sort into a potential restructuring of the class structure.

pyrealm/pmodel/__init__.py
davidorme: See #580

PR #582, opened 2025-12-08
Align PModel methods

### LINKED ISSUE #579: SubdailyPModel not handling Arrhenius method environment arguments
**Describe the bug**

From @Shirleycwj 

> I'm testing different functions in subdaily P model to build up a standard form of subdaily version to would account for both effect of temperature (on phi0) and soil moisture without overlapping, and one of my attempt is to use peaked Arrhenius function and replace Giulia's soil moisture with David's temperature-AI-dependent phi0. 

> However I received an unexpected error that mean_growth_temperature doesn't seem to be passed to the subdaily environment although I provided in the environment variable (please see below screenshot). I'm calculating the Arrhenius function manually as a temporary workaround but I was wondering if you had the same issue when you run the function? I'm using Pyrealm 2.0.0 rc5 at the moment, not sure if there's other updates I missed that might led to this error?

Image

This is because the daily data compilation for the daily optimal model is not including required variables from the `_arrhenius_class` instance used in the model, only the `_optchi_class` and `_kphio_class`.

https://github.com/ImperialCollegeLondon/pyrealm/blob/50ace2feb4de5ca16cb03a1372e5dbae72b633bb/pyrealm/pmodel/pmodel.py#L771-L780

I think it's worth synchronising the `requires` and `required_env_variables` names used in these classes, and probably also adding `_jmaxlim_class`: although at present no JmaxLim subclasses use additional environment variables, it would then make these `method_xyz` class consistent.

davidorme:
# Description

This PR:

* Updates the four PModel `method_` classes (Arrhenius, OptimalChi, QuantumYield and JMaxLimitation).
  * The first three now all use `required_env_variables` as class attributes. I've moved them from lists to tuples, since there's no reason at all for them to be mutable.
  * The JMaxLimitation class is converted from `dataclass` to normal to match the other three, but doesn't require any additional variables, so is left as is.
* The SubdailyPModel code fixes the motivating bug - the daily acclimation window Arrhenius required variables were not being passed in to the optimal daily PModel.
* It adds a test that checks all combinations of the model work OK in both the PModel and SubdailyPModels.

Fixes #579

## Type of change

## Key checklist

## Further checks


tests/unit/pmodel/test_pmodel.py
```diff
@@ -0,0 +1,88 @@
...
+            )
+
+            _ = SubdailyPModel(
+                env=env,
+                acclim_model=acclim,
+                method_arrhenius=arh_meth,
+                method_optchi=chi_meth,
+                method_kphio=phi0_meth,
+                method_jmaxlim=jmax_meth,
+            )
```
sjavis, line 105: My understanding is that this test will check that the `required_env_variables` are defined correctly in each method so that it works for every combination. As well as checking that they generally initialise correctly. Is this right?

If so, would it be better to move the definition of the variables and `PModelEnvironment` into the loop so that it only checks the required variables for that specific combination? Otherwise the test may miss if a method is missing some `required_env_variables` if they are defined in another method.

sjavis (commented):
The changes look sensible to me and I can't spot any issues with them. I just have one comment about the test.

tests/unit/pmodel/test_pmodel.py
davidorme: My original thinking there was that the individual methods don't care about extra variables in `env`, they only need to check that their own required variables are present. So the data contains all possible required variables, but the test then checks that each method can access and use just the specific ones it requires. Moving that inside the loop would just be recreating multiple subsets of the current superset.

sjavis: I suppose my thought is that since it doesn't care about the extra variables but it does care about missing variables it is more relevant to check they work with the subset than the superset. Although if recreating the environment each time causes it to take a long time it might not be worth it. What are your thoughts?

davidorme: It's probably doubling the runtime on the test. Admittedly that's doubling a small number! I've made the switch.

sjavis (approved):
This all looks good to me!

PR #584, opened 2025-12-18
Fix regression test timing bias

### LINKED ISSUE #578: Regression test biased timing
**Describe the bug**
The `performance_regression_testing.sh` script is consistently taking longer for the "new" version than the "old" version by about 15% on average. This remains true even if the "old" and "new" versions are switched.

**Possible causes**
There are two possibilities for the cause:
- The old run uses a git worktree whereas the new run uses the current repo.
- The same virtual environment could be being used, causing caching issues.

sjavis:
# Description

This fixes the issue with the second profiling run consistently taking longer than the first profiling run when averaged over multiple runs (by about 10%). It is not entirely clear what the cause is or if it is consistent across devices but it seems to result from using the same virtual environment for the two runs. So I have ensured they are equivalent by creating new virtual environments for each version — before it would inherit the virtual environment from the calling shell. I have also made them both use worktrees to keep them equivalent and so they don't change the local repo.

I have also added the advanced function-by-function benchmarking to the regression test script since this will presumably face the same issue. It also simplifies it so that you don't need to manually perform the profiling on two versions. The new usage is simply to use the `-a` option (for advanced) when calling the script, i.e.:
```bash
./performance_regression_checking.sh -n [NEW-COMMIT] -o [OLD-COMMIT] -a
```
I have modified the profiling documentation to describe this and hide the details of how to do it manually in a dropdown.

There is still an issue that the variability is high in the profiling times. So it is necessary to run it multiple times to get an accurate idea of the change in performance.

I also updated how the advanced benchmarking parses the profiling results into a pandas dataframe. Before this was using `pd.read_fwf`, assuming the results were in a fixed width table. But this would break when there were a large number of calls to a function.

- [x] Use worktrees for both the older and newer versions.
- [x] Use separate virtual environments (don't inherit from calling shell).
- [x] Add advanced benchmarking option to the script.
- [x] Update profiling documentation.
- [x] Check if scaling up profiling reduces variability.

Fixes #578

## Type of change

## Key checklist


sjavis:
I attempted to scale up the profiling using pytest-repeat to repeat the runs multiple times. However this did not seem to significantly decrease the variablility (measured by STDEV/MEAN). If anything it increased it. So I have left this out and instead added a disclaimer to the docs that the variability may impact the effectiveness of the advanced regression testing.

docs/source/development/profiling_and_benchmarking.md
```diff
@@ -161,10 +154,29 @@ time is spent during the test runs, which identifies bottlenecks, but it is also
...
+./performance_regression_checking.sh -n [NEW-COMMIT] -o [OLD-COMMIT] -a
+```
+
+This will check if any function is more than a certain tolerance (5% by default) slower
+than the old version. It will produce a plot similar to that below in
+`prof/performance-plot.png` showing how the cumulative time per call has changed between
+the two versions.
 
+:::{note}
+Variance in the run time can be significant when looking at individual functions. This
```
MarionBWeinzierl, line 175: Should we add a note that increasing the load might improve that?

MarionBWeinzierl (approved):
These changes can be merged, maybe extending the note on the performance variability. It would still be worthwhile looking into a larger test set to reduce the effect of I/O on the runtime.

sjavis:
I have been looking into the problem size and variable timings a little more and I have noticed a few things:

- I tried increasing the system size used for the profiling using the `--pmodel/splash-profile-scaleup` options. However, like using pytest-repeat, this didn't seem to have any noticeable reduction to the variability (measured by standard deviation / mean of the run times).

- When running on exclusive nodes on CSD3 there is significantly less variability. So it might just be required to run it on a cluster to get accurate performance.

- Unlike running locally where the first run is faster than the second, the first run is slower on CSD3, even when creating separate virtual environments. I'm still not certain of the cause but I think creating separate virtual environments might have simply been spacing out the two runs, hence reducing the impact of the first on the second.

- I have instead tried to do an initial 'warm up' run of the profiling before doing the actual profiling. The results for this below (using the pmodel profiling with different values for `--pmodel-profile-scaleup` running on CSD3) shows that the timing of the first run and second run are now very similar with the first run only <= 2% slower than the first (plots 3 and 4). It also shows the much lower variability (standard deviation about ~1% of the mean). I need to test if doing a warm up run will work on my laptop.

image


- Also, while investigating the variability I spotted that there is quite significant temporary memory allocations in `estimate_aet` in `splash/evap.py`. Shown below using [memray](https://bloomberg.github.io/memray/getting_started.html) using [`memray summary --temporary-allocations`](https://bloomberg.github.io/memray/temporary_allocations.html). This probably isn't the underlying problem and is probably a separate issue, but I thought it was worth mentioning.
image

davidorme:
> * Also, while investigating the variability I spotted that there is quite significant temporary memory allocations in `estimate_aet` in `splash/evap.py`.

I hadn't seen `memray` before - that's really useful. The `splash` code does have a large number of intermediate variables - my guess here is that this is behind that. It calculates a whole bunch of things in one go - the old version didn't try and save anything so probably had a tiny memory usage but a horrible runtime. There might be some kind of intermediate sweet spot!

docs/source/development/profiling_and_benchmarking.md
sjavis: I've added a scaling argument to the script and suggesting using this or exclusive cores on a cluster to reduce the variability.

sjavis:
This should hopefully now be ready. I have added the warm-up step to the profiling and also added an `-s` option to the script to allow the scaleup arguments to be changed (it multiplies by the defaults).

The warm-up run also seems to work when I run it locally, just not quite to the same degree as on the cluster. The results for this are below. For a pmodel scaleup of 6, the standard deviation is about 8% of the mean time and the second run is only about 2% slower - as opposed to the original 10-15% slower (and this wasn't using separate virtual environments which may reduce it further).

pyrealm-variability-prerun-local

MarionBWeinzierl (approved):

PR #585, opened 2025-12-18
New downloadable notebooks

### LINKED ISSUE #563: New approach to downloadable notebooks

davidorme:
# Description

> [!TIP]
> This is a repeat of #564, which was accidentally merged down into #555 not `develop`. 
> 
> The original PR was forked off #555 as a demo, but #555 was closed before this one, so the merge was down into a dead branch. This is now targeted at `develop` instead.

This PR:

* Adds the new `core.datasets` module providing the `get_pyrealm_data` command and some supporting objects and commands.
* Links to the `ipynb` files automatically generated during the `sphinx` build using the `download` role. This replaces the statically maintained parallel `ipynb` versions we had before.
* Replaces the `importlib` stuff in the workbooks to get files with the more portable `get_pyrealm_data` command.
* Processes the automatically generated `ipynb` file to strip out `jupytext` settings. These are a problem if a user _does_ have `jupyter-lab` with `jupytext` as the `ipynb` format is not included in our `jupytext` settings (we only use markdown notebooks in the docs source) and `jupytext` objects to what it sees as a rogue format.

TODO - currently no testing. To do a sensible job, we'd probably need to mock the `pooch.Pooch.fetch` calls (and probably provides some way to set the cache directory rather than just using the system cache within tests).

> [!IMPORTANT]
> We still need to decide when to run the `core.datasets._populate_pooch_registry()` method?
> 
> This automatically creates the links that `pooch` uses to get the raw objects from `repo` - it's only _really_ needed when a new release is made. It seems like overkill to run it on every commit. We can run it as:
>
>  ```sh
>  python -m pyrealm.core.datasets
>  ```

Fixes #563 

## Type of change

## Key checklist

## Further checks


docs/source/conf.py
```diff
@@ -263,3 +270,59 @@ def setup(app):  # type: ignore
...
+
+    app.connect("build-finished", strip_jupytext)
+
+
+def strip_jupytext(app, exception):  # type: ignore
+    """Remove jupytext metadata from download notebooks.
+
+    The build process outputs ipynb versions of the Markdown notebooks to the
+    `build/html/_downloads` directory, which we then link in for users to download.
+    However, the files contain the original `jupytext` metadata that specifies _only_
```
MarionBWeinzierl: I do not understand the second half of the sentence - should it be "the files containing [...]"?

pyrealm/core/datasets.py
```diff
@@ -0,0 +1,131 @@
...
+    package. These datasets are not provided in the binary build of :mod:`pyrealm` so
+    are only locally available from `sdist` installations or from within a ``git`` clone
+    of the package repository. A list of available files is maintained in the
+    ``pyrealm_data_registry.json`` file within the :mod:`pyrealm.core` module and
+    managed through the :data:`DATASETS` file manager.
+
+    The function runs in one of two ways:
+
+    * If ``use_resources=True`` is used or the ``PYREALM_USE_LOCAL_DATA`` environment
+      variable is set, then the function using :func:`importlib.resources.files` to
```
MarionBWeinzierl, line 103:
```suggestion
      variable is set, then the function is using :func:`importlib.resources.files` to
```

?

MarionBWeinzierl (approved):
This looks ok to me, just two docstrings that I do not fully understand

docs/source/conf.py
davidorme: I've rewritten that docstring to be much clearer about how the issue arises and why the function is needed.

PR #588, opened 2026-01-05
Remove overzealous warnings formatting

### LINKED ISSUE #587: Stop hijacking the warnings formatter
The `pyrealm.__init__.py` file sets a format for warnings. After `pyrealm` is imported that basically sets the warning formatting for _any_ warning issued by Python and obfuscates the source of the warning.

I think I setup this up very early, when I was getting a lot of numpy and other warnings. These warnings are now largely correctly handled (either ignored as not relevant or rewritten to handle edges cases leading to warnings) and so I think the correct thing to do is delete this formatting.

davidorme:
# Description

This PR removes code from `pyrealm.__init__.py` that was intended to provide package specific warning formatting but actually globally changes formatting and obscures warning sources. The intended package functionality is not really needed any more.

Fixes #587 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/__init__.py
+++ b/pyrealm/__init__.py
@@ -1,7 +1,4 @@
-"""The pyrealm package.
-
-TODO: complete this documentation
-"""
+"""The pyrealm package."""
 
 import importlib.metadata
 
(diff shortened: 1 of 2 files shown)


MarionBWeinzierl (approved):
This makes sense. Maybe, while updating this file, the todo in line 3 can be removed/updated?

PR #591, opened 2026-01-17
Add golden datasets for Zhu phenology method

### LINKED ISSUE #590: Add golden datasets for the Zhu phenology method
We now have code for the alternative phenology implementation by Ziqi Zhu. The first step for that is to generate golden datasets using that code to use for regressions tests in factoring the new approach into `pyrealm`.

davidorme:
# Description

This adds a new directory to `pyrealm_build_data` that contains Python code from Ziqi Zhu and  Mateusz Lisiewski that implements Ziqi's approach to calculating daily LAI from P Model predictions. It also uses the existing regression datasets to generate some golden outputs from that code:

* Annual fAPAR max values for the 10 years of DE-Gri data, using the fortnightly data
* Daily LAI predictions for the interpolated DE-Gri fortnightly GPP predictions.

The `pyrealm_build_data.phenology` directory is a bit of a muddle now because all of the code is written a single file that is focussed only at Boya Zhou's method. It would be better to reorganise into an `inputs` directory and then `zhu_method` and  `zhou_method` to create outputs but that can wait.

Fixes #590

## Type of change

## Key checklist

## Further checks


pyrealm_build_data/phenology/zhu_method/fapar_limitation.py
```diff
@@ -0,0 +1,118 @@
...
+
+        one_plus_r = 1 + r_local
+        r_clamped = max(r_local, -0.999)
+        r_pow = r_clamped**const_budyko
+        inside = 1 + r_pow
+        root_term = inside ** (1 / const_budyko)
+        fcomb = one_plus_r - root_term
+
+        flat_factor[ii] = fcomb
+        flat_out[ii] = fcomb * fw
```
MarionBWeinzierl: No idea what this is for - maybe add some more comments to explain?

pyrealm_build_data/phenology/zhu_method/plmodel_timeseries.py
```diff
@@ -0,0 +1,186 @@
...
+    A0_data_prc_max = _nanpercentile(A0_data, const_cli_prc)
+    A0_data_prc_max = np.where(
+        A0_data_prc_max < const_min_a0, const_min_a0, A0_data_prc_max
+    )
+
+    with np.errstate(divide="ignore", invalid="ignore"):
+        mratio_cal = lai_max_grid / A0_data_prc_max
+    mratio_cal[~np.isfinite(mratio_cal)] = np.nan
+
+    SLAI15DE1 = A0_data * mratio_cal[:, :, np.newaxis]
```
MarionBWeinzierl, line 142: We do not normally use all-caps in variable names in PyRealm - is this some sort of special constant? Same below

MarionBWeinzierl (approved):
Just some small comments on variable naming and lack of comments. Otherwise this looks good.

pyrealm_build_data/phenology/zhu_method/plmodel_timeseries.py
davidorme: This is Ziqi's code. I suspect it can be refactored to be clearer and more efficient but I'm leaving the original in `pyrealm_build_data` as supplied.

pyrealm_build_data/phenology/zhu_method/fapar_limitation.py
davidorme: I think this is Ziqi's code - might be Mateusz's. I think it's basically a straight port from the original MatLab and it can be made much more efficient (there's a two liner that does the same job `cal_fapar_actually_in_python`). So yes, badly commented but its the reference implementation. I will comment on the production version 😄

davidorme:
@MarionBWeinzierl - I should have pointed out the reference code which isn't up to spec. I don't intend to modify the reference implementation files but I've restructured a bit to bring all the reference code together and clearly noted that it is not up to spec.

PR #598, opened 2026-01-21
Updates to test_broadcasting

### LINKED ISSUE #537: Review skipped methods in the broadcasting test
The broadcasting test contains a number of functions in the `SKIP_METHODS` list which are not currently tested, either because the test is irrelevant for that function, it cannot be called (like methods of abstract base classes), or it requires more work to make it run. The list should be reviewed to see which are in the latter category and see if they can be made to work.

### LINKED ISSUE #597: Remove repeated construction of pyrealm namespace in test_broadcasting.
The `generate_args` function constructs the namespace of the package every time it is called.

davidorme:
# Description

This PR started off as a few efficiency savings in broadcast testing but has grown to include an extension to handle required keywords.

**Efficiency savings**

* Removes code from `generate_args` that compiles a global namespace for the `pyrealm` package. This should be static across all the subtests but is called twice for each test.
* Moved the code into a standalone function that creates the global variable `GLOBALNS` within the `broadcasting.utils` module.
* Restructures `test_array_input_broadcasting` so that `generate_args` isn't required from `__init__` methods. The `__init__` methods get their args directly from the `initialise_class` function , which is designed to run `generate_args` for `__init__` and then run `__init__`. At the moment,  `Class.__init__` methods gets run twice.
* Adds an explicit filter into `get_method_list` that excludes any methods of an abstract class from testing without needing to manually exclude them
* The PR also standardises the names of the `shape` attribute of classes (`["shape", "_shape_", "data_shape"]` --> `"shape"`) and then automatically excludes this attribute from `compare_instances`.

**Required args**

The PR adds a `REQUIRES` global variable that provides a dictionary of additional parameters required as keyword args. The `generate_args` code now looks up the method name and parents in this dictionary and adds any required parameters to the generated arguments.

Fixes #597
Fixes #537 - I think everything excluded from testing should be there.

## Type of change

## Key checklist

## Further checks


tests/broadcasting/utils.py
```diff
@@ -674,11 +753,17 @@ def value_string(val: Any) -> str:
 
 
 def compare_instances(instance1: Any, instance2: Any):
-    """Raises ValueError if the two class instances do not have equal attributes."""
+    """Raises ValueError if the two class instances do not have equal attributes.
+
+    This function ignores the shape attribute of any class, which is not expected to
+    broadcast, and anything in the manually defined list IGNORE_OUTPUTS.
```
sjavis, line 760: This is a sensible way to do it!

tests/broadcasting/utils.py
```diff
@@ -373,6 +441,10 @@ def get_method_list() -> list[tuple[str, Callable, type | None]]:
     method_list = []
     for mod in _get_package_modules(pyrealm):
         for name, method, cls in _get_module_callables(mod):
+            # Don't add methods of abstract classes
+            if cls is not None and isabstract(cls):
+                continue
+
```
sjavis, line 448: This would also skip static methods of abstract classes. Although, I'm not sure if there are any of these are in package.
```suggestion
            # Don't add methods of abstract classes
            if cls is not None and isabstract(cls):
                attr = getattr_static(cls, mname, None)
                if not isinstance(attr, staticmethod):
                    continue

```

sjavis (approved):
This looks like it adds a lot of sensible improvements to the broadcasting test and should tidy it up a little. I have added a suggestion for allowing static methods for abstract classes, but otherwise it looks good to me.

PR #599, opened 2026-01-23
Switch water viscosity and density methods

### LINKED ISSUE #583: Switch default viscosity and density methods
Following on from merging #573, we want to switch the default methods for viscosity and density to the faster approximations picked out in the new `water.md` documentation:

```python
water_density_method="jones_harris_eq6"
water_viscosity_method="vogel"
```

* That's going to throw the tests and the expected values in tests and doctests will need updating very widely, unfortunately. 
* Alternatively, we _could_ retain the old settings in testing by creating a `CoreConsts` fixture that sets the old defaults but that then needs to be passed in to a very large number of tests. It just seems better to update the test values except...
* In specific regression test cases, we need to retain the old settings. The `rpmodel` regression tests are the obvious ones, but maybe `SPLASH` needs a specific implementation. (ETA: actually, no we're good on the SPLASH regression tests. They  already explicitly use a fixture that configures `chen` for water density).

davidorme:
# Description

This PR looks horrible (33 files!) but it's mostly just tweaking settings and rerunning golden datasets:

* Switches the default water density method from `fisher` to `jones_harris_eq6` and the default water viscosity method from `huber` to `vogel`.
* Adds a section to the version `1.0` to `2.0` migration page about this change
* Updates the tests to pass, either by updating the test values or changing the test to use the original water methods. That's a little lazy, but recalculating some of the test values would be hard, and changing to the default water methods isn't relevant to these tests. For one test, I've run it with old and new settings.
* Updated the phenology regression datasets to use the new settings.

Fixes #583

## Type of change

## Key checklist

## Further checks


docs/source/users/versions.md
```diff
@@ -28,9 +28,9 @@ package is being extended and revised. This document is describes the main chang
 functionality between major versions and provides code migration notes for changes in
 the API of `pyrealm` code.
 
-## Migration from 1.0.0 to version 2.0.0
+## Migration from 1.0 to version 2.0
```
MarionBWeinzierl, line 31: Is there a reason that we dropped a `.0`?

MarionBWeinzierl (approved):
Trusting that the changed values are all ok, as you explained

docs/source/users/versions.md
davidorme: It's kinda obfuscating the `2.0.0` versus `2.0.1`. Which is cheating a little bit, but I'd rather people didn't get hung up on installing `2.0.0` specifically.

Issue #600, opened 2026-01-23
[FEATURE] Expand `phenology` module to multiple methods


davidorme:
With the arrival of the Zhu method, we need to support parallel implementations for calculating annual maximum fAPAR and the Phenology time series.

PR #605, opened 2026-01-27
Add regression test of `FaparLimitationNew.from_pmodel`

### LINKED ISSUE #604: Add regression test of FaparLimitationNew.from_pmodel
The regression test suite on FaparLimitationNew does not include `FaparLimitationNew.from_pmodel` and it should.

davidorme:
# Description

I had left out a regression test of the `FaparLimitationNew.from_pmodel` method in the new implementation. This PR:

* Copies the existing test of the method on the old implementation out of `tests/regression/phenology/test_phenology_faparlimitation.py` and then updates it to run both methods available in the new implementation.
* Removes some fixture duplication and aligns the fixtures used by the new and old versions, so that both test suites run off the same fixtures and parameterisation.
* Fixes a bug where the additional variables in `**kwargs` were not being passed down through `from_pmodel` into the class constructor.

This is being merged down into the feature branch for this new functionality (#600)

Fixes #604 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/regression/phenology/test_fapar_limitation_new.py
+++ b/tests/regression/phenology/test_fapar_limitation_new.py
@@ -1,63 +1,15 @@
 """Test the FaparLimitation class."""
 
-import json
-from importlib import resources
-
-import pandas as pd
+import numpy as np
 import pytest
 from numpy.testing import assert_allclose
 
 
-def dataframe_to_dict_of_nparrays(data):
-    """Utility function to preconvert a dataframe of pd.Series to np.array."""
-
-    data_dict = {k: v.to_numpy() for k, v in data.items()}
-
-    return data_dict
-
-
-@pytest.fixture
-def site_data():
-    """Load the site data."""
-
-    datafile = (
-        resources.files("pyrealm_build_data.phenology.inputs.source")
-        / "DE-GRI_site_data.json"
-    )
-
-    with open(str(datafile)) as json_src:
-        site_data = json.load(json_src)
-
-    return site_data
-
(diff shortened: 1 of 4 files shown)


tests/regression/phenology/test_fapar_limitation_new.py
```diff
@@ -69,36 +21,144 @@ def data_fapar_limitation(timescale, method_predictions_dir):
...
+    argnames="method_predictions_dir, method",
+    argvalues=(
+        pytest.param("cai_zhou_method", "cai", id="cai"),
+        pytest.param("zhu_method", "zhu", id="zhu"),
+    ),
+)
+@pytest.mark.parametrize(
+    argnames="timescale",
+    argvalues=(
+        pytest.param("hh", id="subdaily"),
```
MarionBWeinzierl, line 69: This keeps bugging me a bit - "hh" standing for "half hourly" but called "subdaily". Maybe we should already be calling things "subdaily" or "sd" to account for, say, hourly data?

Not a big thing, of course, just keeps nagging me everytime I see it.

MarionBWeinzierl (approved):
lgtm, just one comment which doesn't need addressing here

tests/regression/phenology/test_fapar_limitation_new.py
davidorme: Yeah - I'm with you and I think you're right. It's a hangover from the original implementation of the subdaily model that I haven't cleaned out.

PR #606, opened 2026-01-27
Add multi-method Phenology

### LINKED ISSUE #601: Update Phenology methods to support Zhu
With the reference code for the Zhu methods, we can now build out `Phenology` to support alternative approaches. Zhu requires a separate parameter (AET/PET) and uses a different calculation to produce the time series.

As with FaparLimitation (#592) probably the best way to do this is:

* Update Phenology to handle optional variables like `PModelEnvironment` currently does.
* Follow the practice elsewhere in `pyrealm` of breaking out method specific differences into subclasses of a new `PhenologyMethodABC`. The class should:
    * Check any required variables are provided in `__init__`
    * Provide a method that calculates the time series.

davidorme:
# Description

This PR is to implement an initial version of `PhenologyNew`:

* Adds `PhenologyMethodABC` which is an abstract base class for specific implementations of daily phenology calculation. 
* Adds `PhenologyMethodZhou` and `PhenologyMethodZhu` (may need more distinct names!)
* Adds `PhenologyNew` which wraps the two methods and provides shared additional functionality, including a `from_pmodel` method.
* Updates the phenology regression testing values and adds tests for the phenology methods.

There is an issue in testing the new Zhu method - the reference implementation is written explicitly with annual blocks and to spin up each LAI time series by tiling the year. This doesn't make sense for longer time series but at least in the regression test we need to match - that's fine for a direct test of `PhenologyNew(..., method='zhu',...)` where we control the daily values but it doesn't work well for `PhenologyNew.from_pmodel(..., method='zhu',...)` because the interpolation and aggregation between the full time series and annual blocks is different and the fortnightly data have to be extrapolated out to full years. It would be _incredibly_ finicky for a fringe case that we only want to support for regression tests.

It's also failing some broadcasting tests.

My thoughts:
* This code works against our regression tests, so I think it is a good starting point for the expanded set of methods.

* I'm not _convinced_ by the structure but I think that `PhenologyNew(..., method='xyz', ...)` is an API that allows us to expand functionality later. We _could_ just have `PhenologyZhu` and `PhenologyZhou` and lose the wrapping class.

* ~The `gpp_penalty_factor` propagation is a pain in the arse again - need to feed those values all the way back down to `PModelABC._daily_gpp` implementations.~ (Propagation of penalty factors has been removed - see #616)

Fixes #601 

## Type of change

## Key checklist

## Further checks


pyrealm/phenology/phenology_new.py
```diff
@@ -0,0 +1,447 @@
...
+    @classmethod
+    def __init_subclass__(cls, method: str) -> None:
+        """Initialise a subclass deriving from this ABC."""
+
+        cls.method = method
+        PHENOLOGY_METHOD_CLASS_REGISTRY[cls.method] = cls
+
+
+class PhenologyMethodZhou(PhenologyMethodABC, method="zhou"):
+    """Blah blah blah."""
```
MarionBWeinzierl: Probably needs to be a bit more specific... ;p

pyrealm/phenology/phenology_new.py
```diff
@@ -0,0 +1,447 @@
...
+        )
+        """The realised leaf area index for each day given the modelled lag in responses
+            to changing assimilation."""
+
+        self.steady_state_lai = steady_state_lai
+        self.realised_lai = realised_lai
+
+
+class PhenologyMethodZhu(PhenologyMethodABC, method="zhu"):
+    """Blah blah blah."""
```
MarionBWeinzierl: and here, too

MarionBWeinzierl (commented):
Two spurious non-comments:

pyrealm/phenology/phenology_new.py
davidorme: Ha - yes. I'm mostly after structural feedback but it's harder to review units that don't have a description!

MarionBWeinzierl:
Also fixes #612 ?

davidorme:
> Also fixes #612 ?

Nope - this currently perpetuates the problem that #612 wants to cure. I want to get rid of all of the `gpp_penalty_factor` arguments that we currently have to shoehorn in.

davidorme:
> > Also fixes #612 ?
> 
> Nope - this currently perpetuates the problem that #612 wants to cure. I want to get rid of all of the `gpp_penalty_factor` arguments that we currently have to shoehorn in.

Oh wait - I think I see what you mean. No - this adds fixes for `_get_daily_gpp` which _do_ look like the proposal in #612 but  we currently e.g. have to pass `gpp_penalty_factor` to `FaparLimitationNew` _and_ then to `PhenologyNew` and both of those then have to pass that down to e.g. `_get_daily_gpp`. The point in #612 is to get rid of all of that. The user penalises the model and then just forgets about it - everything uses the penalised value automatically.

pyrealm/pmodel/pmodel.py
MarionBWeinzierl, line 104: This is kind of different but kind of related to the gpp penalty factor, should probably make sure to be consistent.

MarionBWeinzierl:
Hm, if all upstream classes (faparlim, phenology,...) had a `gpp_penalty_factor` attribute which is set to `1.0` per default, could `pmodel`'s `penalise_gpp` function simply do something like `gpp *= faparlim.gpp_penalty_factor *= phenology.gpp_penalty_factor *= self.gpp_penalty_factor` to collect all penalty factors without needing to propagate them as function arguments?

davidorme:
@MarionBWeinzierl 

OK - I have updated this to remove the `gpp_penalty_factor` arguments, which can now be passed in as part of a P Model (#616). I've also updated the docstrings in both the new Phenology classes and the expanded constants class.

It is still failing on broadcasting but - if the big picture structure and docs seem sound - I think I'd rather merge down into the feature branch and then handle broadcasting issues before we pull into `develop` particularly if @sjavis is looking at the structure of the broadcasting testing.

pyrealm/constants/phenology_const.py
```diff
@@ -67,35 +67,75 @@ def calculate_f0(self, aridity_index: NDArray[np.floating]) -> NDArray[np.floati
...
-    support the nutrient demand of those leaves. The default is an empirical estimate
-    from global data."""
+    * The value :math:`z` (mol C m^{-2} year^{-1}) accounts for the carbon costs of
+      building and maintaining leaves and the total below-ground allocation required to
+      support the nutrient demand of those leaves. It is used to calculate an
+      energy-limited estimate of maximum fAPAR.
+
+    * The value :math:`f_0` (-) accounts for water limitation on annual assimilation and
+      hence sets the water-limited estimate of maximum fAPAR.
+    """
```
MarionBWeinzierl, line 88: I am not sure I understand how this description is related to the variables below. Is it so that there is a `z` which is then defined as `cai_z` or `zhu_z`. And the `f0` is somehow somewhere computed from the `cai_f0_coefficients`, whereas for `zhu` it's simple a half? And what is all the rest?

pyrealm_build_data/phenology/zhu_method/plmodel_timeseries.py
```diff
@@ -237,8 +244,33 @@ def _apply_lag(data_cube: np.ndarray, lag_steps_grid: np.ndarray) -> np.ndarray:
...
+    #         "counts": counts.ravel(),
+    #         "csum_values": csum_values.ravel(),
+    #         "csum_counts": csum_counts.ravel(),
+    #         "do_totals_track": do_totals_track.ravel(),
+    #         "do_counts_track": do_counts_track.ravel(),
+    #         "do_idx_start_track": do_idx_start_track.ravel(),
+    #         "result": result.ravel(),
+    #     }
+    # )
+    # df.to_csv("zhu_tracking.csv")
```
MarionBWeinzierl: We might want to delete this?

MarionBWeinzierl (approved):
OK, if you are happy to merge this with failing tests that's fine.

sjavis:
I'm happy to look at the failing tests and resolve them in my work on the broadcasting tests. I would add them to `SKIP_METHODS` to let the tests pass for now.

pyrealm/constants/phenology_const.py
davidorme: Yeah - it is confusing. I've added text to clarify that the other parameters are method specific and that the different method use different default values of $f_0$ and $z$ and that the Cai approach even modulates $f_0$ as a function of aridity.

pyrealm_build_data/phenology/zhu_method/plmodel_timeseries.py
davidorme: I think it is just about worth hanging on to as part of the workflow of the implementation of this method in pyrealm and in figuring out how to vectorise the code in this file. I've explained it in the module docstring and improved the inline description.

davidorme:
@MarionBWeinzierl I've (hopefully) improved the explanations of the bits you queried.
@sjavis Thanks - I've added the methods to `SKIP_METHODS` for now

PR #609, opened 2026-01-28
Fix and improve running performance regression checking shell script

AdelekeBankole:
#Fix and improve running shell script

Since this is a shell script, it is advisable to set execute permission on the script with the `chmod +x` and then running with `./performance_regression_checking.sh`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -112,7 +112,8 @@ with a message indicating whether the new code is faster or has similar performa
 
 ```sh
 cd profiling
-poetry run performance_regression_checking.sh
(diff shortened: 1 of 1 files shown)


MarionBWeinzierl (approved):
This is correct, thanks for adding it to the instructions.

davidorme (approved):
@AdelekeBankole fine by me too. 

Just also to say, the usual practice on this repo is we don't wait for all reviews to come in. We often ask a couple of people and the first 👍 is enough. Unless we _specifically_ want multiple opinions 😄 . Don't know if you were waiting or not!

PR #610, opened 2026-01-29
Incorrect calculation of stomatal conductance in subdaily p model

### LINKED ISSUE #603: Incorrect calculation of stomatal conductance in Subdaily P Model
@dsval reports:

> I was running the subdaily Pmodel version from pyrealm, and I notice in the outputs that the stomata are not closing at night, which is weird because the calculation uses gpp, which is 0 at night in the same outputs. Is this a bug? or am I doing something wrong? Just in case, I am attaching my code.

A couple of things from this:

1. The `PModel` docstrings say that it is calculated from light use efficiency, but of course it isn’t because it needs to be scaled by actual assimilation, right? Otherwise it doesn't change with light.  And in fact, in the code, it _is_ calculated from assimilation. 

2. In the standard PModel code it is calculated from `A_c`, which is fine, because `A_c` = `A_j` in that model. But…. the subdaily model duplicates that code and calculates `gs` from `A_c`, but here `A_c != A_j`. And that is a problem because `A_c` does not scale with light (`I_abs`) and `A_j` does:

https://github.com/ImperialCollegeLondon/pyrealm/blob/d8aa1a39bdea3d2b63498e15ca111cdc99431a26/pyrealm/pmodel/pmodel.py#L964-L988

So the subdaily model should be using `assimilation = np.minimum(self.A_j, self.A_c) ` for calculating assimilation and it doesn’t. And that, I think, will give the behaviour  @dsval describes. Basically, Giulia’s original code didn’t include `gs` and I’ve added it without noticing this difference.

davidorme:
# Description

This PR:

* corrects a bug in calculating stomatal conductance in the Subdaily P Model
* Updates the docstrings of the PModel and SubdailyPModel to correct the explanation of the calculation and the difference between the models.

Fixes #603

## Type of change

## Key checklist

## Further checks


AdelekeBankole (approved):
Thanks @davidorme , this looks good to me.

PR #611, opened 2026-01-30
Add Ade to authors list

AdelekeBankole:
# Description

From meeting with @MarionBWeinzierl and @sjavis, I am updating the authors list.


pyproject.toml
```diff
@@ -7,7 +7,7 @@ version = "2.0.0"
 
 dependencies = [
   "dacite (>1.6.0)",
-  "numpy (>=2.1)",
+  "numpy (==2.3.2)",
   "scipy (>=1.7.3)",
```
davidorme, line 11: Can we make this a bit broader? `(>=2.1,<2.4)` It might only be temporary, but I don't think we should pin to a minor version

davidorme (changes requested):
No problem with the name addition! - just the numpy pin

MarionBWeinzierl:
@AdelekeBankole , the test are currently failing, please check this: 
image

davidorme (approved):

PR #616, opened 2026-02-09
Allow a gpp penalisation to be attached to a PModel instance

### LINKED ISSUE #612: Allow the GPP predictions from PModel instances to be penalised
We have an increasing amount of functionality that takes a PModel as an input and then uses the GPP from that model for various purposes (e.g. phenology, two leaf model). The problem is that sometimes we want to apply a posthoc penalty  to GPP (e.g. Mengoli or Stocker water stress penalties) and that penalised GPP should be used by downstream methods. 

At the moment, we have had to add optional `gpp_penalty_factor` arguments to methods to allow those posthoc penalties to be included. That:

* makes code and method signatures more complex, and
* makes user code (marginally) more complex - they have to keep passing in the penalties.

Instead, we could:

* add a new class method  `PModelABC.apply_gpp_penalty(factor: NDArray[np.floating]): -> None`  that validates `factor` ($f \in (0,1)$, right shape) and uses it to populate a new instance attribute, defined as `PModelABC.gpp_penalty_factor: NDArray[np.floating] | None = None`
* move the `gpp` attribute to `_gpp`.
* Adds a `gpp` property:

```python
@property
def gpp(self) -> NDArray[np.floating]:

    if self.gpp_penalty_factor is None:
        return self._gpp

    return self._gpp * self.gpp_penalty_factor
```
* Adds a `remove_gpp_penalty` to simply reset `PModelABC.gpp_penalty_factor = None`
* Updates the `__repr__` to add a 'penalised' string in the model description and maybe updates `summarize` to include simple stats on the penalty. Just basically to make it obvious in outputs that a penalty is in place.

We can then go through the codebase and remove `gpp_penalty_factor` args.

davidorme:
# Description

This PR updates PModelABC to provide the `apply_gpp_penalty_factor` method. This is primarily aimed at allowing penalised GPP predictions to be used more easily in downstream methods that take a PModel as an input, rather than having to propagate penalty factors through the signatures of those downstream methods.

I've added simple tests of the methods and updated the current `fapar_limitation` code to remove the `gpp_penalty_factor` arguments.

Once this is in `develop`, I'm going to pull it up into #600 and adopt it through the incoming new phenology code in that feature branch.

Fixes #612

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/phenology/fapar_limitation.py
+++ b/pyrealm/phenology/fapar_limitation.py
@@ -246,7 +246,6 @@ class FaparLimitation:
         precip: NDArray[np.floating],
         aridity_index: NDArray[np.floating],
         datetimes: NDArray[np.datetime64] | None = None,
-        gpp_penalty_factor: NDArray[np.floating] | None = None,
         phenology_const: PhenologyConst = PhenologyConst(),
     ) -> FaparLimitation:
         r"""Create a FaparLimitation instance from a P Model and other inputs.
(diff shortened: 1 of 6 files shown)


MarionBWeinzierl (approved):
This looks good to me, including the additional test

PR #629, opened 2026-02-26
Adding fine root masses into demography allometry

### LINKED ISSUE #628: Include fine root mass in demography calculations
The fine root mass under the T model is used in a couple of places (respiration and turnover) but is always calculated implicitly. So, the mass is never included in the allometry outputs and so the total carbon mass of stems is not completely recorded.

davidorme:
# Description

This PR:

* Pulls the calculation of fine root mass into its own function and adds that as an attribute to `StemAllometry`
* Updates the two functions that use fine root mass (currently in a round about way) to directly use the fine root mass and updates their docstrings to explain how the new calculation is equivalent to the formulae in Li et al. 2014.
* Updates the tests to pass - I haven't added in a separate test of the new function because all the testing currently uses the reference implementation values, which don't include fine root masses. However the function is checked as part of the chain of calculating the derived values from fine root mass.

Fixes #628

## Type of change

## Key checklist

## Further checks


pyrealm/demography/tmodel.py
```diff
@@ -270,6 +270,40 @@ def calculate_foliage_masses(
...
+    the product of the crown area (:math:`A_{c}`) and leaf area index (:math:`L`) - and
+    the ratio of fine root mass to leaf area (:math:`zeta`).
+
+    .. math::
+
+        W_r = A_c L \zeta
+
+    Args:
+        crown_area: Crown area of individuals
+        lai: Leaf area index of the PFT
```
```suggestion
        lai: Leaf area index of the PFT
        crown_area: Crown area of individuals
```

sjavis: To match the order of the arguments

pyrealm/demography/tmodel.py
```diff
@@ -562,40 +596,44 @@ def calculate_reproductive_tissue_respiration(
...
     validate: bool = True,
 ) -> NDArray[np.floating]:
     r"""Calculate fine root respiration.
 
-    Calculates the total fine root respiration (:math:`R_{r}`) given the individual
-    foliage mass (:math:`W_f`), along with the fine root respiration rate (:math:`r_r`),
-    the ratio of fine root mass to foliage area (:math:`\zeta`) and the specific leaf
-    area (:math:`\sigma`) :cite:p:`{see Equation 13, }Li:2014bc`
+    Calculates the total fine root respiration (:math:`R_{r}`) given fine root mass
+    (:math:`W_f`) the fine root respiration rate (:math:`r_r`):
```
sjavis: Should the fine root mass be `W_r`?

pyrealm/demography/tmodel.py
```diff
@@ -703,41 +741,45 @@ def calculate_foliage_turnover(
...
 ) -> NDArray[np.floating]:
     r"""Calculate turnover costs.
 
     This function calculates the costs associated with the turnover of fine roots. This
-    is calculated from the total foliage mass of individuals (:math:`W_f`), along with
-    the specific leaf area (:math:`\sigma`) and fine root mass to foliar area ratio
-    (:math:`\zeta`) and the turnover time of fine roots (:math:`\tau_r`) of the plant
-    functional type :cite:p:`{see Equation 15, }Li:2014bc`.
+    is calculated from the total fine root mass of individuals (:math:`W_r`) and the
+    turnover time of fine roots (:math:`\tau_r`) of the plant functional type.
```
sjavis, line 752: The rest of the description only seems to refer `\tau_f`. Should these be `\tau_r` instead?

pyrealm/demography/tmodel.py
```diff
@@ -703,41 +741,45 @@ def calculate_foliage_turnover(
...
 
-        T = W_f \left(\frac{ \sigma \zeta}{\tau_f} \right)
+        T = \frac{ W_r}{\tau_f}
+
+    Equation 15 of :cite:`Li:2014bc` gives this as:
+
+        T = W_f \left(\frac{ \sigma \zeta}{\tau_f} \right),
+
+    given the foliage mass of individuals (:math:`W_f`), the specific leaf area
+    (:math:`\sigma`) and fine root mass to foliar area ratio (:math:`\zeta`), which can
```
```suggestion
    (:math:`\sigma`) and fine root mass to foliage area ratio (:math:`\zeta`), which can
```

sjavis (approved):
This looks good. I have a few issues with the docstrings but other than that it is good to merge.

pyrealm/demography/tmodel.py
davidorme: They should

PR #635, opened 2026-03-03
SPEC0 status of package versions

### LINKED ISSUE #586: SPEC0 review
### Review the SPEC0 status of packages

Check the quarterly [SPEC0 recommendations for dropping version
support](https://scientific-python.org/specs/spec-0000/) and update
`pyproject.toml`.

AdelekeBankole:
# Description

This PR reviews the SPEC0 status of packages, following [SPEC0 recommendations for dropping version
support](https://scientific-python.org/specs/spec-0000/).

Fixes #586 

## Type of change

## Key checklist


pyproject.toml
MarionBWeinzierl:
```suggestion
requires-python = ">=3.12"
```

AdelekeBankole:
@davidorme, I have dropped the package versions we don't need following spec0 support window.

MarionBWeinzierl (dismissed):
Scratch that comment - I have seen that you already did that.

AdelekeBankole:
@davidorme, seems the tests in the ci is still confused about the dropped `3.11` version. I can't even click on these pending checks.

davidorme:
> @davidorme, seems the tests in the ci is still confused about the dropped `3.11` version. I can't even click on these pending checks.

Yup - we'd need to edit the required checks on merging the branch.

davidorme:
It would also be good to update the pre-commit config so that `ruff` detects syntax changes from 3.11 to 3.12. That should just using changes like these:

https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1399/changes#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R1-R17

And then fixing anything that ruff then complains about. Probably something to do with typing 😄

AdelekeBankole:
Hello @davidorme, I fixed the complains that came from ruff and typing as you said. Me and @sjavis will look into other failing check that came up from the `build/docs_build`, presumably coming from sphinx.

AdelekeBankole:
Thanks @sjavis, that went through cleanly.

pyrealm/core/xarray.py
```diff
@@ -1,15 +1,15 @@
...
 
 import numpy as np
 import xarray as xr
 from numpy.typing import NDArray
 
 T = TypeVar("T", bound=np.generic)
 
-type ArrayType[T: np.generic] = NDArray[T] | xr.DataArray
+ArrayType: TypeAlias = NDArray[T] | xr.DataArray  # noqa: UP040
 """Type for array inputs. A union of numpy arrays and xarray DataArrays."""
```
davidorme, line 15: @sjavis and @AdelekeBankole 

Could you add a comment here explaining why we're not using the new Python 3.12 `type` syntax? I know there are edge cases where the new syntax doesn't apply - it would be good just to record why this exception is in place.

sjavis: I added a comment in 15695c4a.

In more detail:
Sphinx doesn't correctly create the cross references to the type aliases. It creates the alias as a py:data object but tries to use py:class for the cross references (this may be fixed in sphinx 9.0 but this is incompatible with the other dependencies currently). I made a workaround for this in e8df7ee, but this doesn't seem to work for the new syntax.

davidorme (approved):
LGTM, thanks @AdelekeBankole, I'm going to update the tests now.

davidorme:
~I'm just trying to switch from branch based protection to the newer rulesets - currently I don't think the ruleset is being applied.~ OK - the new set of rules is now being applied by a ruleset.

davidorme:
Just applied a docs fix: the default intersphinx inventory for `pandas` is now using `pandas 3.0` and for some reason that breaks links to the `DataFrame` class.

AdelekeBankole:
Thank you @davidorme.

PR #641, opened 2026-03-17
Partition allocation to foliage growth

### LINKED ISSUE #628: Include fine root mass in demography calculations

davidorme:
The description in issue #628 isn't quite correct. The `StemAllocation` object has an attribute `delta_foliage_mass` that is actually the combined increase in foliage ($f$) and fine root ($r$) masses (as the T Model calculates fine root mass as a function of canopy leaf area). 

This PR updates:
*  `calculate_growth_increment` to partition $\Delta W_{fr}$ into $\Delta W_{f}$  and $\Delta W_{r}$,
*  `StemAllocation` to store those two values in separate attributes,
*  the docstrings to explain the difference from the original implementation,
* the tests to specifically test that the sum of the two partitions still equal the single output from the reference implementation.

> [!WARNING]
> Oh damn. Made the commits to the wrong branch. Fixing

Fixes #628

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/demography/tmodel.py
+++ b/pyrealm/demography/tmodel.py
@@ -852,7 +852,12 @@ def calculate_growth_increments(
     dbh: NDArray[np.floating],
     stem_height: NDArray[np.floating],
     validate: bool = True,
-) -> tuple[NDArray[np.floating], NDArray[np.floating], NDArray[np.floating]]:
+) -> tuple[
+    NDArray[np.floating],
+    NDArray[np.floating],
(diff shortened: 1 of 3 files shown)


davidorme:
Closed - commits to wrong branch

PR #642, opened 2026-03-17
Add fine root growth mass to allocation

### LINKED ISSUE #640: Add fine root growth mass to allocation
Similar to #629, the T Model implementation has an implicit calculation of carbon allocation to fine root growth, but this is not recorded in the StemAllocation class. Hence the total allocation of carbon is not recorded.

ETA: Actually the allocation to fine root growth is included, but it is lumped in with allocation to foliage growth. This should be documented more clearly and the two components should be separated out.

davidorme:
The description in issue #640 isn't quite correct. The `StemAllocation` object has an attribute `delta_foliage_mass` that is actually the combined increase in foliage ($f$) and fine root ($r$) masses (as the T Model calculates fine root mass as a function of canopy leaf area). 

This PR updates:
*  `calculate_growth_increment` to partition $\Delta W_{fr}$ into $\Delta W_{f}$  and $\Delta W_{r}$
*  `StemAllocation` to store those two values in separate attributes,
*  the docstrings to explain the difference from the original implementation,
* the tests to specifically test that the sum of the two partitions still equal the single output from the reference implementation.

Fixes #640

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyrealm/demography/tmodel.py
+++ b/pyrealm/demography/tmodel.py
@@ -925,7 +925,7 @@ def calculate_growth_increments(
         \end{align*}
       \]
 
-    Note that :cite:`Li:2014bc` use :math:`W_f` to denote the increment in both foliage
+    Note that :cite:`Li:2014bc` use ':math:`W_f`' to denote the increment in both foliage
     and fine root mass, as fine root mass is estimated as a function of foliage area
     through the specific leaf area (:math:`\sigma`) and  ratio of fine root mass to leaf
     area (:math:`\zeta`). Here we use :math:`W_fr` to indicate the combined increments
--- a/pyrealm/demography/tmodel.py
+++ b/pyrealm/demography/tmodel.py
@@ -925,11 +925,12 @@ def calculate_growth_increments(
         \end{align*}
       \]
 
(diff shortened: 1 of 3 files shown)


pyrealm/demography/tmodel.py
```diff
@@ -905,16 +912,33 @@ def calculate_growth_increments(
...
 
       \[
         \begin{align*}
         \Delta W_s &=  \textrm{d}W_s / \textrm{d}t \, \Delta D\\
-        \Delta W_f &=  \textrm{d}W_f / \textrm{d}t \, \Delta D
+        \Delta W_fr &=  \textrm{d}W_fr / \textrm{d}t \, \Delta D
+        \end{align*}
+      \]
+
+    Note that :cite:`Li:2014bc` use :math:`W_f` to denote the increment in both foliage
```
```suggestion
    Note that :cite:`Li:2014bc` use ':math:`W_f`' to denote the increment in both foliage
```
This is a small thing but I think it makes it clear it's talking about the notation rather than the actual variable.

sjavis (approved):
I'm happy with these changes. It seems pretty clear and a relatively minor change. I just have a small suggestion in the docstring.

PR #643, opened 2026-03-19
Xarray inputs for other modules

### LINKED ISSUE #631: `xarray` inputs to remaining modules
Xarray array inputs have been enabled for the `pmodel` module in #630. They should also be enabled for the `splash`, `phenology`, and `core` modules (`demography` has it's own separate requirements for array inputs).

sjavis:
# Description

This branches off #644 so should be only be merged after that.

Adds xarray inputs to modules other than just pmodel.
- [x] Fixes some cross references in the docs (d504a0d).
- [x] Xarray inputs to core module (cbba03d).
- [x] Xarray inputs to splash module (2150c6d, 49c9f04).
- [x] Xarray inputs to phenology module ([8888338 - 05161e0](https://github.com/ImperialCollegeLondon/pyrealm/pull/643/changes/2150c6d418115001832d75b7cdc2dd4f2b816ba0..05161e0a9777dd10459c2d340f2364f18dd0ebeb)).
- [x] Update the SplashModel worked example to not use `to_numpy()` (1cec4f7)

In `SplashModel` and `DailySolarFluxes` the first dimension should be time, but under the xarray dimension rules the first dimension would be set by the first argument which is latitude. To avoid this issue I have:
- Add a `init_dims` argument to allow a list of initial dimensions to `get_common_dimensions` (which gives the dimensions for expanding xarray inputs) (07d38bc).
- Converted `Calendar` to also accept xarray inputs and store the dimension(s) (49c9f04). (I resolved the issue with mypy not recognising the narrowing of the dataclass attribute types in `__post_init__` by using an `__init__` function. I think this is also neater.)
- Using both of the above in `SplashModel` and `DailySolarFluxes` to set the initial dimension to be the time dimension when the `Calendar` argument is initialised using xarray.
- Added to the docstrings that the `Calendar` arguments should be initialised with xarray (or that latitude should include the time dimension).

~~There is also an issue with sphinx autodoc not identifying the correct cross-reference target for `PhenologyConst` in `phenology/fapar_limitation.py` as a result of duplicates of each of the constants classes in [the constants page of the docs](https://pyrealm.readthedocs.io/en/latest/api/constants_api.html). I'm not really sure why this just started turning up and only for `PhenologyConst`, but I have resolved it by using `constants.Phenology` in the file instead. I had tried removing the duplicates from the `constants_api.md` file (either from the `pyrealm.constants` module or the submodules) but this just caused sphinx to break in more places.~~

The sphinx issue is instead due to failing on `ArrayType[np.floating]` which then causes issues with the cross referencing of other arguments in the same functions. I added a patch in sphinx's `conf.py` to resolve this (11bef37)

Fixes #631

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -197,6 +197,7 @@ intersphinx_mapping = {
     "shapely": ("https://shapely.readthedocs.io/en/stable/", None),
     "marshmallow": ("https://marshmallow.readthedocs.io/en/stable/", None),
     "pooch": ("https://www.fatiando.org/pooch/latest/", None),
+    "matplotlib": ("https://matplotlib.org/stable/", None),
 }
 
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -224,6 +224,14 @@ autodoc_type_aliases = {
     "ArrayType": "pyrealm.core.xarray.ArrayType",
 }
 
+# TypeAliasForwardRef doesn't seem to support subscripting, resulting in
+# get_type_hints() failing on ArrayType[np.floating] in modules using `from __future__
+# import annotations`. This patch solves this by ignoring the type parameter, allowing
+# other arguments to resolve correctly.
+from sphinx.util.inspect import TypeAliasForwardRef  # noqa: E402
+
+TypeAliasForwardRef.__getitem__ = lambda self, item: self
+
 # Autodoc configuration:
 # - Suppress signature expansion of arguments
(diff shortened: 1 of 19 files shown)


davidorme (approved):
LGTM  - I think there's an argument for making `Calendar` an actual class rather than s crippled dataclass, but I'm not going to fuss about with it here 😄

PR #644, opened 2026-03-19
Reorganise array inputs test

sjavis:
# Description

This reorganises the array inputs tests to hopefully make it a little bit simpler to use. Since a lot of code has moved about it is easiest to understand by explaining each commit.
- [x] As a separate issue, fixes an issue with the advanced flag in the performance regression test (f6e27a0).
- [x] Move the manual definitions and overrides into their own module (`overrides.py`), also moves `Context` into it's own module to prevent a circular import (4dad11b). No actual changes to the code are made in this commit.
- [x] Change how the manual arguments are defined from using the `defined_method_args` function (this would initialise all manual arguments each time it was called and had no checks that arguments weren't defined twice) to using a registry-based `MANUAL_ARGS` which uses a separate function for the arguments of each callable (9cbf8f3). No changes to the manual arguments themselves are made in this commit.
- [x] A `--debug-tests` option is added to pytest that prints the generated arguments (08fde3bc). This automatically adds the `-s` and `--tb=short` arguments.
<details>
<summary>Example output:</summary>

```
DailySolarFluxes                                                                            
Parents: ['DailyEvapFluxes.estimate_aet', 'DailyEvapFluxes']                                
Arguments:                                                                                  
        latitude: (manual)                                                                  
                DataArray((4,), dims=('a',), dtype=int64)                                   
        elevation: (manual)                                                                 
                DataArray((4, 3), dims=('a', 'b'), dtype=int64)                             
        dates: (manual)                                                                     
                Calendar(1970-01-01, 1970-01-04)                                            
        sunshine_fraction: (manual)                                                         
                DataArray((3, 4), dims=('b', 'c'), dtype=float64)                           
        temperature: (manual)                                                               
                DataArray((3, 4, 4), dims=('b', 'a', 'c'), dtype=int64)                     
        core_const: (default)                                                               
                <factory>                                                                   
                                                                                            
DailyEvapFluxes                                                                             
Parents: ['DailyEvapFluxes.estimate_aet']                                                   
Arguments:                                                                                  
        solar: (automatic)                                                                  
                DailySolarFluxes                                                            
        pa: (manual)                                                                        
                DataArray((4,), dims=('a',), dtype=int64)                                   
        tc: (manual)                                                                        
                DataArray((4,), dims=('a',), dtype=int64)                                   
        kWm: (manual)                                                                       
                ndarray((3, 4), dtype=int64)                                                
        core_const: (default)                                                               
                <factory>           
```

</details>

[...]

PR #645, opened 2026-03-20
Add user docs for new phenology implementation

### LINKED ISSUE #623: Add user facing description of using the new methods
Add a user doc the demonstrates the new fAPAR limitation and phenology classes.

### LINKED ISSUE #624: Snagging of the new classes
There are some minor things that need fixing up:

- [ ] Remove `PModelABC.gpp_conversion_factor` and the related `PModelABC.apply_gpp_conversion_factor` - this conversion is now internal to `PhenologyNew`.
- [ ] Add code to prevent unsupported combinations of FaparLimitationNew and PhenologyNew methods.
- [ ] Probably more from writing the user demonstration docs!

davidorme:
# Description

This PR:

* Updates the FaparLimitation user docs to use the new implementation.
* Adds a new user doc for the new Phenology implementation.
* Fixes up some broken links and docs issues.

Fixes #623 
It probably also fixes #624 - the code in the docs seems fairly clean.

## Type of change

## Key checklist

## Further checks


davidorme:
@sjavis Could you have a look at the test failures? It's in the broadcasting - I think the issue is that the `PhenologyMethodXYZ` and `PhenologyNew` methods in `phenology_new.py` take a `method` argument that is being parameterised using the `automatic` approach and is being set to an empty string. I'd expect the default to be used? That doesn't work for the non default method testing though.

docs/source/users/phenology/fapar_limitation.md
```diff
@@ -76,62 +77,82 @@ Six of the terms above are annual estimates for locations of:
...
+        L_{max} = - ( 1 / k ) \ln {1 -f_{APAR_{max}}}
+$$
 
-The example below uses a time series of 11 years of annual values from the [`DE_Gri`
-Fluxnet site](https://fluxnet.org/doi/FLUXNET2015/DE-Gri). The code below loads some
-site constants and then a data frame of annual values for the required variables,
-including $A_0$ and $\chi$ from a P Model.
+## Example calculations
+
+The examples below uses a time series of half hourly data from the  [`DE_Gri`
```
MarionBWeinzierl, line 120:
```suggestion
The example below uses a time series of half hourly data from the  [`DE_Gri`
```

MarionBWeinzierl (approved):
Looks good

sjavis:
Yes, the checking whether to use default arguments in `_get_parameters` has a bug. Rather than:
```
        elif param.default is not param.empty and not (
            required_args and name not in required_args  # Don't use default if REQUIRES
        ):
```
It should be something like:
```
        elif (
            param.default is not param.empty
            and name not in (required_args or {})  # Don't use default if REQUIRES
        ):
```

There are also two other issues causing the tests to fail. The first is that in `REQUIRES` it uses `FaparLimitationCai` instead of `FaparLimitationMethodCai` and also needs the same for `FaparLimitationMethodZhu`:
```
    (
        "FaparLimitationNew",
        (
            "FaparLimitationMethodCai.calculate_maximum_fapar",
            "FaparLimitationMethodCai",
        ),
    ): _kwarg_params(("aridity_index",)),
    (
        "FaparLimitationNew",
        (
            "FaparLimitationMethodZhu.calculate_maximum_fapar",
            "FaparLimitationMethodZhu",
        ),
    ): _kwarg_params(("aridity_index",)),
```

Secondly, in `defined_method_args` for `PModelEnvironment` for `SubdailyPModel` the time dimension logic seems to have been lost when merging from develop. I would just add `SubdailyPModel` to `SKIP_METHODS` for now since it will be resolved when #644 is merged.

sjavis:
I've added some commits that fix some of the broadcasting test issues, but I have made it skip `PhenologyMethodZhu`, `PhenologyNew`, and `PhenologyNew.from_pmodel`. Because I wasn't sure what the shapes of the inputs should be / the broadcasting didn't seem to work. It should just now be the unit tests which are causing it to fail.

A couple of other changes I made:
- I think there is an issue with the interpolation for `Pmodel._get_daily_gpp`. I think this should be interpolating over axis 0 rather than -1 (78dcf22)
- I think I have corrected some of `PhenologyMethodZhu` (089e427) but I wasn't certain about what the shapes should be (Should `aet_pet_ratio` include the time dimension?) so this may not be correct.

One other thing that I noticed is that `AnnualValueCalculator` will fail somewhat unexpectedly with `The timings values are not equally spaced` if only one time is provided. It may be best to add a case for this.

PR #665, opened 2026-06-13
Demography 2: Stem Allometry and Allocation

### LINKED ISSUE #663: Rewrite T Model code
At the moment, TModel outputs always provide 2D predictions, allowing for height profiles for a cohort. However, this is built up on the idea of allowing the functions to accept PFT/Flora inputs (which have no DBH). The `at_dbh` argument is then needed to set the heights and that could be 1D (specific height) or 2D (range of heights).

In a new model, where we just have `Cohorts` as the starting point, then we can set up explicit and clearer handling of dimensions. The default is just to use the DBH associated with cohorts - this is basically a simple cohorts dataframe to allometry and allocation dataframes with row matching by index (or probably more safely by cohort id). 

If a size profile is needed, then an explicit `at_dbh` argument can be used to override the cohort DBH, providing a DBH series that generates provide 2D array predictions. This could possibly go to 3D predictions if someone wanted to combine a DBH series with a  GPP series in Allocation.

davidorme:
# Description

This PR refactors the `demography.tmodel` module for use in the new `demography_two` structure:

### `demography_two.cohorts`

* Adds a new argument to `Cohorts` for attaching a cohort id generator which is used to assign unique IDs to cohorts.
* Provides a default generator function to provide some alternative cohort ID formats (UUID, integer, etc).

### `demography_two.core`

* Module ported from `demography`
* The old `PandasExporter` ABC has been replaced with a simpler mixin `ToDataFrameMixin`, that just dumps specified attributes of a class into a dataframe.
* The original `demography.core` had a lot more stuff but I think they are all deprecated in `demography_two`

### `demography_two.tmodel`

* Module ported from `demography` and radically reworked. Not sure a diff of the two is helpful.
* Both StemAllometry and StemAllocation are rewritten to be Cohorts focussed. 
     * The standard usage now is simply that StemAllometry calculates the allometry at the DBH of the cohorts and StemAllocation calculates allocation based on a per cohort GPP prediction.
     * Both classes do still provide profiling modes using an array of DBH values or GPP, but these are secondary use cases mostly aimed at plotting curves to understand PFT allometry and growth.
     * The outputs in standard mode are 1D along the cohorts, basically giving a dataframe that aligns perfectly with the original Cohorts data. The profiling modes can lead to 2D or 3D outputs but the previous enforced 2D is dropped completely.

* Massive simplification of the heavy handed validation, primarily by making the individual T model calculations less constrained (and hence a little bit more dangerous) but handling validation at a higher level through the main `Cohorts` ->`StemAllometry`  -> `StemAllocation`  workflow. 
    * Enforcement of DBH  and GPP > 0 is now handled by classes
    * A lot of the dimensionality checks have become redundant and have been removed.
    * The outputs of T model functions now follow normal `numpy` broadcasting rules rather than enforcing 2D outputs.

### Testing

* Regression testing that `demography_two` validates against the golden dataset from `rtmodel`.
* Unit testing of `demography_two.tmodel`: largely a port from testing the original implementations - more checking against `rtmodel` and testing of dimensionality behaviour.

Fixes #663

## Type of change

## Key checklist

## Further checks


sallymatson (approved):
Everything makes sense to me. I didn't do a comprehensive review of the t-model equations but I'm trusting the tests etc :)

PR #671, opened 2026-06-18
Restructure of Cohorts

### LINKED ISSUE #670: Improve API on Cohorts
The new `demography_two` implementation of Cohorts:

* Uses a pydantic validation class to validate cohort inputs - nice, lots of wrapped functionality
* But... then `Cohorts` has to take that as an input, which is clunky and indirect.
* We _could_ try and overload `Cohorts` so that it will accept the raw DBH, PFT and names _or_ CohortData but that's fussy.
* Trying to make `CohortData` internal would be nice but then `from_csv` gets into loops.
* Numpy arrays don't work cleanly with pydantic (without adding extra machinery).

In this case, just having a single class with appropriate validation seems cleaner.

davidorme:
# Description

The current Cohorts structure is rather overcooked - there is a Pydantic validation class `CohortData` and then a `Cohorts` class:

```python
cohorts = Cohorts(
    flora=flora,
    cohorts=CohortData(
        pft_name=np.array(["narrow","medium","wide"]),
        n_individuals=np.array([1,1,1]), 
        dbh_value=stem_dbh
    )
)
```

This PR simplifies the structure radically - `Cohorts` is now just a rebadged `pd.DataFrame` and is created using the `create_cohorts` function either directly or wrapped in `create_cohorts_from_csv`.  The `create_cohorts` function takes on the validation previously in `CohortData` - we lose `pydantic` (which is nice functionality) but  the result is much more flexible. The tests and other modules are updated to match.

The new API is:

```python
from pyrealm.demography_two.cohorts import cohort_id_generator, create_cohorts
from pyrealm.demography_two.flora import Flora

cid_gen = cohort_id_generator()
cohorts = create_cohorts(
    pft_name=np.array(["test"] * 4),
    dbh_value=np.array([0.2, 0.4, 0.6, 0.8]),
    n_individuals=np.array([1] * 4),
    flora=flora,
    cid_generator=cid_gen,
)
```

> [!NOTE]
>
> Early versions of this PR had a halfway house class that was a thin wrapper around a dataframe, along with a CID generator and the flora. Those don't _need_ to be pinned to the class and arguably shouldn't be - floras _could_ change and having the CID generator in the class would mean needing to provide "add cohort" functionality to make sure the generator is called correctly. After discussion with @sallymatson, this PR simplifies all the way down to just using the plain dataframe as the base cohorts structure.

Fixes #670 

## Type of change

## Key checklist

## Further checks


sallymatson (commented):
Looks great!

davidorme:
Just to note: reviewed but not _technically_ approved but not onto `develop` so can still merge.

PR #692, opened 2026-07-10
Variable name updates

### LINKED ISSUE #691: Demystify splash argument names
Off the back of #683, the `splash` module uses a lot of ugly abbreviated argument names that got inherited from Tyler's code. Elsewhere, `pyrealm` uses longer, clearer names.

This is quite a big API change, but it is a quality of life improvement for users - and it makes sense to get it done before 3.0.0.

davidorme:
# Description

This PR renames argument names (mostly in `splash`) to remove old obscure names taken from Tyler's internal variable names.

Fixes #691

## Type of change

## Key checklist

## Further checks


pyrealm/core/bounds.py
```diff
@@ -73,6 +73,7 @@ class BoundsChecker:
     _defaults: tuple[tuple[str, float, float, str, str], ...] = (
         ("elevation", -1000, 9000, "[]", "m"),
         ("tc", -25, 80, "[]", "°C"),
```
Shirleycwj, line 75: Should remove line 75 with adding line 76?

Shirleycwj (changes requested):

pyrealm/core/bounds.py
davidorme: @Shirleycwj  It's a good point - the problem is that the use of `tc` is very widespread in the P Model and elsewhere and actually does convey useful information (`tc` vs `tk`).

I could align all the temperature arguments as (e.g.) `temperature_C` - that would be a lot of changes, but then we'd have one variable name.

davidorme: So - yup - good point. Could you approve this PR (which needs to keep `tc`) and then we can think about the next step of whether we need to change anything else for 3.0?

Shirleycwj: Ah got it, that makes sense if you want to stayed connect with other part of the code, I was wondering why this is the only place where you left both tc and temperature which is a bit confusing. Good to have uniformed variable names for users.

Shirleycwj (approved):

pyrealm/core/bounds.py
davidorme: @Shirleycwj  Do you think it makes sense to align `tc` to `temperature` across all functions? Are there other variable names we should do too?

Shirleycwj: I think it makes sense to have a consistent temperature variable name (in the unit of celsius) across all functions, but I'm not sure if tc or temperature leaves you less workload (i.e. less changes required)? To me tc makes more sense as it both indicates the varname (temperature) and unit (celsius or kelvin). I can't think of other variables that would force to use both unit across functions and processes at the moment.

PR #698, opened 2026-08-01
697 the pyrealm v300rc3 snagging list

### LINKED ISSUE #697: The pyrealm v3.0.0rc3 snagging list
Things to fix:

- [ ] The way in which Floras are created and used is a bit problematic. The data is validated using `pydantic` but then basically always used as a dataframe. E.g.
    https://github.com/ImperialCollegeLondon/pyrealm/blob/e70fad5ea7ccfa78700b8c415152e35a056fa864/pyrealm/demography/cohorts.py#L81-L82

   Having repeated conversion to dataframe is inefficient and can also create annoying behaviour from warnings on pydantic `model_dump` when `create_cohorts` is used. We could just follow the new cohorts model of loading it, passing it through validation and returning a validated data frame as the Flora object rather than a pydantic model.

- [ ] We currently block GPP <= 0 in inputs to StemAllocation(). Obviously negative GPP isn't a thing, but _zero_ GPP is possible for example under severe winter conditions. Can we relax that to GPP < 0? Need to check if any of the equations blow up with zero GPP.

davidorme:
# Description

This PR does two things:

* The StemAllocation class no longer forbids GPP = 0. GPP still has to be >=0 but zero productivity is a possibility. A test is added to check that passing zero into StemAllocation and GrowthIncrements gives actual values (positive or negative) and not `np.nan` or `np.inf`.

* The Flora system is rejigged - the pydantic `Flora` class is renamed to `FloraValidator` and the `create_flora` function is used to apply the validator to inputs and return a `Flora` object, which is simply a label for a pandas DataFrame. The `create_flora` function allows strict and relaxed validation and also allows the validator to be altered, to allow use of extended `FloraValidator` subclasses.

    The main reason for this change is to limit how long the flora data spends as a pydantic model and convert it early into a data frame, which is the main usage model.

Fixes #697

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/users/demography/allometry.md
+++ b/docs/source/users/demography/allometry.md
@@ -30,7 +30,7 @@ import numpy as np
 import pandas as pd
 
 from pyrealm.core.experimental import ExperimentalFeatureWarning
-from pyrealm.demography.flora import Flora
+from pyrealm.demography.flora import create_flora
 from pyrealm.demography.cohorts import create_cohorts, cohort_id_generator
 from pyrealm.demography.tmodel import StemAllometry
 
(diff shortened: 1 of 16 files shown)


sallymatson (approved):
LGTM :)

Issue #476, opened 2025-04-02
Expand documentation calculate_growth_increments in t_model API


arne-exe:
This issue should highlight/expand some parts of the documentation for the function calculate_growth_increments in the t_model script.

Specifically, highlight that dWs/dt refers to the **rate** of change, and briefly clarify how this differs from equation 7 (i.e., clarify why dD/dt is removed and explain why working with **rate** of change is needed for the model). 

See here for useful discussion: https://github.com/ImperialCollegeLondon/pyrealm/pull/461#discussion_r2018706094

Also, adjust the docstrings so that:
-In the equation for dWs/dt: the +2H needs to be placed between the final 2 brackets.
-In the equation for dWf/dt: the last H needs to be put outside of the bracket (see middle term in equation 16). Also, zeta should not be in the denominator (see middle term of equation 16).

Extra: also adjust calculate_foliar_respiration: Rf = P rf (seems like this is still an old formula, think the new formula now uses Wf ak foliage mass).

davidorme:
Yup - happy to see a PR on this.

It would also be really useful to extend `docs/source/users/demography/t_model.md` to include an user facing overview of the T Model, beyond what is embedded in detail in the API docs. I guess that doesn't have to include equations (can link to them) but does explain the moving parts.

davidorme:
Closed by #704

PR #704, opened 2026-09-11
docs: correct T-model growth equations

tannerdulay:
# Description

Updating T-model docs for mathematical clarity. Mainly correcting dW_x/dt derivatives to the correct dW_x/dD form within `calculate_growth_increments`. Names of corresponding quantities in the code are also updated. All changes should be contained within `calculate_growth_increments`. 

## Type of change

## Key checklist

- [X] Make sure you've run the `pre-commit` checks: `$ pre-commit run -a`
- [X] All tests pass: `$ poetry run pytest`

## Further checks


davidorme (commented):
This looks good to me but will let @arne-exe have the final word because he's actually tried to use these equations in the past and found them opaque!

pyrealm/demography/tmodel.py
```diff
@@ -590,16 +592,16 @@ def calculate_growth_increments(
...
-    * the wood  density of the PFT (:math:`\rho_s`),
+    * the wood density of the PFT (:math:`\rho_s`),
     * the maximum height (:math:`H_{m}`),
     * the initial slope of the height/diameter relationship (:math:`a`),
     * the crown area ratio (:math:`c`), and
     * the ratio of fine root mass to leaf area (:math:`\zeta`).
 
     The value of :math:`\Delta D` is unstable when :math:`D = 0` and hence :math:`H = 0`
-    and the rates of change in stem and foliar mass are also zero. If :math:`P_{net} - T
+    and the rates of change in stem and foliar mass are also zero. If :math:`B - T
```
arne-exe, line 602: What's the reason for changing `P_{net}` to `B`?

arne-exe (approved):
LGTM

arne-exe:
With the equations updated, #476 can probably be closed?

pyrealm/demography/tmodel.py
davidorme: That's my bad - these docs were switched to use Biomass production ($B$) as the input rather than NPP ($P_{net}$) to make it explicit that the plant might have other carbon demands on NPP. I'd left a few trailing notations and @tannerdulay  has made this consistent here.

tannerdulay:
Thanks for confirmation about B replacing NPP, @davidorme ! I assumed that was the case but glad you made sure

